PR #2, opened 2018-06-19
Provider: GitHub provider and generic interfaces

mcuadros:
Each provider should include all the functionality required to communicate with the CVS provider (GitHub, Bitbucket, etc), now just implements the `Watcher` interface.


event.go
```diff
@@ -0,0 +1,164 @@
...
+	Type() EventType
+	// Revision returns a commit revision, containing the head and the base of
+	// the changes.
+	Revision() *CommitRevision
+}
+
+// CommitRevision defines a range of commits, from a base to a head.
+type CommitRevision struct {
+	// Base of the revision.
+	Base ReferencePointer
```
smola: Once we merge this, we'll need to check if/how CommitRevision/ReferencePointer interact with structs in the analyzer/changes API. Maybe we can reuse some structs, or make them comply with a common interface.

provider/github/watcher.go
```diff
@@ -0,0 +1,154 @@
...
+	"time"
+
+	"github.com/src-d/lookout"
+	"github.com/src-d/lookout/util/cache"
+
+	"github.com/google/go-github/github"
+	"github.com/gregjones/httpcache"
+	"github.com/gregjones/httpcache/diskcache"
+	"gopkg.in/sourcegraph/go-vcsurl.v1"
+	"gopkg.in/src-d/go-errors.v0"
```
smola: Not `go-errors.v1`?

watcher.go
```diff
@@ -0,0 +1,29 @@
+package lookout
+
+import (
+	"context"
+
+	"gopkg.in/src-d/go-errors.v0"
```
smola: `v1`

util/cache/validable.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+var ErrNotFoundKey = errors.NewKind("Unable to find the given key, %s")
+
+// ValidableCache represents a cache, were each new entry should be validate
+// before be read or write a new entry, otherwise is forget and discarded.
+type ValidableCache struct {
+	httpcache.Cache
+
+	temporal struct {
```
smola: `s/temporal/temporary/`

event.go
mcuadros: Yep, I am not sure if this is the best struct, but the concept is important.

PR #3, opened 2018-06-21
Syncer and Library abstractions

mcuadros:
- Library manages the local copy of the repositories.
- Syncer syncs a given revision of a repository to the local copy.

### DIFF
--- a/library.go
+++ b/library.go
@@ -1,0 +1,122 @@
+package lookout
+
+import (
+	"fmt"
+	"os"
+
+	"gopkg.in/src-d/go-billy.v4"
+	"gopkg.in/src-d/go-errors.v1"
+	"gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/config"
+	"gopkg.in/src-d/go-git.v4/storage"
+	"gopkg.in/src-d/go-git.v4/storage/filesystem"
+)
+
+var (
+	ErrRepositoryExists    = errors.NewKind("repository %s already exists")
+	ErrRepositoryNotExists = errors.NewKind("repository %s not exists")
+)
+
+// Library controls the persistence of multiple git repositories.
+type Library struct {
+	fs billy.Filesystem
+}
+
+// NewLibrary creates a new Library based on the given filesystem.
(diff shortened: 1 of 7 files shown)


library.go
```diff
@@ -0,0 +1,99 @@
...
+	"gopkg.in/src-d/go-git.v4/storage"
+	"gopkg.in/src-d/go-git.v4/storage/filesystem"
+)
+
+var (
+	ErrRepositoryExists    = errors.NewKind("repository %s already exists")
+	ErrRepositoryNotExists = errors.NewKind("repository %s not exists")
+)
+
+// Library is controls the persistence of multiple git repositories.
```
smola: `s/is controls/controls/`

library.go
```diff
@@ -0,0 +1,99 @@
...
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	return r, nil
+}
+
+// Has return true if a repository with the given URL exists.
```
smola: `s/return/returns/`

library.go
```diff
@@ -0,0 +1,99 @@
...
+func (l *Library) Has(url *RepositoryInfo) bool {
+	_, err := l.fs.Stat(l.repositoryPath(url))
```
smola: Do we really want to hide permission errors?

syncer.go
```diff
@@ -0,0 +1,41 @@
+package lookout
+
+import (
+	"context"
+	"fmt"
+
+	git "gopkg.in/src-d/go-git.v4"
```
smola: `s/git //`

syncer.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// Sync syncs the local git repository to the given commit revision.
+func (s *Syncer) Sync(ctx context.Context, rev *CommitRevision) error {
+	r, err := s.l.GetOrInit(rev.Head.Repository)
+	if err != nil {
+		return err
+	}
+
+	opts := &git.FetchOptions{
```
smola: We don't need to execute a fetch if we already have the commit.

mcuadros: But we need to create the branch, so...

smola: Only if it does not exist already.

Issue #8, opened 2018-07-11
Prototype analyzer SDK with means to test it


carlosms:
As discussed, this issue is to create PoC of an SDK for analyzers.

To make it easier to reference to other teams, it will be in `./sdk` folder inside this repo. For now it will only have a main README, containing links to `.proto` files in this repo, and links to documentation on how to generate code. It probably should also contain instructions on how to run bblfshd

As part of release, we need to provide a small pre-built binary to testing the analyzers. This binary, dubbed `look`, will be a simplified version of the whole `lookout` server.

On execution, it will:
- trigger an immediate gRPC call to the analyzer
- provide any required repository data (act as a real DataServer)
- print the analyzer's response in the console
- exit

 It should be as simple as:
  - `look review --path path/to/single-git-repo --from revision --to revision <Analyzer gRPC>`
  - `look push` a hook for training, working the same way

In this case `review` would be a subcommand, to accommodate for any number of methods we may add in the future.
Names, use of arguments, flags or env. vars., etc. all will be added later on.

As suggested by @bzz,  default behaviour would be `look review ip4://localhost:NNNN` that will trigger a review API for changes between latest head and one commit before it for repository in a current directory.

bzz:
One thing we also discussed is - names agnostic to code review platform, so I would suggest using something agnostic: 
 - push
 - review

Most probably default behaviour would be`lookout-sdk review ip4://localhost:NNNN` that will trigger a review API for changes between latest head for repository in a current directory, and one commit before it.

smacker:
Few questions:

1. Do we want to move `.proto` files into sdk directory?
2. What will be the default behavior for `push` command? Same HEAD & HEAD^?
3. Currently in `PullRequestEvent` we fill only `commit_revision` field. But we also have `source` & `merge` fields. Are we okay with keeping it empty? Or do we want to fill it with something? I'm a bit confused how to fill it if we take only HEAD&HEAD^ from the same branch.

bzz:
> 1. 
Yes, only the .proto

> What will be the default behavior for push command? Same HEAD & HEAD^?

Yes,  I think same is nice

> `PullRequestEvent` ... But we also have `source` & `merge` fields

Good question!  `source` is `--to`

And let's add URLs to https://github.com/smacker/lookout/blob/f9e027e867f021283ddb1a129de3225232d6d627/cmd/lookout/analyze.go#L94

PR #10, opened 2018-07-16
add sdk readme

smacker:
Documentation required by #8.

Keep it super simple for now.

It doesn't explain services methods or arguments of lookout review binary because most probably they are about to change.

### DIFF
--- a/sdk/README.md
+++ b/sdk/README.md
@@ -1,8 +1,8 @@
 # Lookout Analyzer SDK
 
-An analyzer should implement [gRPC service](https://github.com/src-d/lookout/tree/master/pb/service_analyzer.proto) to receive events from the server.
+An analyzer is a gRPC server, that implements [Analyzer service](https://github.com/src-d/lookout/tree/master/pb/service_analyzer.proto) to receive events from the server.
 
-It also should use gRPC client for [data service](https://github.com/src-d/lookout/tree/master/pb/service_data.proto) to get content or UAST of the changes.
+An analyzer should use gRPC client to access [Data service](https://github.com/src-d/lookout/tree/master/pb/service_data.proto) to get content or UAST of the changes.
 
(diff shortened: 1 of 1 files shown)


sdk/README.md
```diff
@@ -0,0 +1,39 @@
+# Lookout Analyzer SDK
+
+An analyzer should implement [gRPC service](https://github.com/src-d/lookout/tree/master/pb/service_analyzer.proto) to receive events from the server.
```
bzz: For people who might not be very proficent with gRPC, how about making server/client separation more explicit  and then making both links to .proto to have a consistent wording, like

`An analyzer is a gRPC server, that implements [Analyzer service]() ...`

and then

`An analyzer should use gRPC client to access [Data service]() to ...`

sdk/README.md
```diff
@@ -0,0 +1,39 @@
...
+It also should use gRPC client for [data service](https://github.com/src-d/lookout/tree/master/pb/service_data.proto) to get content or UAST of the changes.
+
+All `.proto` files are located in [/pb](https://github.com/src-d/lookout/tree/master/pb) directory.
+
+
+## Code generation
+
+### Python
+
+Dependencies:
```
bzz: Let's also link https://grpc.io/docs/quickstart/python.html somewhere

sdk/README.md
```diff
@@ -0,0 +1,39 @@
...
+    pb/*.proto
+```
+
+## Testing
+
+Download the latest lookout binary from [releases](https://github.com/src-d/lookout/releases) page.
+
+Babelfish server is required. If you don't have it running, please read the [getting started guide](https://doc.bblf.sh/using-babelfish/getting-started.html), to learn more about how to start and use it.
+
+Run the binary inside a git repository directory with gRPC address of the analyzer to trigger review event.
```
bzz: `to trigger review event` - let's document here the defaults, so the users know which revisions are going to be used for this event as base/head.

PR #12, opened 2018-07-17
Make protogen

smacker:
Move proto generation to Makefile.

The last gogo release breaks everything:
https://github.com/gogo/protobuf/issues/428
https://github.com/gogo/protobuf/issues/430

It's safer to use gogo from vendor directory.


smacker:
repushed with only change `go` -> `$(GOCMD)`

Issue #15, opened 2018-07-18
[SDK] Update Data Service API


bzz:
- Get full repository content, to stream back all files from a given revision 
   (But keep current defaults of `GetChanges`)
- Remove “changed_only” flag
- Add flag to “exclude_vendored”

PR #9, opened 2018-07-16
Review command

smacker:
Review command part of #8 

- rename `analyze` command to `review`
- change arguments according to the issue description
- temporary workaround to avoid ugly logging

PR #22, opened 2018-07-18
Update Data Service API

smacker:
Fix: #15

Based on #9 (only last 2 commits are new)


bzz:
Nice! I belive it's last 3 commits that are new. As #9 is merged now, could you please rebase?

smacker:
Correct. I pushed one more commit after creating PR and writing the description and didn't update it.

Rebased on master. Now the only code for this PR is visible in GH.

service/bblfsh/bblfsh.go
```diff
@@ -42,22 +44,95 @@ func (s *Service) GetChanges(ctx context.Context, req *lookout.ChangesRequest) (
...
 	}, nil
 }
 
-type ChangeScanner struct {
-	underlying    lookout.ChangeScanner
+func (s *Service) GetFiles(ctx context.Context, req *lookout.FilesRequest) (
+	lookout.FileScanner, error) {
+	wantContents := req.WantContents
+	if req.WantUAST {
+		req.WantContents = true
```
bzz: Allthough this is how GetChanges was implemented - this seems a bit strange as there is no way right now for a client, to request only UASTs, isn't it?

service/bblfsh/bblfsh.go
```diff
@@ -42,22 +44,95 @@ func (s *Service) GetChanges(ctx context.Context, req *lookout.ChangesRequest) (
...
+	if !req.WantUAST {
+		return files, nil
+	}
+
+	return &FileScanner{
+		underlying: files,
+		BaseScanner: BaseScanner{
+			ctx:           ctx,
+			client:        s.client,
+			purgeContents: !wantContents,
```
bzz: Could you help me understand if `purgeContents` can even happen here? `wantContents` seems to be always True, if an execution has reached here after L60.

service/bblfsh/bblfsh.go
```diff
@@ -42,22 +44,95 @@ func (s *Service) GetChanges(ctx context.Context, req *lookout.ChangesRequest) (
...
+
+	if s.purgeContents {
+		f.Content = nil
+	}
+
+	return nil
+}
+
+func (s *BaseScanner) parseFile(f *lookout.File) (
+	*uast.Node, error) {
```
bzz: Genuinely interested, is there a convention or styleguide for formatting it like, this instead of

```
func (s *BaseScanner) parseFile(f *lookout.File) (*uast.Node, error) {
}
```
?

smacker: This style is used only when the line is too long. Go doesn't have specific styleguide for it:
https://github.com/golang/go/wiki/CodeReviewComments#line-length

but in this particular case, the function isn't long. So it should be one-liner. 👍

service/bblfsh/bblfsh.go
smacker: We have to set `req.WantContents` to get content from git service.
But we can remove it from the response when we got UAST from it.
I don't think it should be fixed in this PR. So I created an issue: https://github.com/src-d/lookout/issues/27

smacker: oops. There is `purgeContents` already I didn't notice. Forget about it.

bzz:
Question: do we want to have exported methods/functions/types in a new code to have a goDoc?

service/bblfsh/bblfsh.go
smacker: line 58: `wantContents := req.WantContents` creates new boolean with original client request.
line 60: don't change it.

smacker:
- Rebased on master.
- I added *some* godoc. golint still returns 106 warning. We can fix it later.

PR #25, opened 2018-07-18
sdk push command

smacker:
Fix: #8

### DIFF
--- a/cmd/lookout/event.go
+++ b/cmd/lookout/event.go
@@ -1,0 +1,122 @@
+package main
+
+import (
+	"fmt"
+	"path/filepath"
+
+	"github.com/src-d/lookout"
+	"github.com/src-d/lookout/service/bblfsh"
+	"github.com/src-d/lookout/service/git"
+	"google.golang.org/grpc"
+
+	gogit "gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/plumbing"
+)
+
+type EventCommand struct {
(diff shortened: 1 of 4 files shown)


cmd/lookout/event.go
```diff
@@ -0,0 +1,122 @@
...
+	"google.golang.org/grpc"
+
+	gogit "gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/plumbing"
+)
+
+type EventCommand struct {
+	DataServer string `long:"data-server" default:"ipv4://localhost:10301" env:"LOOKOUT_DATA_SERVER" description:"gRPC URL to bind the data server to"`
+	Bblfshd    string `long:"bblfshd" default:"ipv4://localhost:9432" env:"LOOKOUT_BBLFSHD" description:"gRPC URL of the Bblfshd server"`
+	GitDir     string `long:"git-dir" default:"." env:"GIT_DIR" description:"path to the .git directory to analyze"`
```
carlosms: should we make the env var LOOKOUT_GIT_DIR for consistency?

smacker:
added simple test to be sure dummy works

cmd/lookout/event.go
carlosms: The description makes it sound like you need to provide `path/reponame/.git`, but it actually is `path/reponame`, right?

It might be more clear with "path to the Git repository to analyze"

smacker: please create a new issue for that if you think so. It's out of the scope of this pr.

Issue #19, opened 2018-07-18
[SDK] Pass configuration to Analyzers


bzz:
Analyzer should receive a configuration from Lookout server i.e as KV `ReviewEvent.Configuration`
    ```
    repeated Config configuration = N;

     message Config {
      string key = 1;
      string val = 2;
    }
    ```

smacker:
Is there any specific reason to use such format?

Protobuf has nice thing [proto.struct](https://github.com/google/protobuf/blob/master/src/google/protobuf/struct.proto) which allows to send any json object around.
In my opinion, it's more flexible format and we can benefit from it.
The simplest example if analyzer needs a value as a list of number/string.

Issue #24, opened 2018-07-18
[SDK] Add simple integration test to CI


bzz:
Add CI profile for integration tests that
 - builds `lookout` binary (part of SDK)
 - launches dummy analyzer `dummy serve`
 - runs `lookout review ipv4://localhost:10302`
 - at least check `grep BEGIN RESULT`
 - runs `lookout push ipv4://localhost:10302`
 - greps for "dummy comment for push event"

Add another profile that verifies that
 - `make protoc` passes
 - `python -m grpc_tools.protoc` as described in https://github.com/src-d/lookout/tree/master/sdk#python passes

smacker:
Implemented

Issue #20, opened 2018-07-18
Configurable Analyzers


bzz:
- [x] add support for multiple Analyzers in lookout server 
  - [x] read per-repository configuration with DataServer API `GetFiles({IncludePattern: "^\.lookout\.yml$"})` #55 
  - [x] hard-code static global configuration of the server
  - [x] merged global and repo-specific configurations #71
  - [x] pass configuration to each individual Analyzer in `NotifyReview` gRPC  #55

smacker:
> add support for multiple Analyzers in lookout server

Should analyzers be defined in some config file or using flags/env vars as we do right now?

bzz:
As we discussion, Analyzers should be configurable with the same options either 
 - from `.lookout.yml`
 - from server side

On the server side, this issue just includes `hard-code static global configuration of the server`, leaving the details of how to load this configuration for later.

That way, loading it from a file can be handled in separate issue.

smacker:
Okay. Thanks! That means we most probably need to remove current flag/env for an analyzer in the server.

smacker:
> merged global and repo-specific configurations

how exactly do we want to do that? What should be merged and what should be replaced?

bzz:
From discussion elsewhere:

Global config
```analyzers:
  - name: Dummy
    addr: ipv4://localhost:10302
    settings:
      key1: 1
      key2: [1,2]
      key3:
       subkey1: 1
       subkey2: [1, 2]
       subkey3:
        subsubkey: 1```


repo config:
```analyzers:
  - name: Dummy
    settings:
      key2: [5]
      key3:
       subkey1: 3
       subkey3:
        subsubkey2: 2
       subkey4: 4
```

would result in merged 

```
analyzers:
  - name: Dummy
    addr: ipv4://localhost:10302
    settings:
      key1: 1
      key2: [5]
      key3:
       subkey1: 3
       subkey2: [1, 2]
       subkey3:
        subsubkey: 1
        subsubkey2: 2
       subkey4: 4
```

PR #26, opened 2018-07-18
Pass configuration to Analyzers

smacker:
Fix: #19 

Hopefully, I understood right that the issue is only about adding the field but implementation will be in another issue.

This change required changes to protofiles and makefile (spent quite much time to figure out how to do it)
The reason is https://github.com/gogo/protobuf/blob/master/types/struct.pb.go implements Sizer interface but not ProtoSizer.
At the same time, bblfsh UAST implements ProtoSizer one.


bzz:
Would any updates to https://github.com/src-d/lookout/tree/master/sdk#python are required as well?

Might be good idea to wait a bit and rebase on https://github.com/src-d/lookout/issues/24 just to verify.

sdk/event.proto
```diff
@@ -39,12 +38,16 @@ message PushEvent {
...
     uint32 distinct_commits = 5;
+    // Configuration contains any configuration related to specific analyzer
+    google.protobuf.Struct configuration = 6 [(gogoproto.nullable) = false];
 
     CommitRevision commit_revision = 7 [(gogoproto.nullable) = false, (gogoproto.embed) = true];
 }
 
 // ReviewEvent represents a Review (PullRequest in case of Github) being created or updated.
 message ReviewEvent {
+    option (gogoproto.protosizer) = false;
```
bzz: As you say, it takes time to figure it out, so might be a good idea to leave a bread-crumb i.e a comment on why is this required. Wdyt?

smacker:
No need to update python. All changes are related to the go code generation.

bzz:
> Hopefully, I understood right that the issue is only about adding the field but implementation will be in another issue.

Exactly, for now actual implementation is the last item in ☂️ #20

smacker:
Everything was tested locally.
Also in CI, the new command for go and old python command were tested in:
https://github.com/src-d/lookout/pull/31 and https://github.com/src-d/lookout/pull/32

Merging.

PR #30, opened 2018-07-19
Simple integration test

smacker:
part of #24


Makefile
```diff
@@ -16,3 +16,27 @@ protogen:
...
+		kill $$PID; \
+		exit 1; \
+	fi; \
+	$(BIN_PATH)/lookout push ipv4://localhost:10302 | grep "dummy comment for push event"; \
+	if [ $$? != 0 ] ; then \
+		echo "push test failed"; \
+		kill $$PID; \
+		exit 1; \
+	fi; \
+	kill $$PID || true ; \
```
bzz: Scenario looks nice.

But can we please extract this logic to a small .sh file that follows https://google.github.io/styleguide/shell.xml and just call it from this target?

Maintaining Makefiles is a nightmare.

smacker: this is shell script but inside makefile. Absolutely similar to how src-d CI works.

Makefile
```diff
@@ -16,3 +16,27 @@ protogen:
...
+
+# target for integration-test. Regular build target puts linux binary to BIN_PATH on MacOS
```
carlosms: I wonder if this should be done as a new rule of Makefile.main, it might be needed in other projects.

smacker: I don't know any other project where we would need it.
I would suggest to move it to Makefile.main only when there are at least 2 such projects.

Makefile
bzz: 👍 let's extract it to a separate file

smacker: looks like it's the only sh script we have for now. It's only 15 lines of code, moving it to separate sh would make it more complex. (need to handle lack of binaries for example). Let's keep it here for now.
If we will need to make testing more complex we will move code to .sh files or use some other tools.

bzz: Makes sense, agreed! Thank you for thoughtful consideration

PR #31, opened 2018-07-19
Validate that proto files and generated code are in sync 

smacker:
Based on #30

Part of #24 

Protogen is failing on travis with current comment: Updated makefile using command from #26

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,14 +11,30 @@ $(MAKEFILE):
 	git clone --quiet --depth 1 -b $(CI_BRANCH) $(CI_REPOSITORY) $(CI_PATH);
 -include $(MAKEFILE)
 
-# SDK
+# Environment
+OS := $(shell uname)
+
+# SDK binaries
 DUMMY_BIN := $(BIN_PATH)/dummy
 LOOKOUT_BIN := $(BIN_PATH)/lookout
 
+# Protoc
+PROTOC_DIR ?= ./protoc
+PROTOC := $(PROTOC_DIR)/bin/protoc
+PROTOC_VERSION := 3.6.0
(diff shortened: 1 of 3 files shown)


bzz:
Same as in #30 - let's have `install-protoc` logic call a nice small reusable .sh file stored i.e inside `_tools` dir.

smacker:
why?
This is how other src-d projects work:
https://github.com/src-d/ci/blob/master/Makefile.main
https://github.com/bblfsh/client-go/blob/master/Makefile
https://github.com/bblfsh/sdk/blob/master/Makefile
https://github.com/bblfsh/bblfshd/blob/master/Makefile

Makefile
```diff
@@ -11,8 +11,60 @@ $(MAKEFILE):
 	git clone --quiet --depth 1 -b $(CI_BRANCH) $(CI_REPOSITORY) $(CI_PATH);
 -include $(MAKEFILE)
 
-protogen:
+OS := $(shell uname)
+PROTOC_DIR ?= $(HOME)/protoc
```
carlosms: I like the idea of fixing the protoc version. But if we are overriding any global installation, I think it may be better to install it in the current dir, in `./protoc`, or `./.protoc`. `$(HOME)/protoc` could conflict if you work in two src-d projects with different protoc versions. What do you think?

smacker: thanks for the suggestion! 👍 

and there is `/protoc` in gitignore already (because $HOME == $CURRENT_DIR) in travis.

bzz:
>> let's have install-protoc logic call a nice small reusable .sh
>
> why?

Because it easier to maintain it that way then in a makefile

Issue #17, opened 2018-07-18
Data Service: implement vendor exclusion


bzz:
https://github.com/src-d/lookout/issues/15 introduced `ChangesRequest.ExcludeVendored` flag in Data service API.

This issue is about implementing actual exclusion behaviour for of files, detected by [`enry.IsVendor`](https://godoc.org/gopkg.in/src-d/enry.v1#IsVendor) from the `GetChanges` response.

smacker:
fixed in https://github.com/src-d/lookout/pull/44

PR #33, opened 2018-07-19
Data server filter vendor

smacker:
Fix: #17


smacker:
@bzz please provide the code example. I don't really understand what you propose.

bzz:
Answered in #44 - indeed that looks very good!

smacker:
close in favor of #44

Issue #29, opened 2018-07-19
[SDK] Make user experience more like curl


bzz:
Improve user experience of lookout SDK

 0. [x] add `-v` to push/review commands, to hide debug details (i.e all nifty gRPC)  #37
 1. [x] If no bblfsh service available, show just 1 warning that no UASTs will be available and trigger analyzer #34
     Right now it prints lots of logs and takes ~3 sec
 2. [x] if no analyzer running - just print 1 line of warning and stop (same as cURL) #34
     Right now it attempts to re-connect \w lots of verbose logs

smacker:
During code review, we found more ideas for improvements. They are tracked in separated issues.

PR #34, opened 2018-07-19
SDK: Don't reconnect to bblfsh or analyzer

smacker:
Part of #29

### DIFF
--- a/cmd/lookout/event.go
+++ b/cmd/lookout/event.go
@@ -1,8 +1,10 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"path/filepath"
+	"time"
 
 	"github.com/src-d/lookout"
 	"github.com/src-d/lookout/service/bblfsh"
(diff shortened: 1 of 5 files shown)


carlosms:
This is great if you are developing an analyzer that doesn't need UAST, but if you do need it, and fail to connect lookout to bblfsh properly, the current code will ignore `req.WantUAST` silently, right?

A single warning at startup will be easy to miss if the requests themselves return success, but with unexpected contents. I think a request with `wantUAST` should fail, with an error message that informs that lookout could not connect to bblfsh.

carlosms:
Another option I can think of is to make `--bblfshd` optional, and if it is set make lookout try to connect once, print the error and exit.

smacker:
I would prefer to implement in separate PR to unblock sequential PRs and don't make this PR too huge. Here is an issue: https://github.com/src-d/lookout/issues/45

PR #36, opened 2018-07-19
Add travis release uploads

### LINKED ISSUE #23: [SDK] Add CI automation for release
Add automation to produce a release on tag, that include next artefacts:
 - content of `./sdk` dir
 - `lookout` binaries for linux/mac os

in a single archive, uploaded to GH release.

carlosms:
Fix #23.

We need to add GITHUB_TOKEN to the travis settings. Is this something only @bzz can do?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,9 @@ before_install:
   - sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-6 90
   - sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-6 90
 
-script:
-  - make test-coverage codecov
+stages:
+  - name: release to github
+    if: tag IS present
 
 jobs:
   include:
@@ -24,7 +25,12 @@ jobs:
       name: "Unit Tests"
       script: make test-coverage codecov
     - script: make test-sdk
-      name: "Integration Tests"
+      name: "SDK Integration Tests Linux"
+    - script: make test-sdk
+      name: "SDK Integration Tests macOS"
+      os: osx
+      osx_image: xcode9.4
+      before_install: skip
     - script:
         - make protogen
         - make no-changes-in-commit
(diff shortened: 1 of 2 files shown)


bzz:
As a user, I would expect release artifacts for SDK include: and archive per-platform with a single dir containing:
 - a binary `lookout`
 - .proto files

And later, another artefact would be included - a `lookoutd` binary of the server.

bzz:
Sorry, has just noticed a question in parent issue :/ answered in there as well https://github.com/src-d/lookout/issues/23#issuecomment-406561961

carlosms:
@bzz please take another look.

Travis will now upload `lookout_sdk_*.tar.gz` with the lookout bin and the sdk dir as you described.

bzz:
@carlosms 👍  

> MacOS binaries will be unexecutable due to cgo in bblfsh.

yes, I was about to say just that, tested and we need to fix it.

carlosms:
I've pushed a change that builds the packages on macOS, please take a look.

I also added a macOS build with the SDK integration test, to know if something breaks for macOS before a new release.

.travis.yml
```diff
@@ -35,6 +41,24 @@ jobs:
         - export PY_OUT_DIR=py
         - mkdir -p "$PY_OUT_DIR" && python3 -m grpc_tools.protoc -Isdk --python_out=$PY_OUT_DIR --grpc_python_out=$PY_OUT_DIR sdk/*.proto
       name: "Protobuf code generation"
+    - stage: release to github
+      name: linux packages
```
bzz: Let's use `"` as in every other name, for consistency

carlosms: done, and rebased to be in commit only

.travis.yml
```diff
@@ -15,16 +15,22 @@ before_install:
...
       name: "Unit Tests"
       script: make test-coverage codecov
     - script: make test-sdk
-      name: "Integration Tests"
+      name: "SDK Integration Tests Linux"
+    - script: make test-sdk
+      name: "SDK Integration Tests macOS"
+      os: osx
+      osx_image: xcode9.4
+      before_install: echo "skipping before_install for macOS"
```
bzz: Here and below, just wondering, why not using `skip` or `ignore` [keyword](https://docs.travis-ci.com/user/build-stages/#Build-Stages-and-Deployments) ?

carlosms: For no special reason, I just copied it from [gitbase travis.yml](https://github.com/src-d/gitbase/blob/master/.travis.yml) and I was not aware of the skip keyword. I'll replace it before merging.

PR #37, opened 2018-07-19
Lookout verbose option

smacker:
Second (zero) part of #29

### DIFF
--- a/cmd/lookout/event.go
+++ b/cmd/lookout/event.go
@@ -22,6 +22,7 @@ type EventCommand struct {
 	GitDir     string `long:"git-dir" default:"." env:"GIT_DIR" description:"path to the .git directory to analyze"`
 	RevFrom    string `long:"from" default:"HEAD^" description:"name of the base revision for event"`
 	RevTo      string `long:"to" default:"HEAD" description:"name of the head revision for event"`
+	Verbose    bool   `long:"verbose" short:"v" description:"enable verbose logging"`
 	Args       struct {
 		Analyzer string `positional-arg-name:"analyzer" description:"gRPC URL of the analyzer to use"`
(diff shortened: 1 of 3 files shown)


bzz:
Do you think there is a value to have some logs printed for opening/closing a connection  in non-verbose mode?

Right now experience using `time ./lookout review ipv4://localhost:10302` is a bit confusing - after running it, one needs to wait ~3-4sec when nothing happens until `BEGIN RESULT` shows up.

Suggestion:
1. Success 
```
Connected to analyzer at XXX
Sending NotifyReview request
``` 
if something like this is printed, a user would have more patience. 

2. Failure
Right now, if `time ./lookout review ipv4://localhost:666` just has no feedback to user at all.
Suggestion:
```
Failed to connect to analyzer at XXX
Retrying (1 out of Y)
...
Failed to connect to analyzer at XXX
Retrying (2 out of Y)
``` 

Having both of those would allow to wave the mistakes related to networking for a user and thus help building a trust in SDK (he does not need to check with `nc` manually if server is reachable, etc)

Also, do you think there is a value of applying same `-v` for dummy analyzer as well? 
May be does not need to be that detailed, but may be nice to have

smacker:
@bzz 
> Failure to connect to analyzer without -v it's very hard to identify: does it taking to long? does it fail to connect?
this is why commits are in opposite order to issue bullets. 

With the prev commit it will fail fast with correct error.

bzz:
> would be nice 👍 let's update issue/create new one with description of proper success logging across binaries and implement it.

Improving in separate issue sounds good - would you please create one so the feedback is not getting lost?

smacker:
rebased on master. Merge after CI is green.

PR #38, opened 2018-07-19
Multiple analyzers

smacker:
The first point in #20 

Not sure if I understood it correctly.

There is a new target `make dry-run` to run the server.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,13 +14,11 @@
 # ci files
 .ci
 
-<<<<<<< HEAD
 # build results
 /build
 
 # dependencies
 /protoc
-=======
+
 # Configuration
 /config.yml
->>>>>>> 88a0c8b... support multiple analyzers
--- a/.gitignore
+++ b/.gitignore
@@ -14,8 +14,13 @@
 # ci files
 .ci
 
+<<<<<<< HEAD
 # build results
 /build
 
 # dependencies
 /protoc
+=======
+# Configuration
+/config.yml
+>>>>>>> 88a0c8b... support multiple analyzers
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ $(MAKEFILE):
 
 # Environment
 OS := $(shell uname)
+CONFIG_FILE := config.yml
 
 # SDK binaries
 DUMMY_BIN := $(BIN_PATH)/dummy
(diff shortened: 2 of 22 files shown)


bzz:
Seems to address both, 1 and 3 from #20 

`dry-run` idea is 👍 as that is something that can be used/refactored to `-mode cli` for #40

cmd/lookout/serve.go
```diff
@@ -151,22 +175,36 @@ func (c *ServeCommand) handlePR(e *lookout.ReviewEvent) error {
...
-		logger.Infof("no comments were produced")
-		return nil
+	var comments []*lookout.Comment
+
+	for name, a := range c.analyzers {
+		aLogger := logger.With(log.Fields{
+			"analyzer": name,
+		})
+
+		resp, err := a.NotifyReviewEvent(context.TODO(), e)
```
bzz: This would block on every analyzer, waiting for response, before sending a request to the next one, wouldn't it?

smacker: true. Thanks! I was still in a mindset of sdk where everything should be synchronous. 

I update PR using WaitGroup & Mutex.
Though we might use another solution later.

cmd/lookout/serve.go
```diff
@@ -21,10 +25,21 @@ func init() {
...
+	Addr string
+}
+
+// Config is a server configuration
+type Config struct {
+	Analyzers []AnalyzerConfig
+}
+
 type ServeCommand struct {
+	ConfigFile  string `long:"config" short:"c" default:"config.yml" env:"CONFIG_FILE" description:"path to configuration file"`
```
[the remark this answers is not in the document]
smacker: currently, none of the other env vars use this prefix. Should we add it everywhere? What about dummy analyzer and serve commands?

smacker:
rebased on master

.gitignore
```diff
@@ -14,8 +14,13 @@
...
+<<<<<<< HEAD
 # build results
 /build
 
 # dependencies
 /protoc
+=======
+# Configuration
+/config.yml
+>>>>>>> 88a0c8b... support multiple analyzers
```
bzz: Seems like artefact of rebase left by accident

Makefile
```diff
@@ -83,3 +84,9 @@ $(DUMMY_BIN):
 clean-sdk:
 	rm -f $(DUMMY_BIN)
 	rm -f $(LOOKOUT_BIN)
+
+.PHONY: dry-run
+dry-run: $(CONFIG_FILE)
+	go run cmd/lookout/*.go serve --dry-run github.com/src-d/lookout
+$(CONFIG_FILE):
+	cp "$(CONFIG_FILE).tpl" $(CONFIG_FILE);
```
bzz: empty line?

smacker:
Thanks! Fixed. Waiting for CI & Merge.

PR #44, opened 2018-07-20
 remove copy-paste from scanners 

smacker:
Proposal:

Look at the last commit.

The idea:
Create FnScanners for file&change.
They implement Scanner interface and handle start/stop/err correctly.

The function operating on file/change looks like:
```go
func (ch *lookout.Change) (skip bool, err error)
```

If you like the idea I'll change tests too. It's much easier to test such functions than iterators.


bzz:
That, sir, looks very much better indeed!

The only thing is that instead of `Fn`, I would suggest using something like `Filter`, so `FnChangeScanner` -> `ChangeScannerFilter` and then

 `Fn      func(*Change) (bool, error)` could also become something like `Accept      func(*Change) (bool, error)` so the client looks like `skip, err := s.Accept(f)`

Or using `Filter` as a method name as well, instead of `Accept` (but may be a bit less clear)

bzz:
`func (s *FnFileScanner) Next() bool` and `func (s *FnChangeScanner) Next() bool` still look very similar, so there definitely must be opportunity for even further improvements, but that can be left as a future work 😉

smacker:
@bzz filter or accept aren't very good names because those scanners implement not only filtering but any file/change processing.
For example, take a look at BlobScanner. It adds blob to a file. Imo it would be weird to call it filter.

bzz:
@smacker fair enough, it's just that this PR does not touch them, so they are only filtering, thus a suggestion

smacker:
🤔 this PR does touch them: For example https://github.com/src-d/lookout/pull/44/files#diff-8aed9cc71de3309ab895ce6a16baec09R270

smacker:
rebased. merge after CI is green.

PR #54, opened 2018-07-24
Add version to analyzers response

bzz:
Proposal: add Analyzer version field to the response. 

It could be stored on HistoryDB and then taken into account in retroactive analysis about our suggestion quality. 

Another option coiuld be expose it though gRPC but that would increase a traffic between serve<->analyzer.

Either way, as it changes .proto it better be added early on.


cmd/dummy/main.go
```diff
@@ -13,7 +13,10 @@ import (
 
 const maxMsgSize = 1024 * 1024 * 100 // 100mb
 
-var parser = flags.NewParser(nil, flags.Default)
+var (
+	varsion = "local_build_1"
```
carlosms: typo, version

PR #55, opened 2018-07-24
Read config

smacker:
Part of #20

There is no merging (yet?). Local configuration of analyzer just replaces the global one.


server.go
```diff
@@ -2,23 +2,47 @@ package lookout
...
+	Name string
+	// Addr is gRPC URL.
+	// can be defined only in global config, repository-scoped configuration is ignored
+	Addr string
+	// Settings any configuration for an analyzer
+	Settings map[string]interface{}
+}
+
+// Config is a server configuration
+type Config struct {
```
bzz: a name like `ServerConfig` may be could help avoid the confusion?

pb/struct.go
```diff
@@ -0,0 +1,163 @@
...
+
+import (
+	"fmt"
+	"reflect"
+
+	types "github.com/gogo/protobuf/types"
+)
+
+// ToStruct converts a map[string]interface{} to a types.Struct
+func ToStruct(v map[string]interface{}) *types.Struct {
```
bzz: I did not use gogo protobuf very much, so could you please share a bit, why in order to create new Struct a https://godoc.org/github.com/gogo/protobuf/types#NewPopulatedStruct would not work for us?

bzz:
Overall SGTM, but a bit confused with how instances of type.Struct are created.

pb/struct.go
smacker: Because it accepts `randyStruct` but we need `map[string]interface{}`.
We parse unknown structure from yaml, the only type we can use is `map[string]interface{}`.

smacker:
@bzz please suggest any other way to create Struct if you know any.

server.go
smacker: renamed

bzz:
My hope was that may be you, as an author, have already done some kind of research and could share the results i.e if some [gogo/protobuf extensions](https://github.com/gogo/protobuf/blob/master/extensions.md) might be able to provide such thing.

I.e helful feedback would include the fact that official Golang protobuf impl is not going to provide this https://github.com/golang/protobuf/issues/370

smacker:
Of course, I have done research. I found nothing better than creating a custom transformer. So I had to implement it by myself. You see it in PR.

Your point is it looks "confused". Sadly, I don't know how to make it less confusing.
Please provide your suggestion on how to fix it. Otherwise, the discussion isn't very productive.

Issue #40, opened 2018-07-19
CLI input/output for serve


bzz:
Right now `lookout serve` can only trigger analysis with actual Github events.
This issue is about changing that, so that STDIO can be used as Input and Output for it from the CLI.

It includes:
  - a CLI trigger, to separate polling Github and reading Stdio modes:
     i.e `lookout serve -mode cli` that reads stdio and exists VS `lookout serve -mode github` that polls Github or something similar 
  - implementation of CLI Provider that
    * listens to `stdin` of the process
    * prints only the comments to `stdout`
    * supports structured output in JSON i.e with `-o json`
  - prints logs only to `stderr`
  - has simple integration test:
    ```
     ./dummy
     cat seriece-of-event.json | lookout serve -mode cli | grep comment
    ```

PR #57, opened 2018-07-25
Console provider

carlosms:
Part of #40.

This PR adds a new console provider, watcher and poster. It will listen for events in sdtin and output the comments to stdout.

It can be run as:

```bash
go run cmd/lookout/* serve  --provider console repourl
```

The events have to be provided as a line with the format
```
REVIEW/PUSH repoURL baseBranch baseHash headBranch headHash
```
For example:
```
review https://github.com/carlosms/lookout.git master 0ac15b52f3ea4e3f22f347904547c3651dc26f5e console-provider 5ed66cc9ebcc62bed0c58e9b376f4d9baf2122ed
```
```
review file:///home/cmartin/go/src/github.com/src-d/lookout master 0ac15b52f3ea4e3f22f347904547c3651dc26f5e console-provider 5ed66cc9ebcc62bed0c58e9b376f4d9baf2122ed
```

The comments from the dummy analyzer look like this:
```
cmd/lookout/serve.go: The file has increased in 29 lines.
cmd/lookout/serve.go:30: This line exceeded 80 bytes.
cmd/lookout/serve.go:31: This line exceeded 80 bytes.
```

A few notes:

- `lookout serve` requires the positional argument of a repo url, that the new provider ignores. I left it as it is because eventually we'll manage the repos we monitor in a different way (conf file or DB).
- The usage is not as clean as with `lookout review`, where you can just give two nice revision names and it would work, because:
  - We don't know if the repo is a local dir, so here we can't use go-git to get the hash of a revision.
  - `serve` uses `LibraryCommitLoader`, whereas `review` uses `StorerCommitLoader`. For this reason we need to provide both the reference name and the hash.
- Right now you can't trigger an analysis with base and head from two different repositories, so you can't compare a fork with its base repo. This is enforced in the data server side, [loader.go#L43](https://github.com/src-d/lookout/blob/master/service/git/loader.go#L43)

### DIFF
--- a/provider/console/watcher.go
+++ b/provider/console/watcher.go
@@ -1,0 +1,126 @@
+package console
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"io"
+	"strings"
+	"time"
+
+	"github.com/src-d/lookout"
+
+	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-log.v1"
+)
+
+// Provider is the name
+const Provider = "console"
+
+// WatchOptions options to use in the Watcher constructor.
+type WatchOptions struct {
+	Reader io.Reader
+}
+
+// Watcher watches for new events in the console
+type Watcher struct {
+	o       *WatchOptions
+	scanner *bufio.Scanner
+}
+
+// NewWatcher returns a new console watcher
+func NewWatcher(o *WatchOptions) (*Watcher, error) {
+	return &Watcher{
+		o:       o,
+		scanner: bufio.NewScanner(o.Reader),
+	}, nil
+}
+
+// Watch reads from stdin and calls cb for each new event
+func (w *Watcher) Watch(ctx context.Context, cb lookout.EventHandler) error {
+	log.With(log.Fields{"provider": Provider}).Infof("Starting watcher")
+
+	lines := make(chan string, 1)
+	go func() {
+		for w.scanner.Scan() {
+			lines <- w.scanner.Text()
+		}
+	}()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case line := <-lines:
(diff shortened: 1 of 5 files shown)


bzz:
How about the idea of having more structured input as JSON object on every line?
It can consume a `jsonl` format http://jsonlines.org/

Fixtures could be in `.jsonl` file, you can `cat` it well. And if human wants to read it, it can `| jq .` it. And although it's not easy to write in this format- it’s easy to generate from our event structs \w small helper that one can `go run`.

WDYT?

provider/console/watcher.go
```diff
@@ -0,0 +1,126 @@
...
+			ReferenceName:         plumbing.ReferenceName(fmt.Sprintf("refs/heads/%s", toBranch)),
+			Hash:                  toHash,
+		},
+	}
+
+	switch strings.ToLower(cmd) {
+	case "review":
+		event = &lookout.ReviewEvent{
+			Provider:    Provider,
+			InternalID:  "1",
```
bzz: Any reason not to generate a random UUIDs here?

carlosms: If the point is to avoid repetition of the ID, a random ID may repeat itself, and to avoid it we would need to do something too complex for this provider (like keeping track of used IDs, even between restarts when we have a DB).

I'd argue that if repeating the ID is a problem at some point in the future, a fixed "1" will be detected quickly, and at that point we can figure something out.

bzz: Random UUID has very low probability of being repeated and that is something (more or less) that is expected from other providers

carlosms:
As discussed via chat, we are discarding this PR in favor of #61.

provider/console/watcher.go
carlosms: With the more flexible approach in #61 you can set "internal_id" manually for each event, so I think this is no longer an issue.

Issue #45, opened 2018-07-20
SDK: uast handling


smacker:
- [x] implement an optional request for uast in dummy analyzer
- [x] sdk data server should return error if UAST was requested but there is no connection to bblfsh
- [ ] add tests for this cases

PR #59, opened 2018-07-25
return error on WantUAST without bblfsh

smacker:
Part of #45


cmd/lookout/event.go
```diff
@@ -93,6 +94,10 @@ func (c *EventCommand) makeDataServerHandler() (*lookout.DataServerHandler, erro
 	bblfshConn, err := grpc.DialContext(timeoutCtx, c.Bblfshd, grpc.WithInsecure(), grpc.WithBlock())
 	if err != nil {
 		log.Warningf("bblfsh service is unavailable. No UAST will be provided to analyzer. Error: %s", err)
```
bzz: While here, why not improving this error msg as well?

`"bblfshd instance could not be found at %s. No UASTs will be available to analyzers. Error: %s"`

Issue #47, opened 2018-07-20
Logging


smacker:
1. All commands (serve/review/push/dummy) should have `-v` flag.
2. When logging is verbose successful path should be printed. It includes at least `Connected to`, `Send`/`Revieved` event. For servers it should reports when a server has started. Please inspect the code to make find more useful cases.
3. What should we do with grpc logging? There was a suggestion to replace it with our logging: https://github.com/src-d/lookout/pull/37#issuecomment-406548743 @bzz @carlosms need your input
4. We may also want to be able to configure level more flexible. Example: https://github.com/smola/go-cli/pull/1/files#diff-308aaa719425f4333033c97cfb3b9868
5. review/push/(dummy?) should have INFO level by default while serve most probably WARNING. -v flag would make it one level more versbose

bzz:
Proposal looks solid.

> 3. What should we do with grpc logging? There was a suggestion to replace it with our logging: #37 (comment) @bzz @carlosms need your input

I would say custom gPRC logging is important right now for SDK case, and for me suggesting for #37 holds for that case in particular. May be middleware-based logging would allow for exactly this granularity?

For `serve`, I would say if we want to keep current super-verbose logs - we need to add `-vv` and m.b apply  #11 ?

smacker:
For `serve` maybe something similar to configuration from `go-cli` would work?
(actually I would better copy-past than it will be easier to migrate to `go-cli` later)

bzz:
Personally, I'm mostly concerned with SDK usability for now - we should build the trust for the initial users so there is no need for them to double-check networking accessibility and always be able to tell what's going on for review/push.

On `serve` I'm not sure what do you mean with go-cli - could you give an example? `-vv` looks common for CLI tools

smacker:
The example is in the link attached to the issue description 🤔 . Once again: https://github.com/smola/go-cli/pull/1/files#diff-308aaa719425f4333033c97cfb3b9868

For the successful case, the only thing left is grpc logs in verbose mode. For error case, we need to wrap errors in more human-friendly format also. If you feel grcp-middlleware is a good fit for that, let's implement it. (PRs are welcome)

But for flags, env vars, ... I think here is no conclusion, in my opinion. I would prefer to follow our guidelines [for configuration](https://github.com/src-d/guide/blob/master/engineering/conventions.md#configuration) than `-vv` or any other custom flags.

And instead of inventing such configuration I would copy-past it from go-cli to simplify our life in the future.

Quote @smola:
[...]

PR #61, opened 2018-07-25
Add json provider watcher & poster

### LINKED ISSUE #40: CLI input/output for serve

carlosms:
Fixes #40.

As discussed via chat, we are discarding the previous console provider (#57) in favor of this json stdio provider.

It can be run as:

```bash
go run cmd/lookout/* serve  --provider json repourl
```

The events have to be provided as one line json with the same format as protobuf events (reference for [ReviewEvent](https://github.com/src-d/lookout/blob/master/pb/event.pb.go#L103)), plus a field: `"event" :  "review/push"`.

For example, having this content in `events.jsonl`:
```json
{"event":"review", "commit_revision":{"base":{"internal_repository_url":"https://github.com/carlosms/lookout.git","ReferenceName":"refs/heads/master","Hash":"0ac15b52f3ea4e3f22f347904547c3651dc26f5e"},"head":{"internal_repository_url":"https://github.com/carlosms/lookout.git","ReferenceName":"refs/heads/console-provider","Hash":"5ed66cc9ebcc62bed0c58e9b376f4d9baf2122ed"}}}
```

You can run 
```
cat events.jsonl | go run cmd/lookout/* serve  --provider json repourl
{"file":"cmd/lookout/serve.go","text":"The file has increased in 29 lines."}
{"file":"cmd/lookout/serve.go","line":30,"text":"This line exceeded 80 bytes."}
{"file":"cmd/lookout/serve.go","line":31,"text":"This line exceeded 80 bytes."}
{"file":"cmd/lookout/serve.go","line":32,"text":"This line exceeded 80 bytes."}
...
```

A few comments:
- `lookout serve` requires the positional argument of a repo url, that the new provider ignores. I left it as it is because eventually we'll manage the repos we monitor in a different way (conf file or DB).
- When redirecting the stdout (`> out.txt 2> err.txt`) the logs go correctly to stderr, except for gRPC ones. But I think this will be part of #47 and should not be fixed in this PR.
- Here is the previous input json formatted, for convenience. These are the minimum json fields that need to be provided for the dummy analyzer to work.
```json
{
   "commit_revision" : {
      "head" : {
         "Hash" : "5ed66cc9ebcc62bed0c58e9b376f4d9baf2122ed",
         "internal_repository_url" : "https://github.com/carlosms/lookout.git",
         "ReferenceName" : "refs/heads/console-provider"
      },
      "base" : {
         "Hash" : "0ac15b52f3ea4e3f22f347904547c3651dc26f5e",
         "internal_repository_url" : "https://github.com/carlosms/lookout.git",
         "ReferenceName" : "refs/heads/master"
      }
   },
   "event" : "review"
}
```

### DIFF
--- a/pb/event.go
+++ b/pb/event.go
@@ -3,7 +3,9 @@ package pb
 import (
 	"crypto/sha1"
 	"encoding/hex"
+	"errors"
 	"fmt"
+	"reflect"
 
 	"gopkg.in/sourcegraph/go-vcsurl.v1"
 )
@@ -53,6 +55,16 @@ func (e *ReviewEvent) Revision() *CommitRevision {
 	return &e.CommitRevision
 }
 
+// Validate honors the Event interface.
+func (e *ReviewEvent) Validate() error {
+	var zeroVal ReviewEvent
+	if reflect.DeepEqual(*e, zeroVal) {
+		return errors.New("this ReviewEvent event is empty")
+	}
+
+	return nil
+}
+
 // ID honors the Event interface.
 func (e *PushEvent) ID() EventID {
 	return ComputeEventID(e.Provider, e.InternalID)
@@ -66,6 +78,16 @@ func (e *PushEvent) Type() EventType {
 // Revision honors the Event interface.
 func (e *PushEvent) Revision() *CommitRevision {
 	return &e.CommitRevision
+}
+
+// Validate honors the Event interface.
+func (e *PushEvent) Validate() error {
+	var zeroVal PushEvent
+	if reflect.DeepEqual(*e, zeroVal) {
+		return errors.New("this PushEvent event is empty")
+	}
+
+	return nil
 }
 
 type RepositoryInfo = vcsurl.RepoInfo //TODO(mcuadros): improve repository references
--- a/provider/json/watcher.go
+++ b/provider/json/watcher.go
@@ -1,0 +1,104 @@
+package json
+
+import (
+	"bufio"
+	"context"
+	"encoding/json"
+	"io"
+	"strings"
+
+	"github.com/src-d/lookout"
+
+	"gopkg.in/src-d/go-log.v1"
+)
+
+// Provider is the name
+const Provider = "json"
+
+// Watcher watches for new json events in the console
+type Watcher struct {
+	o       *lookout.WatchOptions
+	scanner *bufio.Scanner
+}
+
+// NewWatcher returns a new json console watcher
+func NewWatcher(reader io.Reader, o *lookout.WatchOptions) (*Watcher, error) {
+	return &Watcher{
+		o:       o,
+		scanner: bufio.NewScanner(reader),
+	}, nil
+}
+
+// Watch reads json from stdin and calls cb for each new event
+func (w *Watcher) Watch(ctx context.Context, cb lookout.EventHandler) error {
+	log.With(log.Fields{"provider": Provider}).Infof("Starting watcher")
+
+	lines := make(chan string, 1)
+	go func() {
+		for w.scanner.Scan() {
+			lines <- w.scanner.Text()
+		}
+	}()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case line := <-lines:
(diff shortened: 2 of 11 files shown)


provider/json/watcher.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/src-d/lookout"
+
+	"gopkg.in/src-d/go-log.v1"
+)
+
+// Provider is the name
+const Provider = "json"
+
+// WatchOptions options to use in the Watcher constructor.
+type WatchOptions struct {
```
smacker: Could you please explain me the reason to have this struct?

Github provider accepts `lookout.WatchOptions` which should be common for different providers.
I see currently there is only `URL` prop which doesn't make any sense for console provider.

I would suggest to remove this struct to avoid confusion and change constructor to either:
- `NewWatcher(io.Reader)`
- `NewWatcher(io.Reader, lookout.WatchOptions)`

carlosms: As you said, `lookout.WatchOptions` should be common, but the current `URL` does not apply to this new provider. I just put `WatchOptions` to make it consistent.

If we change this to `NewWatcher(io.Reader)`, do you think we should remove `lookout.WatchOptions` and make `github NewWatcher(transport http.RoundTripper, URL string)`?

smacker: to make it consistent I would do:
```go
func NewWatcher(io.Reader, *lookout.WatchOptions)
```

similar to github
```go
func NewWatcher(http.RoundTripper, *lookout.WatchOptions)
```

The first argument is "the source" (http for github and io for console), the second is common options.

About removing `lookout.WatchOptions` completely:

I would keep it at least for now. I see how it can be useful for other providers (gitlab, bitbucket, ...). Console provider is a way too special.

bzz:
Overall looks great, 👏 ! 

Last thing that seems to be missing is an integration test on CI (in a separate profile).

Something like along #40 description
>  cat seriece-of-event-fixture.json | lookout serve  --provider json <> | grep <>

smacker:
@carlosms I took another look. Do you think we might need some validation for required fields? It looks like now it's possible to pass `{"event": "review"}` json. It will trigger an event but such event would cause panic or weird error somewhere, no?

carlosms:
for the record: we discussed IRL and the agreement is to add a new Validate method to Event interface, and call it from server HandlePush/Review. For now a dummy or basic validation that can be improved later.

pb/event.go
```diff
@@ -53,6 +55,16 @@ func (e *ReviewEvent) Revision() *CommitRevision {
 	return &e.CommitRevision
 }
 
+// Validate honors the Event interface.
+func (e *ReviewEvent) Validate() error {
+	var zeroVal ReviewEvent
+	if reflect.DeepEqual(*e, zeroVal) {
```
smacker: you don't need `reflect.DeepEqual`. `*e == zeroVal` should work just fine

carlosms: I tried that first, but fails with:
`invalid operation: *e == zeroVal (struct containing types.Struct cannot be compared)`

carlosms:
Changes pushed:
- The `NewWatcher` improvement from https://github.com/src-d/lookout/pull/61#discussion_r205201657 is squashed in 4204488
- Event.Validate in 2993d8d
- And an integration test in d500a88

pb/event.go
smacker: wow. 👍

PR #66, opened 2018-07-26
setting log config through env vars

smacker:
Part of #47 

Also renamed `util/flags/ → util/cli/` and make the interface more similar to the one proposed by @smola . But without wrapper for `AddCommand` (there is no agreement on how it should work yet)


util/cli/common_options.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+type initializer interface {
+	init(*App)
+}
+
+// CommonOptions contains common flags for all commands
+type CommonOptions struct {
+	LogOptions
+	GrpcMaxMsgSize int `long:"grpc-max-message-size" default:"100" env:"LOOKOUT_GRPC_MAX_MSG_SIZE" description:"max. message size to send/receive to/from clients (in MB)"`
```
carlosms: nitpick comment: a shorter name would be nicer in my opinion, e.g. `grpc-max-size` and `LOOKOUT_GRPC_MAX_SIZE`

smacker: I took the name from bblfshd. Though I prefer a shorter one too. 👍

bzz: shorter version might be nice, but in this case consistency between project seem more valuable.

Also, there are number of parameters that you can set in gRPC regarding MAX size, and in this case it's only max _message_ size, so it's good

util/cli/log.go
```diff
@@ -0,0 +1,48 @@
...
+type LogOptions struct {
+	LogLevel       string `long:"log-level" env:"LOG_LEVEL" default:"info" description:"Logging level (info, debug, warning or error)"`
+	LogFormat      string `long:"log-format" env:"LOG_FORMAT" description:"log format (text or json), defaults to text on a terminal and json otherwise"`
+	LogFields      string `long:"log-fields" env:"LOG_FIELDS" description:"default fields for the logger, specified in json"`
+	LogForceFormat bool   `long:"log-force-format" env:"LOG_FORCE_FORMAT" description:"ignore if it is running on a terminal or not"`
+	Verbose        bool   `long:"verbose" short:"v" description:"enable verbose logging"`
+}
+
+// Init initializes the default logger factory.
+func (c *LogOptions) init(app *App) error {
```
bzz: Probably a nitpick, but would it make sense to pass only `app.Name` in here?

util/cli/common_options.go
```diff
@@ -0,0 +1,20 @@
+package cli
+
+import (
+	"github.com/src-d/lookout/util/grpchelper"
+)
+
+type initializer interface {
```
bzz: Again, most probably a nitpick, but would `initializable` better represent it's purpose?

Then the client might look something like
```go
if v, ok := command.(initializable); ok {
	v.initialize(app.Name)
}
```

util/cli/log.go
smacker: The API copies go-cli. It definitely makes sense to pas App because it will have more values required here or there, not only name.

util/cli/common_options.go
smacker: https://golang.org/doc/effective_go.html#interface-names

PR #69, opened 2018-07-26
Clarify comments for ReviewEvent.{source,merge}

vmarkovtsev:
After the discussion in the kitchen 26.07

### DIFF
--- a/pb/event.pb.go
+++ b/pb/event.pb.go
@@ -111,10 +111,9 @@ type ReviewEvent struct {
 	UpdatedAt time.Time `protobuf:"bytes,4,opt,name=updated_at,json=updatedAt,stdtime" json:"updated_at"`
 	// IsMergeable, if the pull request is mergeable.
 	IsMergeable bool `protobuf:"varint,5,opt,name=is_mergeable,json=isMergeable,proto3" json:"is_mergeable,omitempty"`
-	// Source reference to the original branch and repository were the changes
-	// came from.
+	// Source reference to the original branch and repository where the changes came from.
 	Source ReferencePointer `protobuf:"bytes,8,opt,name=source" json:"source"`
-	// Merge test branch where the PullRequest was merged.
+	// Merge branch in the destination repository where the Pull Request is stored.
 	Merge ReferencePointer `protobuf:"bytes,9,opt,name=merge" json:"merge"`
 	// Configuration contains any configuration related to specific analyzer
 	Configuration google_protobuf2.Struct `protobuf:"bytes,10,opt,name=configuration" json:"configuration"`
--- a/sdk/event.proto
+++ b/sdk/event.proto
@@ -77,7 +77,7 @@ message ReviewEvent {
     bool is_mergeable = 5;
     // Source reference to the original branch and repository where the changes came from.
     ReferencePointer source = 8 [(gogoproto.nullable) = false];
-    // Merge branch in the destination repository where the Pull Request is stored.
+    // Merge reference to the branch and repository where the merged Pull Request is stored.
     ReferencePointer merge = 9 [(gogoproto.nullable) = false];
     // Configuration contains any configuration related to specific analyzer
     google.protobuf.Struct configuration = 10 [(gogoproto.nullable) = false];
--- a/sdk/event.proto
+++ b/sdk/event.proto
@@ -75,10 +75,9 @@ message ReviewEvent {
     google.protobuf.Timestamp updated_at = 4 [(gogoproto.nullable) = false, (gogoproto.stdtime) = true];
     // IsMergeable, if the pull request is mergeable.
     bool is_mergeable = 5;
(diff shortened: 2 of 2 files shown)


bzz:
@vmarkovtsev As SDK is in the same codebase with the server, and this change does not include re-generated Protobuf code that is vendored - thus it [fails the CI](https://travis-ci.org/src-d/lookout/jobs/408480677).

You would need to run `make protogen` and include the changes in separate commit to pass the CI

vmarkovtsev:
@bzz I have no idea why editing the comments breaks the build.

```
$ make no-changes-in-commit
diff --git a/pb/event.pb.go b/pb/event.pb.go
index 32e8619..c60e19d 100644
--- a/pb/event.pb.go
+++ b/pb/event.pb.go
@@ -68,7 +68,7 @@ type ReferencePointer struct {
 	InternalRepositoryURL string `protobuf:"bytes,1,opt,name=internal_repository_url,json=internalRepositoryUrl,proto3" json:"internal_repository_url,omitempty"`
 	// ReferenceName is the name of the reference pointing.
 	ReferenceName gopkg_in_src_d_go_git_v4_plumbing.ReferenceName `protobuf:"bytes,2,opt,name=ReferenceName,proto3,casttype=gopkg.in/src-d/go-git.v4/plumbing.ReferenceName" json:"ReferenceName,omitempty"`
-	// Hash is the hash of the reference pointing.
+	// hash is the hash of the reference pointing.
 	Hash string `protobuf:"bytes,3,opt,name=Hash,proto3" json:"Hash,omitempty"`
 }
 
@@ -111,10 +111,9 @@ type ReviewEvent struct {
 	UpdatedAt time.Time `protobuf:"bytes,4,opt,name=updated_at,json=updatedAt,stdtime" json:"updated_at"`
 	// IsMergeable, if the pull request is mergeable.
 	IsMergeable bool `protobuf:"varint,5,opt,name=is_mergeable,json=isMergeable,proto3" json:"is_mergeable,omitempty"`
-	// Source reference to the original branch and repository were the changes
-	// came from.
+	// Source reference to the original branch and repository where the changes came from.
 	Source ReferencePointer `protobuf:"bytes,8,opt,name=source" json:"source"`
-	// Merge test branch where the PullRequest was merged.
+	// Merge branch in the destination repository where the Pull Request is stored.
 	Merge ReferencePointer `protobuf:"bytes,9,opt,name=merge" json:"merge"`
 	// Configuration contains any configuration related to specific analyzer
 	Configuration  google_protobuf2.Struct `protobuf:"bytes,10,opt,name=configuration" json:"configuration"`
 M pb/event.pb.go
generated assets are out of sync
make: *** [no-changes-in-commit] Error 2
The command "make no-changes-in-commit" exited with 2.
```

I don't see `no-changes-in-commit` in the top-level Makefile so I am clueless.

smacker:
@vmarkovtsev `no-changes-in-commit` is defined in [.travis.yml](https://github.com/src-d/lookout/blob/master/.travis.yml#L48). For us, it's the easiest way to make sure proto files and code are in sync. (different comments in code & proto means they arent).

To pass CI you need to run `make protogen` and commit changes of go files. If you don't have necessary dependencies for that, please let me know, I'll run it on top of your changes and push to this branch.

smacker:
One more thing. Do you guys need this field? I'm not sure other providers would have it, so maybe we could just remove it?

vmarkovtsev:
We are not using them.

bzz:
+1 for clarification in comments that @smacker proposed. 

But if both GH and Bitbuket do provide it, I would keep it at least for now. 

May be worth to file a separate issue - remove un-used fields?

smacker:
@bzz in gitlab & bitbucket there is a field with the same name but different purpose.

All descriptions:
- Merge test branch where the PullRequest was merged.
- Merge branch in the destination repository where the Pull Request is stored.
- Merge branch that contains a base branch + merge commit

are valid only for github.

For other providers it will be something like:
- Point to merge commit after the PullRequest is successfully merged

That's why I'm raising this question here.

bzz:
Thank you for explanations about different providers!

Yes, that does not make sense to have provider-specific behaviour documented in our API, esp in case it is not beeing used.

Removing it 🙏 using separate PR and 🙏 mentioning it in Release note's change makes much more sense to me as well now.

bzz:
Rebased and code re-generated.

Will merge if there is no further discussion

smacker:
huh? `ReferencePointer` isn't a repository. At least in terms of git.

smacker:
Comment should start with the name if attribute.

I see for source description is
```
// Source reference to the original branch and repository where the changes came from.
```

Maybe we can do similar?
```
// Merge reference to the branch and repository where the merged Pull Request is stored.
```

wdyt?

bzz:
Sounds good, addressed in 1e6e064

PR #70, opened 2018-07-26
add uast option to dummy analyzer

smacker:
Part of #45

### DIFF
--- a/cmd/dummy/main.go
+++ b/cmd/dummy/main.go
@@ -19,8 +19,9 @@ var (
 
 type ServeCommand struct {
 	cli.CommonOptions
-	Analyzer   string `long:"analyzer" default:"ipv4://localhost:10302" env:"LOOKOUT_ANALYZER" description:"gRPC URL to bind the analyzer to"`
-	DataServer string `long:"data-server" default:"ipv4://localhost:10301" env:"LOOKOUT_DATA_SERVER" description:"grPC URL of the data server"`
+	Analyzer    string `long:"analyzer" default:"ipv4://localhost:10302" env:"LOOKOUT_ANALYZER" description:"gRPC URL to bind the analyzer to"`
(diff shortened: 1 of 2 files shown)


dummy/dummy.go
```diff
@@ -94,6 +99,22 @@ func (a *Analyzer) maxLineWidth(ch *lookout.Change) []*lookout.Comment {
 	return comments
 }
 
+func (a *Analyzer) noUAST(ch *lookout.Change) []*lookout.Comment {
```
bzz: This seems like `maybeAddNoUastComment()`

cmd/dummy/main.go
```diff
@@ -19,8 +19,9 @@ var (
 
 type ServeCommand struct {
 	flags.CommonOptions
-	Analyzer   string `long:"analyzer" default:"ipv4://localhost:10302" env:"LOOKOUT_ANALYZER" description:"gRPC URL to bind the analyzer to"`
-	DataServer string `long:"data-server" default:"ipv4://localhost:10301" env:"LOOKOUT_DATA_SERVER" description:"grPC URL of the data server"`
+	Analyzer    string `long:"analyzer" default:"ipv4://localhost:10302" env:"LOOKOUT_ANALYZER" description:"gRPC URL to bind the analyzer to"`
+	DataServer  string `long:"data-server" default:"ipv4://localhost:10301" env:"LOOKOUT_DATA_SERVER" description:"gRPC URL of the data server"`
+	RequestUAST bool   `long:"uast" env:"LOOKOUT_REQUEST_UAST" description:"analyzer will request UAST from the data server"`
```
bzz: `analyzer will request UAST from the data server` -> `always request UAST from the data server` ?

dummy/dummy.go
smacker: the naming is similar to the other methods like `maxLineWidth` or `lineIncrease`. Not every line exceeds max width or a file increase number of lines. The same here. Not every change don't have UAST.

PR #71, opened 2018-07-26
Merge config

smacker:
Fix: #20

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,37 @@ Any of the analyzers or a default dummy one, included in this repository
 1. Create a new pull requires in the repository
 
 
+# Configuration file
+
+Server configuration `config.yml`:
+
+```yml
+analyzers:
+  - name: Example name # required, unique name of the analyzer
+    addr: ipv4://localhost:10302 # required, gRPC address
+    disabled: false # optional, false by default
+    settings: # optional, this field is sent to analyzer "as is"
+        threshold: 0.8
+```
+
+It's possible to override configuration for a repository.
+To do that `.lookout.yml` must be present.
+
+Example:
+```yml
+analyzers:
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -47,6 +47,37 @@ Any of the analyzers or a default dummy one, included in this repository
...
+```yml
+analyzers:
+  - name: Example name # required, unique name of the analyzer
+    addr: ipv4://localhost:10302 # required, gRPC address
+    disabled: false # optional, false by default
+    settings: # optional, this field is sent to analyzer "as is"
+        threshold: 0.8
+```
+
+It's possible to override configuration for a repository.
```
bzz: It's possible to override Analyzers configuration for a particular repository.

README.md
```diff
@@ -47,6 +47,37 @@ Any of the analyzers or a default dummy one, included in this repository
...
+To do that `.lookout.yml` must be present.
```
bzz: To do that `.lookout.yml` must be present in the root of that repository.

README.md
```diff
@@ -47,6 +47,37 @@ Any of the analyzers or a default dummy one, included in this repository
 1. Create a new pull requires in the repository
 
 
+# Configuration file
+
+Server configuration `config.yml`:
```
bzz: > Global server configuration is stored in `config.yml`

We use word global later, so it's nice to introduce it early in here

smacker:
Thank you for the suggestions! Fixed.

I thought to add some kind of integration tests but as you mentioned it might be out of the scope of this PR and I also think it's a little bit premature and would slow down CI.
There are unit tests for enabled/disabled analyzers already. I hope it's enough for now. We can improve testing later.

Issue #76, opened 2018-07-27
Don't fail on broken cache dir


smacker:
Steps to reproduce:

```
rm -rf /tmp/lookout/github.com/src-d/lookout
mkdir -p /tmp/lookout/github.com/src-d/lookout
make dry-run
```

results in:
```
repository does not exist
exit status 1
```

In real life, it happens when the process dies during cloning. After that `lookout serve` doesn't work until the cache dir is deleted manually.

PR #87, opened 2018-07-27
recreate cache on error

smacker:
Fix: #76 

The simplest way to solve the issue just remove and recreate the cache.
I catch only ErrRepositoryNotExists because looks like it's the only relevant error exported by go-git.

Actually, we don't really need mutex because we call `GetFiles()` on every event before sending it to analyzers. But I wouldn't rely on this and better make code thread-safe.

Guys, please validate I didn't mess up with Lock/Unlock.

### DIFF
--- a/service/git/library.go
+++ b/service/git/library.go
@@ -3,6 +3,7 @@ package git
 import (
 	"fmt"
 	"os"
+	"sync"
 
 	"github.com/src-d/lookout"
 
@@ -21,6 +22,7 @@ var (
 
 // Library controls the persistence of multiple git repositories.
 type Library struct {
+	m  sync.Mutex
 	fs billy.Filesystem
 }
 
@@ -47,6 +49,13 @@ func (l *Library) GetOrInit(url *lookout.RepositoryInfo) (
 
 // Init inits a new repository for the given URL.
 func (l *Library) Init(url *lookout.RepositoryInfo) (*git.Repository, error) {
+	l.m.Lock()
+	defer l.m.Unlock()
+
+	return l.init(url)
+}
+
+func (l *Library) init(url *lookout.RepositoryInfo) (*git.Repository, error) {
 	has, err := l.Has(url)
 	if err != nil {
 		return nil, err
@@ -94,6 +103,17 @@ func (l *Library) Has(url *lookout.RepositoryInfo) (bool, error) {
 
 // Get get the requested repository based on the given URL.
 func (l *Library) Get(url *lookout.RepositoryInfo) (*git.Repository, error) {
+	r, err := l.get(url)
+
+	// it can happen if the repository in a broken state
+	if err == git.ErrRepositoryNotExists {
+		return l.recreate(url)
+	}
+
+	return r, nil
+}
+
+func (l *Library) get(url *lookout.RepositoryInfo) (*git.Repository, error) {
 	has, err := l.Has(url)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


service/git/library.go
```diff
@@ -124,3 +144,19 @@ func (l *Library) repositoryStorer(url *lookout.RepositoryInfo) (
 func (l *Library) repositoryPath(url *lookout.RepositoryInfo) string {
 	return fmt.Sprintf("%s/%s", url.RepoHost, url.FullName)
 }
+
+func (l *Library) recreate(url *lookout.RepositoryInfo) (*git.Repository, error) {
```
carlosms: It looks to me like `l.get` should be done inside the lock.
Otherwise you could have:
- 2 goroutines (A, B) check the repo at the same time
- both try to lock
- A gets the lock, inits, returns the repo, unlocks
- B gets the lock, removes the repo, breaks whatever was using A's return value

Or did I misunderstand the code?

smacker: `get` doesn't recreate repo. only `.Get` does.

carlosms:
I have two questions:

1) I don't quite get the `Syncer.Mutex`, please help me to understand it.
It looks like its only purpose is to protect access to `Library.GetOrInit`. But `LibraryCommitLoader.LoadCommits` calls `Library.GetOrInit` without any mutex.
If `Library.GetOrInit` is not concurrent-safe, shouldn't the mutex be managed by `Library` itself?

2) With only one mutex we are blocking all of concurrent operations on different repos. Maybe we could have a `map[string]sync.Mutex` inside `Library`, protected with a global mutex, to store a mutex for each repo url. What do you think?

smacker:
1. We need mutex in 2 cases: init & recreate. If one goroutine started init others should wait. If cache is broken goroutine will delete and recreate it. Others should wait then. I didn't really get the second part of the question. Mutex is a part of Library structure. 
2. It's a good idea for the future. Though I think for now only one mutex is enough. We are going to work only with very small numbers of repos right now and we block only during init.

carlosms:
1) I think we mixed the two mutex. I was asking about the mutex in [syncer.go](https://github.com/src-d/lookout/pull/87/files#diff-6bd7698193142bec9f5cd80af5f41970R28).
I thought it looks redundant because it only protects `s.l.GetOrInit`, but reading more carefully I see it's there to protect `r.FetchContext`, right? So it makes sense :+1:

2) It's true, for `Library` it will only block for new repos and that's not so bad. But the `Syncer.Sync` method will be blocked for every new PR while the new changes are fetched, not just on a new repo init. Or did I get it wrong?
Maybe it's a bit of a premature optimization, but I feel we should try to make the sync as final as possible now, to find and fix any potential problems as early as possible.
Having said that, leaving it as one mutex also looks good to me for now. I'd like to get @smola or @bzz opinion on this.

smacker:
1. Oh sorry. Yes. I was talking about mutex in the library. Mutex in Syncer protects Fetch. (it looks like this method isn't safe to call concurrently, but I might be wrong. That's why I also request a review from @smola)
2. Yeah. Makes perfect sense. Let's wait for confirmation from @smola that we need that mutex and then improve it.

smola:
Indeed, it would be possible to lock only per-repository (with an additional mutex that only locks the per-repository mutex map.

It makes sense that Syncer is also protected with a mutex (at least for the same repository), since git fetch is not thread-safe.

Borges uses a more fine-grained approach here that is fully transactional: https://github.com/src-d/core-retrieval/blob/5252fe35bfba3e0635a06ff5633758b53127c588/repository/repository.go#L20

I think lookout will end up using the same abstraction. In fact, I expect that at some point, once we move this borges functionality to a library, lookout would use it internally too.

We could avoid reimplementing the transactional thing here at the moment.

smacker:
- mutex per repo on Syncer added.
- rebased on master.

smacker:
Comments applied. More fine-grained solution will be implemented when borges lib is ready.
Merge.

Issue #85, opened 2018-07-27
Refactor github provider to use pull requests api


smacker:
I started digging into #77 and found out current /events API isn't suitable for us.

Example of the current state of lookout:
There is a pull request #81. It was re-pushed force after rebase on master. Now let's look at github responses:

Pull requests API:
https://api.github.com/repos/src-d/lookout/pulls
https://gist.github.com/smacker/6378f49c4c6108b50bae2e1041e3fa2f

`[0].head.sha = 'cd15e2c050763e2ac3d444521eb28d06c9f47326'`

Events API:
https://api.github.com/repos/src-d/lookout/events
https://gist.github.com/smacker/1c82e55e001f2e27efdfe0242b33bd9d

`[13].payload.pull_request.head.sha = 'dfa9b65007a3c737d682e561ae2f7a1741d81b2b'`

there is no push event to pull request branch or any other event that updates PR:
```
$ grep "cd15e2c050763e2ac3d444521eb28d06c9f47326" events.json  | wc -l
       0
```

So there is no way to get up-to-date sha for pull request using event api.

bzz:
This is super-strange! 

Why do you think in this case there was no a https://developer.github.com/v3/activity/events/types/#pushevent that could be picked up from Event API as well?

Issue #77, opened 2018-07-27
Discard old events


smacker:
Steps to reproduce:

1. Create PR
2. Amend commit & push --force
3. Run the server on this repository

Results in:
```
[2018-07-27T13:54:10.417860965+02:00]  INFO processing pull request app=lookout head=refs/pull/74/head provider=github repository=git://github.com/src-d/lookout.git
...
[2018-07-27T13:54:11.32091199+02:00]  INFO processing pull request app=lookout head=refs/pull/74/head provider=github repository=git://github.com/src-d/lookout.git
object not found
exit status 1
```

This error may also happen if `push --force` was done during events processing.

PR #88, opened 2018-07-30
Refactor github provider to use PRs API

smacker:
Fix: #85

### DIFF
--- a/provider/github/utils.go
+++ b/provider/github/utils.go
@@ -64,7 +64,7 @@ func castHash(sha1 *string) plumbing.Hash {
 	return plumbing.NewHash(*sha1)
 }
 
-func castPullRequestEvent(r *lookout.RepositoryInfo, pr *github.PullRequest) *lookout.ReviewEvent {
+func castPullRequest(r *lookout.RepositoryInfo, pr *github.PullRequest) *lookout.ReviewEvent {
 	pre := &lookout.ReviewEvent{}
 	pre.Provider = Provider
 	pre.InternalID = string(pr.GetID())
--- a/provider/github/utils.go
+++ b/provider/github/utils.go
@@ -20,13 +20,6 @@ func castEvent(r *lookout.RepositoryInfo, e *github.Event) (lookout.Event, error
 		}
 
 		return castPushEvent(r, e, payload.(*github.PushEvent)), nil
-	case "PullRequestEvent":
-		payload, err := e.ParsePayload()
-		if err != nil {
-			return nil, ErrParsingEventPayload.New(err)
-		}
-
-		return castPullRequestEvent(r, e, payload.(*github.PullRequestEvent)), nil
 	}
 
 	return nil, nil
@@ -71,36 +64,27 @@ func castHash(sha1 *string) plumbing.Hash {
 	return plumbing.NewHash(*sha1)
 }
 
-func castPullRequestEvent(
-	r *lookout.RepositoryInfo,
-	e *github.Event, pr *github.PullRequestEvent,
-) *lookout.ReviewEvent {
-
-	if pr.PullRequest == nil && pr.PullRequest.GetID() != 0 {
-		log.Warningf("missing pull request information in pull request event")
-		return nil
-	}
-
+func castPullRequestEvent(r *lookout.RepositoryInfo, pr *github.PullRequest) *lookout.ReviewEvent {
 	pre := &lookout.ReviewEvent{}
 	pre.Provider = Provider
-	pre.InternalID = e.GetID()
-	pre.Source = castPullRequestBranch(pr.PullRequest.GetHead())
+	pre.InternalID = string(pr.GetID())
+	pre.Source = castPullRequestBranch(pr.GetHead())
 	pre.Merge = lookout.ReferencePointer{
 		InternalRepositoryURL: r.CloneURL,
-		ReferenceName:         plumbing.ReferenceName(fmt.Sprintf("refs/pull/%d/merge", pr.PullRequest.GetNumber())),
-		Hash:                  pr.PullRequest.GetMergeCommitSHA(),
+		ReferenceName:         plumbing.ReferenceName(fmt.Sprintf("refs/pull/%d/merge", pr.GetNumber())),
+		Hash:                  pr.GetMergeCommitSHA(),
 	}
 
-	pre.Base = castPullRequestBranch(pr.PullRequest.GetBase())
+	pre.Base = castPullRequestBranch(pr.GetBase())
 	pre.Head = lookout.ReferencePointer{
 		InternalRepositoryURL: r.CloneURL,
-		ReferenceName:         plumbing.ReferenceName(fmt.Sprintf("refs/pull/%d/head", pr.PullRequest.GetNumber())),
-		Hash:                  pr.PullRequest.GetHead().GetSHA(),
+		ReferenceName:         plumbing.ReferenceName(fmt.Sprintf("refs/pull/%d/head", pr.GetNumber())),
+		Hash:                  pr.GetHead().GetSHA(),
 	}
 
-	pre.IsMergeable = pr.PullRequest.GetMergeable()
+	pre.IsMergeable = pr.GetMergeable()
 
-	pr.PullRequest.GetHead().GetRepo().GetURL()
+	pr.GetHead().GetRepo().GetURL()
 
 	return pre
 }
--- a/provider/github/utils.go
+++ b/provider/github/utils.go
@@ -84,8 +84,6 @@ func castPullRequestEvent(r *lookout.RepositoryInfo, pr *github.PullRequest) *lo
 
 	pre.IsMergeable = pr.GetMergeable()
 
-	pr.GetHead().GetRepo().GetURL()
-
 	return pre
 }
 
--- a/provider/github/watcher.go
+++ b/provider/github/watcher.go
@@ -39,7 +39,7 @@ type Watcher struct {
 	cache *cache.ValidableCache
 
 	// delay for pull requests
-	prPullInterval time.Duration
+	prPollInterval time.Duration
 	// delay is time in seconds to wait between requests for events
 	eventsPollInterval time.Duration
 }
@@ -70,7 +70,7 @@ func NewWatcher(transport http.RoundTripper, o *lookout.WatchOptions) (*Watcher,
 			Transport: t,
 		}),
 
-		prPullInterval:     minInterval,
+		prPollInterval:     minInterval,
 		eventsPollInterval: minInterval,
 
 		cache: cache,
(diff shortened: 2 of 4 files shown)


provider/github/watcher.go
```diff
@@ -132,6 +175,29 @@ func (w *Watcher) handleEvent(r *lookout.RepositoryInfo, e *github.Event) (looko
...
+) {
+	ctx, cancel := context.WithTimeout(ctx, RequestTimeout)
+	defer cancel()
+
+	prs, resp, err := w.c.PullRequests.List(ctx, username, repository, &github.PullRequestListOptions{})
+	if err != nil {
+		return resp, nil, err
+	}
+
+	w.updatePollInterval(resp)
```
carlosms: It looks like ' X-Poll-Interval'  is only returned by [events](https://developer.github.com/v3/activity/events/), and the [pulls](https://developer.github.com/v3/pulls) response does not include it.

provider/github/utils.go
```diff
@@ -71,36 +64,27 @@ func castHash(sha1 *string) plumbing.Hash {
...
-		Hash:                  pr.PullRequest.GetHead().GetSHA(),
+		ReferenceName:         plumbing.ReferenceName(fmt.Sprintf("refs/pull/%d/head", pr.GetNumber())),
+		Hash:                  pr.GetHead().GetSHA(),
 	}
 
-	pre.IsMergeable = pr.PullRequest.GetMergeable()
+	pre.IsMergeable = pr.GetMergeable()
 
-	pr.PullRequest.GetHead().GetRepo().GetURL()
+	pr.GetHead().GetRepo().GetURL()
```
carlosms: Is there something missing here?

carlosms:
I may need some more context on this.

From the issue description, I understand that the bug is only present when a force-push is done on a PR branch. `/events` will not report a PushEvent, and we need to use `/pulls` to get the updated info on that force-pushed PR branch. Is that right?

But if the `/pulls` response already contains all the PRs with updated info, including any push or force-push, why do we also need the `/events` call? Isn't it going to report redundant PushEvents?

smacker:
We use `/pulls` to get correct info about pull requests, yes. No matter was they pushed forced or not.

`/events` returns pushes to master branch. Analyzers might retrain model or do whatever else when they receive it.

Does it make sense?

carlosms:
A few more details on the issue, to help future-me:

Using `/events` (as in master) will never return a `PullRequestEvent` with new commits pushed to a PR branch, no matter if it's forced or not. The event will contain the first group of existing commits when the PR was created.

> We use `/pulls` to get correct info about pull requests, yes. No matter was they pushed forced or not.
> 
> `/events` returns pushes to master branch. Analyzers might retrain model or do whatever else when they receive it.
> 
> Does it make sense?

Now the confusing thing is that `PushEvent`s from `/events` will trigger a `NotifyPushEvent` call, that we define as an opportunity for the analyzer to get the new code and maybe re-train. But we also define that the analyzers are supposed to return a list of comments in [NotifyPushEvent](https://github.com/src-d/lookout/blob/master/sdk/service_analyzer.proto)! That doesn't add up.
And even if the analyzers return any comments, the github poster will [never post them for a push event](https://github.com/src-d/lookout/blob/master/provider/github/poster.go#L58).

So now what I'm thinking is: should we remove the comments from the `NotifyPushEvent` response?

smacker:
@carlosms I have reworked the PR almost completely and pushed force because it's better to review from scratch again. Please another pass.

provider/github/utils.go
```diff
@@ -71,36 +64,25 @@ func castHash(sha1 *string) plumbing.Hash {
...
-	r *lookout.RepositoryInfo,
-	e *github.Event, pr *github.PullRequestEvent,
-) *lookout.ReviewEvent {
-
-	if pr.PullRequest == nil && pr.PullRequest.GetID() != 0 {
-		log.Warningf("missing pull request information in pull request event")
-		return nil
-	}
-
+func castPullRequestEvent(r *lookout.RepositoryInfo, pr *github.PullRequest) *lookout.ReviewEvent {
```
carlosms: I'd rename this to `castPullRequest`, since now it takes a `PullRequest` instead of a `PullRequestEvent`

provider/github/watcher.go
```diff
@@ -35,8 +38,10 @@ type Watcher struct {
 
 	cache *cache.ValidableCache
 
-	// delay is time in seconds to wait between requests
-	pollInterval time.Duration
+	// delay for pull requests
+	prPullInterval time.Duration
```
carlosms: typo? I think it should be `prPollInterval`

provider/github/watcher.go
```diff
@@ -73,34 +81,94 @@ func NewWatcher(transport http.RoundTripper, o *lookout.WatchOptions) (*Watcher,
...
-		repoIndex = (repoIndex + 1) % len(w.r)
+		resp, prs, err := w.doPRListRequest(ctx, repo.Username, repo.Name)
+		if err != nil && !NoErrNotModified.Is(err) {
+			errCh <- err
+			return
+		}
+
+		if err := w.handlePrs(cb, repo, resp, prs); err != nil {
+			if lookout.NoErrStopWatcher.Is(err) {
+				return
```
carlosms: NoErrStopWatcher comment says it should stop the main `.Watch` method, but with this return we are only stopping this goroutine while all the rest will continue. Right?

```
// NoErrStopWatcher if a new error of this kind is returned by EventHandler
// the Watcher.Watch function exits without error.
```

provider/github/watcher.go
```diff
@@ -73,34 +81,94 @@ func NewWatcher(transport http.RoundTripper, o *lookout.WatchOptions) (*Watcher,
...
+			errCh <- err
+			return
 		}
 
 		select {
 		case <-ctx.Done():
-			return ctx.Err()
-		case <-time.After(w.pollInterval):
+			return
+		case <-time.After(w.eventsPollInterval):
```
carlosms: :thinking: I believe this will not obey the `X-Poll-Interval` header. `watchEvents` is run as a goroutine for each repo, and this `for` will wait the interval between each call _for the same repo_. But calls to different repos are actually done concurrently, without any interval between them.

provider/github/watcher.go
smacker: Oh! Thank you! Very good catch!

provider/github/watcher.go
smacker: Hm... I don't see in the documentation if the `X-Poll-Interval` is for any `/events` or per repo. When I was writing this code I assumed it's per repo. Though now I think you are right.

smacker:
Fixed `NoErrStopWatcher` and `X-Poll-Interval`.

provider/github/watcher.go
```diff
@@ -146,23 +234,44 @@ func (w *Watcher) doEventRequest(ctx context.Context, username, repository strin
...
-		"events":             len(events),
-	}).Debugf("Request to events endpoint done.")
+	w.responseLogger(resp).With(log.Fields{"poll-interval": w.eventsPollInterval}).
+		Debugf("Request to events endpoint done with %d events.", len(events))
 
 	return resp, events, err
 }
 
+func (w *Watcher) newInterval(resp *github.Response) time.Duration {
+	remaining := resp.Rate.Remaining / 2 // we call 2 endpoints for each repo
```
carlosms: This is saying that half of the remaining rate is allocated to `/events` and half to `/pulls`.
But they have different intervals, so this is a bit odd.

For example, a fresh start will have a rate 5000 requests for the next hour. And an `X-Poll-Interval` of 60 seconds.
This line allocates 2500 requests to `/events` for the next hour, but we won't make more than 60.

carlosms: What about requests made to post comments, will that decrease the remaining rate? Should we accommodate some margin here for requests that are to polling?

smacker: Any other calls will decrease remaining. What margin do you want to add?

carlosms: What happens when remaining reaches 0? We should wait until `resp.Rate.Reset`, but it looks like that won't be the case in this method.

carlosms: :thinking:  I'm not sure how to solve it, but I'm a bit concerned with the last few calls before `Reset`. We are adjusting the polling without accounting for any other calls, so the last calls to post comments may fail hitting our rate limit.

smacker: Ah. I see what you mean now!
Though I can't predict how many comments will be posted.
The poster must handle rate limit correctly. 

I don't think solving this problem is within the scope of this PR.
I also don't think we should fix it right now. I would better wait for the multi-token support first.
As a quick workaround we can always change `minInterval` from 1 second to 2,3,4,5 and it will give us extra calls.

smacker: We will try to pull every `minInterval`. Github will start rejecting our calls. For the scope of this PR it's okay I think. But I need a check for zero 👍

smacker:
Comment got collapsed:

> I'm not sure how to solve it, but I'm a bit concerned with the last few calls before Reset. We are adjusting the polling without accounting for any other calls, so the last calls to post comments may fail hitting our rate limit.

Ah. I see what you mean now!
Though I can't predict how many comments will be posted.
The poster must handle rate limit correctly.

I don't think solving this problem is within the scope of this PR.
I also don't think we should fix it right now. I would better wait for the multi-token support first.
As a quick workaround we can always change minInterval from 1 second to 2,3,4,5 and it will give us extra calls.

carlosms:
Yes, just to be safe, I would change the min interval to 2 seconds.
With 2 concurrent loops, a min interval of 1s, we are doing at maximum 2 * 60 * 60=7200 polling requests/hr. Which would be adjusted to 5000 req/hr, but we risk running out of requests to post comments.

Changing it to 2s we do max 3600 poll requests/hr, well below the 5000 req/hr limit.
This way we avoid dealing with weird bugs when we reach the rate limit.

smacker:
Updated. Are we good to merge now? @carlosms

bzz:
@carlosms thank you very much for extensive comments with context!

Please, correct me if I'm wrong, but this change affects the number of GH API requests that lookout makes: 
 - from `O(1)` before (every N sec, or poll interval)
 - to `O(number of repositories)`

would that be correct?

Also a bit confused, as from what I understood, fixing https://github.com/src-d/lookout/issues/77 seems to belong to Phase3.

smacker:
It is not correct.
Before it was calling `{repo}/events` every N sec. With `{repo}` in cycle.
Now it is calling `{repo}/events` and `{repo}/prs` (in parallel) every N sec. With `{repo}` in cycle.
Though #109 changes it.

#77 might belong to Phase 3. But #85 is needed ASAP. Otherwise we can't proccess any repushed PR.

bzz:
👍 thanks for explanation!

> Otherwise we can't proccess any repushed PR.

does not matter much, but it still seems to me to be a part of Phase3, from the current design doc

smacker:
> does not matter much

Imo it does mater.
- Lookout would fail on half of _my_ PRs to lookout.
- It's also major change that is better to incorporate on a early stage.

PR #90, opened 2018-07-30
Report GitHub status

### LINKED ISSUE #42: Add option to github provider to report status
With `repo:status` OAuth scope one can report status to Github using [Status API](https://developer.github.com/v3/repos/statuses/)

This issue is about adding an option to enable reporting in-progress status of the code analysis by lookout to PR using `state:pending` of GH status API. 

Most probably this should be specific to a provider, as i.e in GitLab similar thing is related to [Pipelines](https://docs.gitlab.com/ee/api/pipelines.html)

Initial implementation should only report a global status of PR: it's pending, if it's beeing process by any of the analysers in lookout.

carlosms:
Fixes #42.

Using the [status API](https://developer.github.com/v3/repos/statuses/) the server sets the status to _pending_, and then _success_ when the comment posting finishes.

The states follow the GitHub names since for now it's our reference provider. But I think they are flexible enough to fit into new providers. For example for [GitLab pilelines](https://docs.gitlab.com/ee/api/pipelines.html) we have:

| GitHub | Gitlab |
| -- | -- |
| error | canceled |
| failure | failed |
| pending | pending |
| success | success |

### DIFF
--- a/poster.go
+++ b/poster.go
@@ -2,8 +2,36 @@ package lookout
 
 import "context"
 
+// AnalysisStatus is the status reported to the provider to
+// inform that we are performing an analysis, or that it has finished
+type AnalysisStatus int
+
+const (
+	_ AnalysisStatus = iota
+	// ErrorAnalysisStatus represents an error status
+	ErrorAnalysisStatus
+	// FailureAnalysisStatus represents a failure status
+	FailureAnalysisStatus
+	// PendingAnalysisStatus represents a pending status
+	PendingAnalysisStatus
+	// SuccessAnalysisStatus represents a success status
+	SuccessAnalysisStatus
+)
+
+func (st AnalysisStatus) String() string {
+	names := [...]string{"unknown", "error", "failure", "pending", "success"}
+	if st < ErrorAnalysisStatus || st > SuccessAnalysisStatus {
+		return names[0]
+	}
+
+	return names[st]
+}
+
 // Poster can post comments about an event.
 type Poster interface {
 	// Post posts comments about an event.
 	Post(context.Context, Event, []*Comment) error
+
+	// Status sends the current analysis status to the provider
+	Status(context.Context, Event, AnalysisStatus) error
 }
--- a/provider/github/poster.go
+++ b/provider/github/poster.go
@@ -20,10 +20,16 @@ var (
 	ErrEventNotSupported = errors.NewKind("event not supported")
 )
 
+const (
+	statusTargetURL = "https://github.com/src-d/lookout"
+	statusContext   = "lookout"
+)
+
 // Poster posts comments as Pull Request Reviews.
 type Poster struct {
 	rc ReviewCreator
 	cc CommitsComparator
+	sc StatusCreator
 }
 
 var _ lookout.Poster = &Poster{}
@@ -38,6 +44,7 @@ func NewPoster(t http.RoundTripper) *Poster {
 	return &Poster{
 		rc: ghClient.PullRequests,
 		cc: ghClient.Repositories,
+		sc: ghClient.Repositories,
 	}
 }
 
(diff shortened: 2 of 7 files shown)


poster.go
```diff
@@ -2,8 +2,36 @@ package lookout
 
 import "context"
 
+// AnalysisStatus is the status reported to the provider to
+// inform that we are performing an analysis, or that it has finished
+type AnalysisStatus int
+
+const (
+	_ AnalysisStatus = iota
+	// ErrorAnalysisStatus represents an error status
```
smacker: could you explain what is the difference between Error and Failure in our case?

smacker: ah. I see. Failure is when we added some comments. But do we really want to show red cross?

provider/github/poster.go
```diff
@@ -192,3 +194,68 @@ func createReviewRequest(
...
+	owner, repo, _, err := p.validatePR(e)
+	if err != nil {
+		return err
+	}
+
+	statusStr, description, err := statusStrings(status)
+	if err != nil {
+		return err
+	}
+	targetURL := "https://github.com/src-d/lookout"
```
smacker: hard code?

poster.go
carlosms: The difference is not so obvious here, but in `github.Provider` you can see the descriptions. Error would be an internal error that made the processing fail; and failure is a normal processing that yielded a negative result (like a failing CI build)

```
	case lookout.ErrorAnalysisStatus:
		return "error", "There was an error during the analysis", nil
	case lookout.FailureAnalysisStatus:
		return "failure", "The analysis result was negative", nil
```

provider/github/poster.go
carlosms: For now I didn't see it clearly if we should put it in an env var, CLI argument, or field in the conf... so I left it like that as a placeholder.

I think this field is optional, it's the link for "Details" in the PR check list. If you think the hardcoded value looks too bad we can simply leave it unset, and add it later when we have a proper UI to link.

poster.go
carlosms: Also, please note that the PR code only uses _pending_ and _success_. _failure_ may be used at some point in the future, but not for now.

server.go
```diff
@@ -124,6 +128,8 @@ func (s *Server) HandlePush(ctx context.Context, e *PushEvent) error {
 		return err
 	}
 
+	s.poster.Status(ctx, e, PendingAnalysisStatus)
```
smacker: I saw above you post status only for `lookout.ReviewEvent`. But this is push event. Do you think it makes sense to call this method if we know it will always fail?

server.go
```diff
@@ -88,6 +88,8 @@ func (s *Server) HandleReview(ctx context.Context, e *ReviewEvent) error {
 		return nil
 	}
 
+	s.poster.Status(ctx, e, PendingAnalysisStatus)
```
smacker: maybe we should log error here?

server.go
```diff
@@ -102,6 +104,8 @@ func (s *Server) HandleReview(ctx context.Context, e *ReviewEvent) error {
 	comments := s.concurrentRequest(ctx, logger, conf, send)
 
 	s.post(ctx, logger, e, comments)
+	s.poster.Status(ctx, e, SuccessAnalysisStatus)
```
smacker: and here

poster.go
smacker: Thank you very much for the explanation. Yeah, it's a good idea to implement all statuses even if don't use them right now. 👍

provider/github/poster.go
smacker: yeah. I got it's a link. I'm okay to keep hard-coded value for now. But maybe we should move it to a constant on the top of the file so it will be easier to see/find later? I would do the same for `context` variable.

server.go
carlosms: yes, I forgot about it. Fixed in 5c32b9f

provider/github/poster.go
carlosms: Done in 3318d22.

server.go
carlosms: Actually I think it doesn't make sense, but that's a conversation I want to open later.

The `s.post` below will also fail always for a push, so I think we should leave the `status` call also, and handle both things better later.

PR #93, opened 2018-07-30
Uast tests

smacker:
Fix: #45 

I think it's easier to write such tests in go than in shell scripts.

Docker for example also uses go for integration testing:
https://github.com/moby/moby/tree/master/integration-cli

There is no docker in macOS image in travis. I have removed that section completely. I'm not sure we really need to test it 2 times. Go suppose to work cross platform... But let me know if you think otherwise. We can try to add a simpler test for macOs that doesn't require docker.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,8 @@ jobs:
       name: "Unit Tests"
       script: make test-coverage codecov
     - script:
-        - docker run -d --privileged -v $HOME/bblfshd:/var/lib/bblfshd -p "9432:9432" bblfsh/bblfshd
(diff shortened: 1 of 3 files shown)


carlosms:
It's a shame we loose the tests for macOS, but if we can't run the bblfsh docker in travis there is not much we can do.

The code looks good. Is there a reason why you didn't use the testing & testify packages?

smacker:
I thought that testing/testify wouldn't help much in this case. We need to print custom reports, not stacktraces.

Though now when I'm re-reading the code maybe it makes sense. But I'm still not sure. I see only one place where we can benefit from testing lib, it's `Teardown`. Do you see any others?

smacker:
discussed on slack, it's good enough for now. We can start using testing package if we need it later. Merge!

Issue #89, opened 2018-07-30
Document how sdk/lookout binary works


smacker:
Describe the process and flags of the command.

PR #94, opened 2018-07-31
Improve SDK docs

carlosms:
Fix: #89.

cc @vmarkovtsev

### DIFF
--- a/sdk/README.md
+++ b/sdk/README.md
@@ -1,47 +1,149 @@
 # Lookout Analyzer SDK
 
-An analyzer is a gRPC server, that implements [Analyzer service](https://github.com/src-d/lookout/tree/master/sdk/service_analyzer.proto) to receive events from the server.
+An analyzer is a [gRPC](https://grpc.io/) server that implements the [Analyzer service](./service_analyzer.proto) to receive events from the lookout server.
 
-An analyzer should use gRPC client to access [Data service](https://github.com/src-d/lookout/tree/master/sdk/service_data.proto) to get content or UAST of the changes.
+At the same time, an analyzer acts as a gRPC client to access the [Data Server](./service_data.proto) exposed by the lookout server. This endpoint will provide the necessary source code and [Babelfish Universal Abstract Syntax Trees (UAST)](https://doc.bblf.sh/uast/uast-specification.html). See `LOOKOUT_DATA_SERVER` environment variable below for more details.
 
-All `.proto` files are located in [/sdk](https://github.com/src-d/lookout/tree/master/sdk) directory.
-
+All the needed `.proto` files are located in the top-level [/sdk directory](./).
 
 ## Code generation
 
+gPRC communication is performed using protocol buffers. You can create a new analyzer in any language that supports protocol buffers, generating code from the `.proto` definitions. The resulting code will provide data access classes, with accessors for each field, as well as methods to serialize/parse the message structures to/from bytes.
+
+Please refer to the [official Protocol Buffers documentation](https://developers.google.com/protocol-buffers/) to learn how to get started.
+
 ### Python
 
-Dependencies:
+For convenience we are including here a simplified example of how to generate Python code.
+You have more detailed instructions in the [official gRPC Python Quickstart guide](https://grpc.io/docs/quickstart/python.html).
+
+First install the dependencies:
 
 ```bash
 $ pip install grpcio-tools
 ```
 
-Read more about gRPC in [Python Quickstart](https://grpc.io/docs/quickstart/python.html).
-
-Generation:
+Now clone this repository and run this command from the root directory:
 
 ```bash
 $ PY_OUT_DIR=<directory for generated files> mkdir -p $PY_OUT_DIR && \
     python -m grpc_tools.protoc -Isdk \
-    --python_out=$PY_OUT_DIR --grpc_python_out=$PY_OUT_DIR \
-    sdk/*.proto
+    --python_out="$PY_OUT_DIR" --grpc_python_out="$PY_OUT_DIR" \
+    $(find sdk -name *.proto)
 ```
 
-## Testing
+_Note_ Currently there are conflicts when the generated code and the `bblfsh` Python package are used. See [this file](https://github.com/src-d/style-analyzer/blob/master/lookout/core/generate_rpc.sh) for a workaround.
 
-Download the latest lookout binary from [releases](https://github.com/src-d/lookout/releases) page.
+## Developing an Analyzer
 
-Babelfish server is required. If you don't have it running, please read the [getting started guide](https://doc.bblf.sh/using-babelfish/getting-started.html), to learn more about how to start and use it.
+### Requirements
 
-Run the binary inside a git repository directory with gRPC address of the analyzer
+Before getting started you will need the latest lookout binary. You can download it from the [releases page](https://github.com/src-d/lookout/releases).
+
+If your analyzer makes use of UAST, you will need a [Babelfish server](https://doc.bblf.sh/using-babelfish/getting-started.html) running.
+To start it using [Docker Compose](https://docs.docker.com/compose/) clone this repository, or download [`/docker-compose.yml`](/docker-compose.yml), and run:
+
+```bash
+$ docker-compose up
+```
+
+This will create the following Docker containers:
+
+* [bblfshd](https://github.com/bblfsh/bblfshd), listening on `localhost:9432`
+* [PostgreSQL](https://www.postgresql.org/), listening on `localhost:5432`, password `example`. _Note_: this will be a requirement for future functionality, it is not used right now.
+
+
+### lookout SDK Commands
+
+The lookout binary has the following sub-commands:
+- `review`: Performs a `NotifyReviewEvent` call to the analyzer and serves the Data Server endpoint.
+- `push`: Performs a `NotifyPushEvent` call to the analyzer and serves the Data Server endpoint.
+- `serve`: This is the main lookout server execution mode. In the context of the SDK it can be ignored.
+
+Both `lookout push` and `lookout review` can be configured with environment variables or CLI arguments:
+
+| Env var | Option | Description | Default |
+| -- | -- | -- | -- |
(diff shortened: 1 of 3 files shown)


sdk/README.md
```diff
@@ -1,25 +1,29 @@
 # Lookout Analyzer SDK
 
-An analyzer is a gRPC server, that implements [Analyzer service](https://github.com/src-d/lookout/tree/master/sdk/service_analyzer.proto) to receive events from the server.
+An analyzer is a [gRPC](https://grpc.io/) server that implements the [protocol buffers Analyzer service](./service_analyzer.proto) to receive events from the lookout server.
 
-An analyzer should use gRPC client to access [Data service](https://github.com/src-d/lookout/tree/master/sdk/service_data.proto) to get content or UAST of the changes.
+At the same time, an analyzer will act as a a gRPC client to access the [Data Server](./service_data.proto) exposed by the lookout server. This endpoint will provide the necessary source code and [Babelfish Universal Abstract Syntax Trees (UAST)](https://doc.bblf.sh/uast/uast-specification.html).
```
smacker: > will act as a a

one extra "a" :-D

sdk/README.md
```diff
@@ -28,20 +32,106 @@ $ PY_OUT_DIR=<directory for generated files> mkdir -p $PY_OUT_DIR && \
...
+
+Before getting started you will need:
+
+- The latest lookout binary. You can download it from [releases page](https://github.com/src-d/lookout/releases).
+- If your analyzer makes use of UAST, you will need a Babelfish server running. To learn how to start and use it please refer to the [Babelfish getting started guide](https://doc.bblf.sh/using-babelfish/getting-started.html).
+
+### lookout SDK Commands
+
+The lookout binary has the following sub-commands:
+- `push`: Performs a `NotifyReviewEvent` call to the analyzer and serves the Data Server endpoint.
```
smacker: review & push are mixed up

smacker:
awesome

but let's wait for comments from Vadim as the main user of our sdk.

vmarkovtsev:
I am here, will review in a few hours

carlosms:
I added a new commit mentioning docker-compose, somehow I forgot we had that for the requirements.

sdk/README.md
```diff
@@ -1,25 +1,29 @@
 # Lookout Analyzer SDK
 
-An analyzer is a gRPC server, that implements [Analyzer service](https://github.com/src-d/lookout/tree/master/sdk/service_analyzer.proto) to receive events from the server.
+An analyzer is a [gRPC](https://grpc.io/) server that implements the [protocol buffers Analyzer service](./service_analyzer.proto) to receive events from the lookout server.
 
-An analyzer should use gRPC client to access [Data service](https://github.com/src-d/lookout/tree/master/sdk/service_data.proto) to get content or UAST of the changes.
+At the same time, an analyzer will act as a gRPC client to access the [Data Server](./service_data.proto) exposed by the lookout server. This endpoint will provide the necessary source code and [Babelfish Universal Abstract Syntax Trees (UAST)](https://doc.bblf.sh/uast/uast-specification.html).
```
vmarkovtsev: analyzer ~will act~ acts

sdk/README.md
```diff
@@ -1,25 +1,29 @@
 # Lookout Analyzer SDK
 
-An analyzer is a gRPC server, that implements [Analyzer service](https://github.com/src-d/lookout/tree/master/sdk/service_analyzer.proto) to receive events from the server.
+An analyzer is a [gRPC](https://grpc.io/) server that implements the [protocol buffers Analyzer service](./service_analyzer.proto) to receive events from the lookout server.
```
vmarkovtsev: I think that after gRPC implies protocol buffers so we can remove "protocol buffers"

sdk/README.md
vmarkovtsev: How about saying about the different port here?

sdk/README.md
```diff
@@ -1,25 +1,29 @@
...
 
-An analyzer should use gRPC client to access [Data service](https://github.com/src-d/lookout/tree/master/sdk/service_data.proto) to get content or UAST of the changes.
+At the same time, an analyzer will act as a gRPC client to access the [Data Server](./service_data.proto) exposed by the lookout server. This endpoint will provide the necessary source code and [Babelfish Universal Abstract Syntax Trees (UAST)](https://doc.bblf.sh/uast/uast-specification.html).
 
-All `.proto` files are located in [/sdk](https://github.com/src-d/lookout/tree/master/sdk) directory.
+All the needed `.proto` files are located in the current [/sdk directory](./).
```
vmarkovtsev: current -> top-level

sdk/README.md
```diff
@@ -1,25 +1,29 @@
...
 
 
 ## Code generation
 
+You can create a new analyzer in any language that supports protocol buffers.
```
vmarkovtsev: We should say a few words why we need code generation at all.

gRPC communication is performed in protocol buffers format, and it is possible to generate the source code to send and receive messages. Blah-blah-blah

sdk/README.md
```diff
@@ -1,25 +1,29 @@
...
+Please refer to the [official Protocol Buffers documentation](https://developers.google.com/protocol-buffers/) to learn how to get started.
+
 ### Python
 
-Dependencies:
+For convenience we are including here a simplified example on how to generate Python code.
```
vmarkovtsev: example ~on~ of

sdk/README.md
```diff
@@ -1,25 +1,29 @@
...
+First install the dependencies:
 
 ```bash
 $ pip install grpcio-tools
 ```
 
-Read more about gRPC in [Python Quickstart](https://grpc.io/docs/quickstart/python.html).
-
-Generation:
+Now clone this repository and run this command from the root directory:
```
vmarkovtsev: This example actually does not work because we need to generate two dependencies - without them nothing works. See https://github.com/src-d/style-analyzer/blob/master/lookout/core/generate_rpc.sh#L10

sdk/README.md
```diff
@@ -28,20 +32,116 @@ $ PY_OUT_DIR=<directory for generated files> mkdir -p $PY_OUT_DIR && \
...
 ```
 
-## Testing
+## Developing an Analyzer
 
-Download the latest lookout binary from [releases](https://github.com/src-d/lookout/releases) page.
+### Requirements
 
-Babelfish server is required. If you don't have it running, please read the [getting started guide](https://doc.bblf.sh/using-babelfish/getting-started.html), to learn more about how to start and use it.
+Before getting started you will need the latest lookout binary. You can download it from [releases page](https://github.com/src-d/lookout/releases).
```
vmarkovtsev: **the** releases page

sdk/README.md
```diff
@@ -28,20 +32,116 @@ $ PY_OUT_DIR=<directory for generated files> mkdir -p $PY_OUT_DIR && \
...
 
-Run the binary inside a git repository directory with gRPC address of the analyzer
+If your analyzer makes use of UAST, you will need a Babelfish server running.
```
vmarkovtsev: Maybe a link to https://doc.bblf.sh/user/getting-started.html

sdk/README.md
```diff
@@ -28,20 +32,116 @@ $ PY_OUT_DIR=<directory for generated files> mkdir -p $PY_OUT_DIR && \
...
+
+For example, if you place the `lookout` binary in a git repository directory, you can run:
 
 ```bash
 $ ./lookout review ipv4://localhost:10302
 ```
 
-By default it would trigger review event with changes from `HEAD^` to `HEAD`.
-You can change it using additional flags `--from` and `--to`. Both flags accept [git revision](https://git-scm.com/docs/gitrevisions#_specifying_revisions).
+In this case the `lookout` binary will send a gRPC `NotifyReviewEvent` call to your Analyzer, that should be listening on `ipv4://localhost:10302`. The analyzer will be able to request file contents and UAST to the gRPC Data Server endpoint, exposed by the `lookout` binary at `ipv4://localhost:10301`.
```
vmarkovtsev: ~, that~ which

sdk/README.md
bzz: Thank you for pointing this out!

Suggestion: change to

```
python -m grpc_tools.protoc -Isdk \
     --python_out=./sdk/out --grpc_python_out=./sdk/out \
     $(find sdk -name *.proto)
```
and add `find sdk/out -iname gogo_pb2_grpc1.py | grep ".*"` to the code generation CI.

carlosms: :+1: I added `ls "$PY_OUT_DIR"/*.py` instead, which will fail if there isn't any `.py`, and should help us to keep an eye on what files were generated.

vmarkovtsev: Fellows, I have recently found a very nasty problem with this. It conflicts with `bblfsh` Python package: whichever is imported first wins, the second guy explodes :) The reason is because the protobuf descriptors have global names and `bblfsh` uses the same dependencies. I had to hack to fix it: https://github.com/src-d/style-analyzer/blob/master/lookout/core/generate_rpc.sh#L12

But anyway this example will work in isolation from the other packages.

smacker: @carlosms I didn't get how `ls` helps here. Current command generates files, but it has the line:
```
from github.com.gogo.protobuf.gogoproto import gogo_pb2 as github_dot_com_dot_gogo_dot_protobuf_dot_gogoproto_dot_gogo__pb2
```
which would fail in runtime.

we need to add generation for gogo or use similar bblfsh hack as Vadim described.

bzz: Eh, @carlosms indeed the purpose of this discussion was different :) 

It was not about adding a check, but rather about to make sure we generate code for gogo things that was not generated before  (see `find` above).

There must be a better way of handling conflict that @vmarkovtsev found, but that should be discussed/coordinated with LA in a separate issue.

For now, let's include a gogo generation in SDK doc example (and add notice about possible runtime conflict \w bblfsh + a link to workaround from above).

I think it's OK to keep ls for debugging, at the same time - let's add a check specifically to make sure gogo things are generated as well.

carlosms: Got it. Done in 67e44ed.

Issue #35, opened 2018-07-19
Add DB storage service \w kallax


bzz:
This is initial step of adding storage to the `lookout serve`.

Let's store something really simple, like only the input (events) and the output (comments), linked to each other.

PR #95, opened 2018-07-31
Add DB storage

smacker:
Fix: #35 

Code is here. CI pass. But should we update any docs?


server/server.go
```diff
@@ -29,47 +32,81 @@ type ServerConfig struct {
...
+		logger.Errorf(err, "can't save event to database")
+		return
+	}
+
+	if status == models.EventStatusProcessed {
+		logger.Infof("event successfully processed, skipping...")
+		return
+	}
+
+	defer func() {
```
carlosms: If we have only one return statement, do we need a defer?
Wouldn't it be more clear if this code was after the switch?

server/server.go
```diff
@@ -103,14 +138,17 @@ func (s *Server) HandleReview(ctx context.Context, e *ReviewEvent) error {
 	}
 	comments := s.concurrentRequest(ctx, logger, conf, send)
 
-	s.post(ctx, logger, e, comments)
-	s.status(ctx, logger, e, SuccessAnalysisStatus)
+	if err := s.post(ctx, logger, e, comments); err != nil {
+		return fmt.Errorf("posting analysis failed: %s", err)
```
carlosms: I think this will leave the github status in pending, and we should set it to lookout.ErrorAnalysisStatus.

server/server.go
```diff
@@ -143,16 +180,18 @@ func (s *Server) HandlePush(ctx context.Context, e *PushEvent) error {
...
```
carlosms: Same comment as in HandleReview: I think this will leave the github status in pending, and we should set it to lookout.ErrorAnalysisStatus.

server/server.go
smacker: true. Thank you. I believe before I was returning inside switch that's why I needed defer. But not it will be much better to move after switch.

server/server.go
smacker: true. Thanks!

smacker:
Adding migration to the binary is a fairly big task. It's better to implement it separately. Here is an issue for that: https://github.com/src-d/lookout/issues/100

PR #96, opened 2018-07-31
Docker images for lookout and dummy

### LINKED ISSUE #79: Add server Docker image to release artefacts
To be able to deploy lookout server \w dummy analyzer on staging \w k8s, we need to have a Docker images for:
 - `lookout serve`
 - dummy analyzer

It makes sense to add Server one as a part of our usual release process CI automation (so eventually it will include `lookoutd` instead of `lookout serve`).

Not sure if the same is needed for Dummy - it might be enough just to have a 1 command automation to push a new Dummy image version manually from time to time.

carlosms:
Fix #79.

Add 2 dockerfiles for `lookout` and `dummy`. Only the lookout docker image is set to be released on tag by travis.

With the new `Makefile.dummy` we can use the common `docker-build` or `docker-push` targets, as easy as:
```bash
$ make -f Makefile.dummy docker-push
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ before_install:
 
 stages:
   - name: tests
-  - name: release to github
+  - name: release
     if: tag IS present
 
 jobs:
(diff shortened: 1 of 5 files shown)


carlosms:
I forgot to mention that this change requires the DOCKER_USERNAME and DOCKER_PASSWORD env vars in travis, which I believe can only be set by @bzz.

smola:
Added DOCKER_USERNAME and DOCKER_PASSWORD to Travis. Hopefully it didn't change with respect last password I had.

Issue #82, opened 2018-07-27
Correctness guarantees: same comment should never be posted twice


bzz:
Same review request event can be processed multiple times by lookout service for a number of different reasons.

Part of the service correctness guarantees for the users is that he never gets same feedback multiple times in the same review thread.

Most probably this can be done by a Serve checking with HistoryDB, if same comments were already provided to this review thread, before asking a specific Poster to actually go and post/create a new ones in a code review systeem.

(At at current Phase1, we want to keep pulling code review system for events and not be smart about scheduling things based on HistoryDB yet)

Marked as umbrella as this would most probably require some new integration tests in a separate profile on CI.

PR #98, opened 2018-08-01
Don't post the same comments

smacker:
Based on #95 

Fix: #82


pb/event.pb.go
```diff
@@ -117,7 +117,9 @@ type ReviewEvent struct {
 	// Merge test branch where the PullRequest was merged.
 	Merge ReferencePointer `protobuf:"bytes,9,opt,name=merge" json:"merge"`
 	// Configuration contains any configuration related to specific analyzer
-	Configuration  google_protobuf2.Struct `protobuf:"bytes,10,opt,name=configuration" json:"configuration"`
+	Configuration google_protobuf2.Struct `protobuf:"bytes,10,opt,name=configuration" json:"configuration"`
+	// Number is internal provider id of review scoped by repository
+	Number         uint32 `protobuf:"varint,11,opt,name=number,proto3" json:"number,omitempty"`
```
smola: Maybe something more descriptive than just `Number`?

smacker: any suggestions?
On github they call it `number`.
On gitlab `iid`.

smacker:
The comment got collapsed. Post it here:

> Maybe something more descriptive than just Number

any suggestions?
On github they call it `number`.
On gitlab `iid`.

smola:
> any suggestions?

Not really :( `number` is ok then, at least it matches GitHub.

smacker:
rebased on master.

`AnalyzerComments` makes code really ugly :(

carlosms:
> AnalyzerComments makes code really ugly :(

Thanks.

If it's a problem we can try to refactor it. For instance one alternative I can think of is:
- revert the AnalyzerComments stuff
- remove the analyzer feedback url from the main yml config file
- add a new string field to each `pb.Comment` for the feedback link
- filling this field is responsibility of each analyzer

This would also allow an analyzer to provide a different feedback url for different types of comments.

smacker:
Haha. No offense :-D
I have create issue: https://github.com/src-d/lookout/issues/104
Let's put it in the bottom of our backlog and return to it when we have time.

Issue #100, opened 2018-08-02
Include migration in lookout binary


smacker:
We want to have a single binary as a release artifact. It means migrations must be included in it. It also makes sense to expose some of kallax migration commands through lookout.

- use go-bindata to include sql files with migration into the binary
- expose `migrate` command that would do the same as `kallax migrate up` (maybe with `--all` by default) but with migrations from go-bindata.
- update makefile and travis to use new command
- update documentation & docker-compose to run migration on start

smola:
@smacker We probably do not need the run migration on start flag.

At least for our Kubernetes deployment, we can just use a init container running the migration, which runs before the main container starts, this way we have finer-grained control of the lifecycle.

smacker:
Yes. We definitely shouldn't use this flag for real deployment.
We don't want the service being down while migration is running or multiple containers starting migration at the same time. :-D
But it might be handy when you run it locally. Though as I mentioned `lookout migrate && lookout serve` is simple enough. Description updated.

PR #102, opened 2018-08-02
Add go-bindata, use it for store/migrations/ files

carlosms:
Part of #100


carlosms:
I've added the `-nometadata` flag, otherwilse `make no-changes-in-commit` failed because of the file timestamp

Issue #78, opened 2018-07-27
Configurable polling interval for Github


bzz:
Right now github watcher uses  `Basic Authentication` with a single user's [oath token](https://developer.github.com/v3/auth/#via-oauth-tokens) to poll Events API. It works well, but @carlosms well noted that this API, on top of the usual 1 token API limits, is subject of [additional restrictions on frequency AKA "X-Poll-Interval" ](https://developer.github.com/v3/activity/events/).

Current implementation potentially should be enough for Phase1 of development, but again, as @carlosms noted, polling interval can be sometimes i.e 1min and then latency to new PR processing will be unacceptably high, even for small deployments like we are targeting this Q. 

This issue is about changing github provider, so it: 
 - has a configurable max poll delay, N sec, with some low default (i.e every 5 sec)
 - supports multiples Access tokens, that for now are provided though Server configuration file
    (but eventually those will be tokens, based on Github Applications Installations)
 - provides ability to poll actual GH api at >=  given frequency by using next GH client with a different token, if current one has Poll-Interval to high to satisfy configured frequency.

This would allow to get poll interval as close as possible to configured N, given bigger number of tokens, and obeying API usage policy.

Most probably, such low interval would require implementation to be careful by not scheduling events that were already processed, not to overwhelm analyzers i.e by checking event status with DB from #53 first.

bzz:
Expose min poll interval though configuration per-repository.

PR #106, opened 2018-08-02
support multiple tokens in watcher

smacker:
start working on #78 though now we use different api to get pull requests, so the original issue isn't very relevant anymore. Delay in `push` events isn't critical for us.

What is relevant - we need to support multiple transports/tokens for different repositories (transports are going to be github apps installations that issue tokens later)

this PR implements support for multiple tokens in watcher. But don't read them from configuration yet (but I updated yml file with example)

There are no tests for TokenTransport because I assume it will be replaced by apps transport pretty soon.

Current plan for work after this PR:
- calculate pull intervals based on transport (there are limits per user for tokens and per installation for apps tokens)
- use github.Transport in Poster
- read configuration from yml file


smacker:
Actually this abstraction might be wrong. Let me implement new interval and poster before creating PR.

PR #107, opened 2018-08-03
New subcommand, lookout migrate

carlosms:
Part of #100 

Kallax uses [mattes/migrate](https://github.com/mattes/migrate), but apparently it is deprecated in favor of [golang-migrate/migrate](https://github.com/golang-migrate/migrate).

I needed to change the bindata prefix because otherwise the migrate bindata source could not find the files.


Makefile
```diff
@@ -145,7 +144,3 @@ packages-sdk: build
 			tar -cvzf $(PROJECT)_$${TAR_VERSION}_$${os}_$${arch}.tar.gz $(PROJECT)_$${os}_$${arch}/; \
 		done; \
 	done
-
-.PHONY: migrate
-migrate: $(DEPENDENCIES)
-	kallax migrate up -d ./store/migrations --dsn '$(DSN)' --all
```
smacker: Do we still need `DSN` variable in Makefile then?

smacker:
Looks like `golang-migrate/migrate` isn't compatible with `mattes/migrate`.

Maybe we should document somewhere that one can't use `kallax migrate` anymore for applying migrations, only for the generation... Though it's confusing anyway...

carlosms:
I can't seem to find anything in the docs. Can you please share any link with the incompatibilities?

If we need to use the mattes lib is not a bit change. I think it doesn't support bindata directly as a source, but we can write the files to a tmp dir.

smacker:
Wait. Master of mattes/migrate already points on v3. Then no problem. Sorry.

Makefile
carlosms: True, I missed that. Fixed in 39cb2a5

PR #108, opened 2018-08-03
Apply schema migrations to empty DB automatically

carlosms:
Part of #100, done on top of #107 (only last 2 commits are relevant to this PR)

`lookout serve` applies the migration when connecting to an uninitialized DB.
If the DB is initialized, the DB schema version is checked, but no migration is applied. In this case the admin would need to perform `lookout migrate` manually.

This needs to be documented, but I'd prefer to get the OK for the functionality in here and #107 and then do the docs in a new PR.


smacker:
@smola do you think we need it? I have doubts. I think it might create some unexpected behavior during deployment.

carlosms:
Can you share what potential problems do see?

The initialization on empty DB removes friction for users, and I think it would be the expected behaviour for most people. You point lokoout to an empty DB and it makes use of it, there shouldn't be any extra steps required.

And we can still run `lookout migrate` separately if we plan some deployment that requires the DB schema to be ready on advance. Like some future distributed mode.

smacker:
Those problems aren't very likely to happen but if they do it will be very unexpected. The biggest problem it's hard to foresee them. I have one example:

Something happened to DB. It started without previous data.
k8s restart lookout, it recreates schema (instead of failing) and re-process all PRs.

By "something" a lot of different things can happen. In my past experience, we fucked-up configuration a little and one slave was running without persistent storage. One day it got promoted to master.

That's why initially I suggested to add optional flag (disabled by default).
But then comparing `lookout serve --init-db` vs `lookout migrate && lockout serve` I don't see any benefits from adding this flag.

smola:
Having separate init/migrate would be safer at the moment.

carlosms:
I don't really get why it's any safer simply because it's a separate step that does exactly the same. But I'll close this PR and open a new only with the DB version check.

carlosms:
Closing in favor of #117.

PR #109, opened 2018-08-03
support multiple tokens

smacker:
start working on #78 though now we use different api to get pull requests, so the original issue isn't very relevant anymore. Delay in push events isn't critical for us.

What is relevant - we need to support multiple transports/tokens for different repositories (transports are going to be github apps installations that issue tokens later)

Coverage did decrease but not as codecov shows it. It decreased mostly because of `NewClientPoolFromTokens` that suppose to be temporary.

### DIFF
--- a/cmd/lookout/serve.go
+++ b/cmd/lookout/serve.go
@@ -61,6 +61,13 @@ type Config struct {
 	Providers     struct {
 		Github github.ProviderConfig
 	}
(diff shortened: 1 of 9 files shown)


bzz:
So sad that go-github leaves information from headers un-exported and no API to override the back-off logic, so it needs to be re-implement with `limitRoundTripper`

Issue #101, opened 2018-08-02
Recreate initial migration before release for the first deployment


smacker:
It's better to squash all migrations to one before the first release with DB.
They aren't going to work on DB with data anyway.

PR #111, opened 2018-08-07
squash migrations

smacker:
fix: #101


smacker:
Small changes to `make bindata` target:
- added chmod to fix differences in flags (same as we do in dashboard)
- replace modtime by 1 instead of some random number to make it clear it's faked modtime

Issue #103, opened 2018-08-02
Better testing for server


smacker:
Currently, we have small `make test-json` test which definitely isn't enough.

Proposal:

Core tests. They make sure the core of our service works correctly and we don't rely on github too heavily (means everything works with json provider):

1. Rework current `make test-json` using the same (or even improved) approach as we do with `make test-sdk`
2. Use new DB for each run, remove it after the test is finished
3. Add tests for:
  - wrong CommitRevision
  - multiple analyzers
  - tests for push event
  - test for analyzer response with error
  - local `.lookout.yml` (as a bonus because it would require separate repository and it might be too much for this task)

We also need better tests for github provider (both unit&integration) but it will be covered in another issue.

bzz:
> We also need better tests for github provider (both unit&integration) but it will be covered in another issue.

#122

smacker:
all except the last point. It requires special repo. I have created separate issue for that #134

PR #114, opened 2018-08-07
Better integration tests

smacker:
Part of #103


smacker:
Ooops. Tests are failing after rebase. I'm investigating.

smacker:
fixed.

bzz:
Very nice!

As soon as these are better integration tests, and those are not run with `go test -tag=integration` - may be it's a good time to drop a few lines of Developer Docs to the https://github.com/src-d/lookout/blob/master/CONTRIBUTING.md on how to 
 - build project
 - run tests
 - run integration tests

WDYT?

Makefile
```diff
@@ -79,32 +79,12 @@ $(PROTOC):
...
 test-sdk: clean-sdk build-sdk
-	DUMMY_BIN=$(DUMMY_BIN) LOOKOUT_BIN=$(LOOKOUT_BIN) $(GOCMD) run sdk-test/main.go
+	DUMMY_BIN=$(DUMMY_BIN) LOOKOUT_BIN=$(LOOKOUT_BIN) $(GOCMD) run cmd/sdk-test/main.go
 
 # Integration test for lookout serve
 .PHONY: test-json
-test-json: build-sdk
-	$(DUMMY_BIN) serve &>/dev/null & \
-	DUMMY_PID=$$!; \
-	cat fixtures/events.jsonl | $(LOOKOUT_BIN) serve --provider json dummy-repo-url > serve.txt 2> serve-err.txt & \
```
bzz: A bit confused, could you help me understand: if all the logic from here was ported to server-test - is that expected not to have any mentions of `events.jsonl` below?

smacker:
I added commands for build&tests in contributing.

Makefile
smacker: All the logic was ported.
Every single input from `events.jsonl` is run by it now (in the same way through stdin but with go instead of bash).
It gives much more fine-grained control on what to what/when to send and what is more important - output related only to failing test instead of the huge log of everything.
So I don't see any reason to pipe `events.jsonl` into lookout anymore.

Does it answer your question?

bzz: Yes, makes perfect sense!

It's just that I can not find `events.jsonl` file neither mentioned/read, nor remove in this PR below.

bzz:
> It gives much more fine-grained control on what to what/when to send and what is more important - output related only to failing test instead of the huge log of everything.

that's super nice!

CI seems to be failing:

```
DUMMY_BIN=build/bin/dummy LOOKOUT_BIN=build/bin/lookout go run cmd/server-test/main.go
start integration testing
success review...OK!
skip review event...OK!
process review but don't post anything...OK!
wrong commit revision...OK!
success push...'msg="New status" status=success' is not found in output:
stopping services
exit status 1
make: *** [test-json] Error 1
```

but could not immediately tell which case from the logs though

smacker:
Fixed by removing decreasing of the timeout.
In your example, I can clearly see - nothing is happening :-D
Most probably because we don't have logs for sync yet.

Tough relying on fetch from gh in tests doesn't sound like a very good idea. But let's leave it out of this pr.

bzz:
👍 last open question is https://github.com/src-d/lookout/pull/114#discussion_r208942858

PR #117, opened 2018-08-08
Check DB version on lookout serve start

carlosms:
Part of #100, replaces #108.

`lookout serve` checks the DB is initialized and the current migration version.

### DIFF
--- a/cmd/lookout/serve.go
+++ b/cmd/lookout/serve.go
@@ -20,6 +20,7 @@ import (
 	"github.com/src-d/lookout/util/cli"
 	"github.com/src-d/lookout/util/grpchelper"
 
+	"github.com/golang-migrate/migrate"
 	_ "github.com/lib/pq"
 	"google.golang.org/grpc"
 	"gopkg.in/src-d/go-billy.v4/osfs"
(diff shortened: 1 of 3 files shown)


cmd/lookout/serve.go
```diff
@@ -230,6 +232,48 @@ func (c *ServeCommand) startServer(srv *lookout.DataServerHandler) error {
...
+
+	maxVersion, err := store.MaxMigrateVersion()
+	if err != nil {
+		return nil, err
+	}
+
+	if dbVersion != maxVersion {
+		return nil, fmt.Errorf(
+			"database version mismatch. Current version is %v, but this binary needs version %v. "+
+				"Use 'lookout migrate' to upgrade your database", dbVersion, maxVersion)
```
bzz: A nitpick, but would not referring to a specific binary version improve the ability to debug this failure?

E.g  `"Current DB version is %v, but this binary %v build %v needs DB version %v. ", ... version, build, ...`

carlosms: good idea, I'll squash this small change and merge

Issue #116, opened 2018-08-08
Fix race conditions


carlosms:
`make test-race` reports too many race conditions.

smacker:
After fixes in tests and logger the only race I see created by:
`vendor/gopkg.in/src-d/go-git.v4/storage/filesystem/dotgit/writers.go:52`

The line:
```go
go writer.buildIndex()
```
https://github.com/src-d/go-git/blob/master/storage/filesystem/dotgit/writers.go#L52

so the race is in go-git and not in our code. Though after quick inspecting go-git code I'm not sure it's really a race...

smacker:
Update: more races were introduced by #128 🤦‍♂️

PR #115, opened 2018-08-08
Fix github PR ID cast to string

carlosms:
Testing on a real github repository, I was getting the same event-id reported in the logs for 2 different github PRs.

```
[2018-08-08T11:48:01.620369587+02:00]  INFO event successfully processed, skipping... event-id=a29ea85b92358e61e3a56281c5f6e885fa170ab7 event-type=2 source=server/server.go:62
[2018-08-08T11:48:01.620912275+02:00]  INFO event successfully processed, skipping... event-id=a29ea85b92358e61e3a56281c5f6e885fa170ab7 event-type=2 source=server/server.go:62
```

With some logs I saw that the string casting was producing some unicode
```
[2018-08-08T11:48:01.618595027+02:00] DEBUG castPullRequest pr.GetID=206773903 pre.InternalID=� source=github/utils.go:78
```

After the change in this PR the ids match:

```
[2018-08-08T11:51:40.137916423+02:00] DEBUG castPullRequest pr.GetID=206773903 pre.InternalID=206773903 source=github/utils.go:79
```

PR #118, opened 2018-08-08
Fix race conditions in the github provider tests 

carlosms:
Part of #116, done on top of #115.

There were times when the tests failed.
`TestWatch_WithError` in particular failed or passed depending on the event type that got processed first.

### DIFF
--- a/provider/github/watcher_test.go
+++ b/provider/github/watcher_test.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"net/url"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -46,11 +47,12 @@ func (s *WatcherTestSuite) SetupTest() {
 }
 
 func (s *WatcherTestSuite) TestWatch() {
-	var callsA, callsB, events, prEvents, pushEvents int
+	var callsA, callsB, events, prEvents, pushEvents int32
 
-	pullsHandler := func(calls *int) func(w http.ResponseWriter, r *http.Request) {
+	pullsHandler := func(calls *int32) func(w http.ResponseWriter, r *http.Request) {
 		return func(w http.ResponseWriter, r *http.Request) {
-			*calls++
+			atomic.AddInt32(calls, 1)
+
 			etag := "124567"
 			if r.Header.Get("if-none-match") == etag {
 				w.WriteHeader(http.StatusNotModified)
@@ -62,9 +64,10 @@ func (s *WatcherTestSuite) TestWatch() {
 		}
 	}
 
-	eventsHandler := func(calls *int) func(w http.ResponseWriter, r *http.Request) {
+	eventsHandler := func(calls *int32) func(w http.ResponseWriter, r *http.Request) {
 		return func(w http.ResponseWriter, r *http.Request) {
-			*calls++
+			atomic.AddInt32(calls, 1)
+
 			etag := "124567"
 			if r.Header.Get("if-none-match") == etag {
 				w.WriteHeader(http.StatusNotModified)
@@ -108,11 +111,11 @@ func (s *WatcherTestSuite) TestWatch() {
 		return nil
 	})
 
-	s.True(callsA > 2)
-	s.True(callsB > 2)
-	s.Equal(4, events)
-	s.Equal(2, prEvents)
-	s.Equal(2, pushEvents)
+	s.True(atomic.LoadInt32(&callsA) > 2)
+	s.True(atomic.LoadInt32(&callsB) > 2)
+	s.EqualValues(4, events)
+	s.EqualValues(2, prEvents)
+	s.EqualValues(2, pushEvents)
 	s.EqualError(err, "context deadline exceeded")
 }
 
@@ -134,16 +137,12 @@ func (s *WatcherTestSuite) TestWatch_WithError() {
 	s.NoError(err)
 
 	err = w.Watch(context.TODO(), func(e lookout.Event) error {
-		switch e.Type() {
-		case pb.ReviewEventType:
-			s.Equal("96de6d9d321aa94faa0f053c72e2684a44961e5b", e.ID().String())
-		case pb.PushEventType:
-			s.Equal("d1f57cc4e520766576c5f1d9e7655aeea5fbccfa", e.ID().String())
-		}
+		s.Equal("d1f57cc4e520766576c5f1d9e7655aeea5fbccfa", e.ID().String())
+
 		return fmt.Errorf("foo")
 	})
 
-	s.Error(err, "foo")
+	s.EqualError(err, "foo")
 }
 
 func (s *WatcherTestSuite) TearDownSuite() {


provider/github/watcher_test.go
```diff
@@ -121,7 +133,7 @@ func (s *WatcherTestSuite) TestWatch_WithError() {
 		fmt.Fprint(w, `[{"id":1}]`)
 	})
 	s.mux.HandleFunc("/repos/mock/test/events", func(w http.ResponseWriter, r *http.Request) {
-		fmt.Fprint(w, `[{"id":"1", "type":"PushEvent", "payload":{"push_id": 1}}]`)
+		fmt.Fprint(w, `[{"id":1}]`)
```
smacker: why did you change payload? Now the code is just skipping event, no?

provider/github/watcher_test.go
```diff
@@ -133,16 +145,18 @@ func (s *WatcherTestSuite) TestWatch_WithError() {
...
-			s.Equal("d1f57cc4e520766576c5f1d9e7655aeea5fbccfa", e.ID().String())
-		}
+		wg.Add(1)
+
+		s.Equal("d1f57cc4e520766576c5f1d9e7655aeea5fbccfa", e.ID().String())
+
 		return fmt.Errorf("foo")
 	})
 
+	wg.Wait()
```
smacker: I don't see `wg.Done()` anywhere. This line waits forever.

provider/github/watcher_test.go
```diff
@@ -108,11 +117,14 @@ func (s *WatcherTestSuite) TestWatch() {
 		return nil
 	})
 
+	mutex.Lock()
```
smacker: do we really need this lock?
we check `s.EqualError(err, "context deadline exceeded")` it means `Watch` function stopped already and there shouldn't be any changes to this variables.

provider/github/watcher_test.go
```diff
@@ -46,11 +47,16 @@ func (s *WatcherTestSuite) SetupTest() {
...
 func (s *WatcherTestSuite) TestWatch() {
-	var callsA, callsB, events, prEvents, pushEvents int
+	var callsA, callsB, events, prEvents, pushEvents int32
 
-	pullsHandler := func(calls *int) func(w http.ResponseWriter, r *http.Request) {
+	var mutex = &sync.Mutex{}
+
+	pullsHandler := func(calls *int32) func(w http.ResponseWriter, r *http.Request) {
 		return func(w http.ResponseWriter, r *http.Request) {
+			mutex.Lock()
```
smacker: Maybe it would be easier to use `atomic` package here: `atomic.AddUint32(&calls, 1)`
And later when we read it `atomic.LoadUint32(&callsA)`

provider/github/watcher_test.go
carlosms: With or without the payload the resulting event.ID is the same. Since what we are testing here is what happens when the watch callback returns an error, I though I'd simplify this part that is not adding much to the test.

The event is not skipped, that is done by the callback `(s *Server) handleEvent`, replaced here with our callback that just checks the ID.

provider/github/watcher_test.go
carlosms: The funny thing is: `Done` was there, I did a small refactor and it got lost... But it does not wait forever because `Watch` finishes when there is an error unmarshalling the `PushEvent` (changed a string to an int).

Then `s.Error(err, "foo")` is not actually checking the error message for "foo", it should be `s.EqualError(err, "foo")`. So, the test was passing...

provider/github/watcher_test.go
carlosms: Without it `go test -race` complains. I think that even if we end the `Watch` goroutine it might happen just when an http is in process, and the handler func would still be finish running and write callsA/B, even if the response goes nowhere.

provider/github/watcher_test.go
carlosms: ok, pretty similar

carlosms:
Did some fixes for my previous fixes in 0ada83a, hopefully it's correct now

provider/github/watcher_test.go
smacker: Sorry. I don't understand how this test works after this change:

- We call [Watch](https://github.com/src-d/lookout/blob/master/provider/github/watcher.go#L86) with custom callback.
- It calls [watchEvents](https://github.com/src-d/lookout/blob/master/provider/github/watcher.go#L131)
- On event (changed above) it calls [handleEvents](https://github.com/src-d/lookout/blob/master/provider/github/watcher.go#L182) which just [castEvent](https://github.com/src-d/lookout/blob/master/provider/github/utils.go#L25)
- without type it return `nil, nil`
- then we go back to [handleEvents](https://github.com/src-d/lookout/blob/master/provider/github/watcher.go#L188) that does `continue` if event is `nil` and never calls callback.

smacker:
Please check this one: https://github.com/src-d/lookout/pull/118#discussion_r208674567 (it collapsed)

provider/github/watcher_test.go
```diff
@@ -133,17 +137,21 @@ func (s *WatcherTestSuite) TestWatch_WithError() {
...
+	var wg sync.WaitGroup
+
 	err = w.Watch(context.TODO(), func(e lookout.Event) error {
-		switch e.Type() {
-		case pb.ReviewEventType:
-			s.Equal("96de6d9d321aa94faa0f053c72e2684a44961e5b", e.ID().String())
-		case pb.PushEventType:
-			s.Equal("d1f57cc4e520766576c5f1d9e7655aeea5fbccfa", e.ID().String())
-		}
+		wg.Add(1)
```
smacker: this code is super weird. It calls `wg.Add` & `wg.Done` without any concurrency involved. There is only synchronous `s.Equal` check between them. Could you please help me understand how it works.

provider/github/watcher_test.go
smacker: Ah. I see how it works now. Test returns review above. Then maybe it’s better to replace this response by empty array to make it clear that callback is never called for events endpoint?

carlosms: Yes, it's working because the path for the pull event enters the callback and returns the foo error. But it's strange to force a path this way, so I think I'll put back the payload and maybe later we can duplicate all tests and run them  with pull or push events.

provider/github/watcher_test.go
carlosms: Yes, this is a left over of other scaffolding things I was testing...
I put them trying to fix a failing assertion after the test ends:
`panic: Fail in goroutine after TestWatcherTestSuite/TestWatch_WithError has completed`

If the pull event goroutine returns the foo error, the push event goroutine will still call this callback before checking for `ctx.Done`.

Ideally the `Add` would be done in `(w *Watcher) watchEvents` before the `doEventRequest`. But putting `Add` works, I guess because it must be quicker than the `s.Equal`, and forces the `TestWatch_WithError` to wait before exiting.

I'll remove it because it's not technically correct. And this is just there to avoid a panic when an assertion fails, but that's not a concern if the test if failing anyway.

carlosms: changes pushed in ce7184b

PR #119, opened 2018-08-08
Docker compose

smacker:
`docker-compose up` would work only with the fixed image.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,9 +16,8 @@ The included [`./docker-compose.yml`](./docker-compose.yml) allows to start all 
 Clone the repository, or download [`./docker-compose.yml`](./docker-compose.yml), and run:
 
 ```bash
-docker-compose up
+docker-compose up bblfsh postgres
 ```
-
 
 # Example
 
@@ -29,6 +28,19 @@ If you are developing an Analyzer, please check [SDK documentation](./sdk/README
 It includes a curl-style binary that allows to trigger Analyzers directly, without launching a full lookout server.
 
 ## Server
+
+### Docker
+
+To trigger the analysis on an actual pull request of a GitHub repository do:
+
+1. Obtain [GitHub access token](https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/)
+2. Run:
+
+```bash
+GITHUB_USER=<user> GITHUB_TOKEN=<token> REPO=<github repository> docker-compose up
+```
+
+### Manual
 
 To trigger the analysis on an actual pull request of a GitHub repository do:
 
--- a/README.md
+++ b/README.md
@@ -2,48 +2,43 @@
 
 A service for assisted code review, that allows running custom code Analyzers on pull requests.
 
+# SDK
+
+If you are developing an Analyzer, please check [SDK documentation](./sdk/README.md).
+
+It includes a curl-style binary that allows to trigger Analyzers directly, without launching a full lookout server.
+
 # Installation
 
 `go get github.com/src-d/lookout`
 
-## Dependencies
+# Dependencies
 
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -1,6 +1,27 @@
...
+      GITHUB_TOKEN: ${GITHUB_TOKEN}
+    ports:
+      - "10301:10301"
+    entrypoint: ["/bin/sh"]
+    # sleep because container with db is up but the db itself doesn't accept connections yet
+    command: ["-c", "sleep 5 && lookout migrate && lookout serve ${REPO}"]
+    volumes:
+      - ./config.yml:/config.yml
+  dummy:
+    image: "smacker/dummy-analyzer:latest"
```
carlosms: Is this intended or did you forget to change this? If we want to include the dummy analyzer we should probably publish it as srcd. Or we could not include it as part of the compose, and write how setup an external analyzer. Which should be easy with the network mode host, I guess.

smacker: I added it to have simple `docker-compose up` working. Without an analyzer the server wouldn't start.

We don't have any real analyzer for now and I don't see any value in publishing dummy as a project of src-d. I would just replace it by real analyzer from ML as soon as it's ready.

carlosms:
The readme now has a strange layout. If you follow it step by step you would execute docker-compose twice, for the dependencies and for the main execution.

Maybe we could reorder like this:

```
# lookout
# SDK
# Installation
# Usage
- obtain github token
## With Docker
- docker-compose up
## Without Docker
- docker-compose up bblfsh postgres
# Configuration file
```

docker-compose.yml
carlosms: It's not a big deal and I'm not going to block the PR, but I do see the value in doing all the required releases the same way, with the same credentials.

README.md
```diff
@@ -2,36 +2,43 @@
...
-## SDK
+Run:
 
-If you are developing an Analyzer, please check [SDK documentation](./sdk/README.md).
-
-It includes a curl-style binary that allows to trigger Analyzers directly, without launching a full lookout server.
-
-## Server
+```bash
+GITHUB_USER=<user> GITHUB_TOKEN=<token> REPO=<github repository> docker-compose up
```
bzz: A nitpick, but `github repository` here may be a bit confusing: is it a path, or a name of the repo, or a `user/repo`?  Do you think it might be worth to be specific, to avoid the confusion?

docker-compose.yml
bzz: I agree with Carlos and do not see a strong reason not to publish a dummy at the src-d account as part of the normal release process, esp if we are going to use it as part of the staging env.

smacker: it requires changes to the release process :) we can do it in separate PR if we want.
Though I would suggest another solution to this problem. We can include dummy into lookout itself as a subcommand.

smacker:
github repo path is updated. About dummy I would prefer to leave it out of this PR.

docker-compose.yml
bzz: > it requires changes to the release process
If that is anything more then adding ./docker/Dockerfile.dummy to `DOCKERFILES` - handling it in subsequent PR makes perfect sense. 

About the single image suggestion - at the same time I see value of going though the similar configuration as actual analyzers would go though on deployment.

smacker: it is. Because our docker files aren't self-containing multistep. It requires the building of the binary on host machine. Which would lead to different release artifacts. We may want it or not. Let's track what we are doing and how it the separate issue.

smacker:
https://github.com/src-d/lookout/issues/123

Issue #113, opened 2018-08-07
Network failures end the serve process


carlosms:
For lookout serve, with the github provider, if the network fails the process ends with the message 'context deadline exceeded'.

We probably don't want that, and instead the watcher should be able to keep polling if the network problem gets fixed.

bzz:
`context deadline exceeding`

PR #121, opened 2018-08-09
Don't stop serve process on watcher error

smacker:
Fix: #113


server/server.go
```diff
@@ -38,10 +38,25 @@ func NewServer(w lookout.Watcher, p lookout.Poster, fileGetter lookout.FileGette
...
+		errCh <- err
+	}()
+
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case err := <-errCh:
+		if err != nil {
+			log.Errorf(err, "error during watching, restart watching")
+			return s.Run(ctx)
```
bzz: Could you help me understand - would this be an equivalent of the infinite retry policy, in case of networking failure, though a recursive `s.Run()` calls?

smacker: It's equivalent of the infinite retry policy in case of any failure in the watcher. (except manual stopping by the special type of error, most common error here is a network of course but might be something else)

It means server wouldn't die if something went wrong there (dying at least for now isn't a very good option. We can post comments at the same time for example)

From the administrative point of view, it means we should monitory error logs, not only server restarts.

bzz: makes sense!

What confuses me is not infinite retry policy per se, but potentially infinite recursion, as heap usually is not infinite.

smacker:
changed recursion -> for loop

Issue #125, opened 2018-08-10
Put logger to the context


smacker:
In our code, we often want to extend log messages by the information from caller (and caller of caller and caller of caller of caller).

It's a common solution for this problem in go to keep logger in context that is passed through anyway.

But we do that by passing logger as one more argument. Example:
```go
status(ctx context.Context, logger log.Logger, e lookout.Event, st lookout.AnalysisStatus)
```

I suggest to refactor it.

P.S. Most probably it will solve some race-conditions from #116 too.

PR #127, opened 2018-08-10
Log context

smacker:
Fix: #125 

- Cleaner function signatures
- More informative logs especially for something like `empty pull request branch given`
- Bonus: no more complaining about races in logs from `go test -race`

There are very few places left where we still use global logger. Those are git/bblfsh services because they don't use context at all. (maybe we will want to change it later or maybe not)

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/src-d/lookout/mock"
 	"github.com/src-d/lookout/pb"
 	"github.com/src-d/lookout/store"
+	"github.com/src-d/lookout/util/ctxlog"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/grpc"
 	log "gopkg.in/src-d/go-log.v1"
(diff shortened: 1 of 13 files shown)


server/server_test.go
```diff
@@ -347,6 +347,7 @@ func TestWatcherError(t *testing.T) {
 	srv := NewServer(watcher, poster, fileGetter, analyzers, &store.NoopEventOperator{}, &store.NoopCommentOperator{})
 
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	ctx = ctxlog.Set(ctx, logMock)
```
carlosms: It looks like `Set` is only used in this test. Why is it needed here? Can't you leave the previous `log.DefaultLogger = logMock`, and let `ctxlog.Get` return that for this `ctx`?

smacker: We need it because this test actually checking log messages.

We are interested of the log messages of this test only and don't what to affect other tests.

Now when the logger is scoped it's better to set mock to particular test scope than playing with globals.

Another problem with using globals here: this test would rely on particular behaviour of `ctxlog.Get` (returning `log.DefaultLogger`) but it shouldn't.

smacker:
@bzz imo the rebase will be easy either way but we can wait...

PR #128, opened 2018-08-10
Add github api http errors tests

### LINKED ISSUE #122: More tests for Github provider
More rigorous test for things that might go wrong:
 - Poster/Watcher network failure, partial JSON download, etc
 - Exceeding API limits

carlosms:
Fix #122.

There are many changes, but you can check the individual commits to make it easier to review.

Besides adding tests, there are the following changes in the code:
- `Watcher.Watch` does not exit for expected errors.
We already have a restart in `Server.Run`, but it also makes sense to deal with expected errors in the watcher. This is important because we many be polling several repositories in `Watcher.Watch`, and a network or auth error in only one request will stop all the other requests for other repos that may be working fine.
- `Watcher.watchEvents` uses `time.After(interval)` instead of `time.After(pollInterval)`. This was a bug, discussed via chat.
- `Poster.Status` uses the `ErrGitHubAPI` as return values, to make it similar to `Poster.Post`.

### DIFF
--- a/provider/github/watcher.go
+++ b/provider/github/watcher.go
@@ -55,8 +55,8 @@ func (w *Watcher) Watch(ctx context.Context, cb lookout.EventHandler) error {
 	errCh := make(chan error)
 
 	for client, repos := range w.pool.Clients() {
-		go w.watchPrs(ctx, client, repos, cb, errCh)
-		go w.watchEvents(ctx, client, repos, cb, errCh)
+		go w.watchLoop(ctx, client, repos, w.processRepoPRs, cb, errCh)
+		go w.watchLoop(ctx, client, repos, w.processRepoEvents, cb, errCh)
 	}
 
 	select {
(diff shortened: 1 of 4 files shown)


provider/github/watcher_test.go
```diff
@@ -144,6 +144,95 @@ func (s *WatcherTestSuite) TestWatch_CallbackError() {
...
+	s.True(atomic.LoadInt32(&callsErr) > 1)
+
+	s.EqualError(err, "context deadline exceeded")
+}
+
+func (s *WatcherTestSuite) TestWatch_HttpTimeout() {
+	var calls, callsErr int32
+
+	// Change the request timeout for this test
+	RequestTimeout = 5 * minInterval
```
smacker: I don't see where you return it original value. It means all next tests also use this timeout.
It's better to reset it back or define on the top of the file so it would be visible that all tests run with different timeout.

provider/github/watcher.go
```diff
@@ -73,7 +73,18 @@ func (w *Watcher) Watch(ctx context.Context, cb lookout.EventHandler) error {
 func (w *Watcher) watchPrs(ctx context.Context, c *Client, repos []*lookout.RepositoryInfo, cb lookout.EventHandler, errCh chan error) {
 	for {
 		for _, repo := range repos {
+			select {
+			case <-ctx.Done():
+				return
+			case <-time.After(w.newInterval(c.Rate(coreCategory))):
```
smacker: could you please double check it works correctly?
the problem is before we were calling `c.Rate` after a request is done, now before. So it returns empty value. As I remember I didn't really test it with empty value.

provider/github/watcher.go
```diff
@@ -83,21 +94,27 @@ func (w *Watcher) watchPrs(ctx context.Context, c *Client, repos []*lookout.Repo
...
 
 func (w *Watcher) watchEvents(ctx context.Context, c *Client, repos []*lookout.RepositoryInfo, cb lookout.EventHandler, errCh chan error) {
+	interval := minInterval
+
 	for {
 		for _, repo := range repos {
+			select {
+			case <-ctx.Done():
+				return
+			case <-time.After(interval):
```
smacker: why do we need to wait for the interval before doing anything?

carlosms: To ensure we wait in between iterations after the new `continue` if the error is `ErrGitHubAPI`.

The other options I thought of are `if/else`, which goes a bit against the early return style and avoiding nested code:

```go
func (w *Watcher) watchEvents(ctx context.Context, c *Client, repos []*lookout.RepositoryInfo, cb lookout.EventHandler, errCh chan error) {
	for {
		for _, repo := range repos {
			resp, events, err := w.doEventRequest(ctx, repo.Username, repo.Name)

			var pollInterval time.Duration

			if err != nil {
				if ErrGitHubAPI.Is(err) {
					log.With(log.Fields{"repository": repo.FullName, "response": resp}).Errorf(err, "request for events list failed")
				} else if !NoErrNotModified.Is(err) {
					errCh <- err
					return
				}
			} else {
				if err := w.handleEvents(cb, repo, resp, events); err != nil {
					errCh <- err
					return
				}

				interval := w.newInterval(c.Rate(coreCategory))
				pollInterval = c.PollInterval(eventsCategory)
				if pollInterval > interval {
					interval = pollInterval
				}
			}

			select {
			case <-ctx.Done():
				return
			case <-time.After(pollInterval):
				continue
			}
		}
	}
}
```
Or wrapping the block in a function that can return at any moment, and still end in the main last `select`:

```go
func (w *Watcher) watchEventsC(ctx context.Context, c *Client, repos []*lookout.RepositoryInfo, cb lookout.EventHandler, errCh chan error) {
	for {
		for _, repo := range repos {
			var pollInterval time.Duration

			err := func() error {
				resp, events, err := w.doEventRequest(ctx, repo.Username, repo.Name)
				if ErrGitHubAPI.Is(err) {
					log.With(log.Fields{"repository": repo.FullName, "response": resp}).Errorf(err, "request for events list failed")
					return nil
				}

				if err != nil && !NoErrNotModified.Is(err) {
					return err
				}

				if err := w.handleEvents(cb, repo, resp, events); err != nil {
					return err
				}

				interval := w.newInterval(c.Rate(coreCategory))
				pollInterval = c.PollInterval(eventsCategory)
				if pollInterval > interval {
					interval = pollInterval
				}

				return nil
			}()

			if err != nil {
				errCh <- err
				return
			}

			select {
			case <-ctx.Done():
				return
			case <-time.After(pollInterval):
				continue
			}
		}
	}
}
```

Do you think any of these 2 are better options?

provider/github/watcher.go
carlosms: Yes, with zero values it returns the `minInterval`

provider/github/watcher_test.go
carlosms: ooops! fixed in 88235aa, thanks

provider/github/watcher.go
smacker: Yes. I think the second option is better. Because waiting for the interval on start instead of start watching immediately looks strange.

If you use the second option you can also move for-for loop, select and error handling to separate function and reuse it in watchPRs too.

carlosms: Pushed the changes in d6c77ad, please take a look.

Issue #126, opened 2018-08-10
Include the language name for each File


vmarkovtsev:
The current definition in `service_data.proto`:

```proto
message File {
    option (gogoproto.sizer) = false;
    option (gogoproto.protosizer) = true;

    // File path.
    string path = 1;
    // POSIX-style file mode.
    uint32 mode = 2;
    // Hash of the file contents.
    string hash = 3;
    // Raw content of the file.
    bytes content = 4;
    // UAST.
    gopkg.in.bblfsh.sdk.v1.uast.Node uast = 5 [(gogoproto.customname) = "UAST"];
}
```

We have realized that it's not enough: we need to know the language of the file in order to process the UAST correctly.

bzz:
Most probably simplest way would be to add enry support to bblfsh service, in order to get the language and also set it for the bblfhs call.

PR #131, opened 2018-08-13
Add language to file

smacker:
Fix: #126 

I believe language information can be useful without UAST so I implemented it using enry.

Changes:
- add `want_language` to requests
- add `language` string to response
- `want_uast` sets `want_language` to true and don't prune it
- adding language information implemented by new `enry` service
- prune function extracted from bblfsh service to a separate one
- it calls bblfsh service with a language now (we need a new issue to ignore calls for unknown languages using `supportedLanguages` method of bblfsh)


sdk/service_data.proto
```diff
@@ -70,4 +73,5 @@ message FilesRequest {
     bool exclude_vendored = 4;
     bool want_contents = 5;
     bool want_uast = 6 [(gogoproto.customname) = "WantUAST"];
+    bool want_language = 7;
```
bzz: While at it, shall we add a `/n` to the end of the file as well?

service/bblfsh/bblfsh.go
```diff
@@ -121,6 +111,7 @@ func (s *BaseScanner) parseFile(f *lookout.File) (*uast.Node, error) {
 		Filename: f.Path,
 		Content:  string(f.Content),
 		Encoding: protocol.UTF8,
+		Language: strings.ToLower(f.Language),
```
carlosms: why do we want to convert to lowercase here?

smacker: because otherwise bblfsh returns error like `driver for language Go isn't installed`.

cmd/lookout/event.go
```diff
@@ -104,6 +107,8 @@ func (c *EventCommand) makeDataServerHandler() (*lookout.DataServerHandler, erro
 		dataService = bblfsh.NewService(dataService, dataService, bblfshConn)
 	}
 
+	dataService = purge.NewService(dataService, dataService)
```
bzz: This code takes some time to understand why the same variable name is used 3 times in a row. 

One possible way to change it could be by introducing a convenience method with a single argument:
```
purgedDataService := purge.NewServiceUsingSame(dataService)
```

In general, for all the services as far as I can see - `NewService` method with 2 arguments is used only in tests, the rest or the code passes same instances in both.

Although I can see benefits of having those 2 separated, on the API level it could be worth changing to:
 - exported version of `NewService(s)` receives 1 argument and delegates to
 - `newServicesUsingDifferent(s, s)`
 - which can also be used in all the tests

What do you think?

smacker: I don't think refactoring of constructors is in a scope of this PR.

PR #135, opened 2018-08-14
SDK Integration tests on macOS

### LINKED ISSUE #105: Run tests on macos
1. Don't run tests that require bblfsh if `-short` flag is provided
2. Add tests on macos to CI with this flag on master branch.

carlosms:
Fix #105.

I refactored the existing `main.go` to be a go test. This way we can use the `short` flag, and other nice things like `-timeout`, `-run` individual tests, etc.

If you guys agree I'd like to refactor `server-test/main.go` to be a test also, and use testing assertions in `util/cmdtest`.

For travis the condition is not `if: branch = master` because for PRs that is the base branch name, so it would run for every PR against master.


Makefile
```diff
@@ -76,15 +76,26 @@ $(PROTOC):
...
 # Integration test for sdk client
 .PHONY: test-sdk
 test-sdk: clean-sdk build-sdk
-	DUMMY_BIN=$(DUMMY_BIN) LOOKOUT_BIN=$(LOOKOUT_BIN) $(GOCMD) run cmd/sdk-test/main.go
+	DUMMY_BIN=$(PWD)/$(DUMMY_BIN) \
+	LOOKOUT_BIN=$(PWD)/$(LOOKOUT_BIN) \
+	$(GOTEST_INTEGRATION) github.com/src-d/lookout/cmd/sdk-test
+
+# Same as test-sdk, but skipping tests that require a bblfshd server
+.PHONY: test-sdk-macos
```
bzz: May be `test-sdk-short` would make it a bit more clear? As one may want to run short version on any os

.travis.yml
```diff
@@ -34,6 +34,13 @@ jobs:
         - docker run -d --name bblfshd --privileged -v $HOME/bblfshd:/var/lib/bblfshd -p "9432:9432" bblfsh/bblfshd
         - docker exec -it bblfshd bblfshctl driver install go bblfsh/go-driver:v2.0.3
         - make test-sdk
+    - name: "SDK Integration Tests macOS"
+      if: NOT type = pull_request
```
bzz: Could you please also help to understand, why we need this _only_ for macOS?

carlosms: The issue description says to run these new tests only on master.
I remember we discussed doing this to avoid making the tests for PRs take longer, while still running everything that we merge to master on macOS.

Makefile
carlosms: :+1: done in 5b4e0b4.

.travis.yml
smacker: yep. Sometimes travis starts macos container with a big delay.
It's enough to run them only on master. In case something went wrong we will receive an email.

smacker:
I have doubts about using test package for integration tests.

`-integration` tag doesn't really work in real life.

```
$ go test -integration ./...
flag provided but not defined: -integration
```


We can't use almost any flags here, like `-parallel`, `-timeout`.
The only one that will work `-run regexp`. Most of other fill just fail tests which is kind of unexpected when you run `go test`.

Overall I think running them using `go test` is confusing.

carlosms:
All go test flags should work properly. Give it a try like this:

```bash
make clean-sdk
make build-sdk

export DUMMY_BIN=$PWD/build/bin/dummy
export LOOKOUT_BIN=$PWD/build/bin/lookout

go test -tags=integration ./...

go test -tags=integration -test.short github.com/src-d/lookout/cmd/sdk-test
go test -v -tags=integration -test.short github.com/src-d/lookout/cmd/sdk-test
go test -v -tags=integration -test.short -run ^TestIntegrationSuite$ -testify.m ^TestPush$ github.com/src-d/lookout/cmd/sdk-test
go test -v -parallel 2 -tags=integration -test.short github.com/src-d/lookout/cmd/sdk-test
go test -v -timeout 1s -tags=integration -test.short github.com/src-d/lookout/cmd/sdk-test
```

DUMMY_BIN and LOOKOUT_BIN need to be set because the defaults in `/util/cmdtest/cmds.go` don't work for `go test`, these can be changed if we refactor `server-test/main.go`.

smacker:
sorry I don't understand how for example `parallel` would work. Integration tests run binaries that bind on port. If you run 2 tests (maybe not in sdk but in server for sure) at the same time they would fail.

`-timeout=2m` wouldn't do much because we have timeout in context https://github.com/src-d/lookout/blob/master/util/cmdtest/cmds.go#L39 btw default timeout in go test is 10m. When we will have more tests we can easily exceed it if we continue cloning repositories. 

For `-tags=integration` yeah, I misunderstood. So it would enable/disable not a particular test but a file. Ok.

Though I like that we can use `-run` flag. I also like using `suite`. So my feelings are kind of mixed.

carlosms:
> sorry I don't understand how for example `parallel` would work. Integration tests run binaries that bind on port. If you run 2 tests (maybe not in sdk but in server for sure) at the same time they would fail.
> 
I'm not proposing to use `-parallel` or any other flag, and I'm not claiming that any of them will make the tests pass. That was just a counter example to show that the `go test` command itself is not broken, and works as expected.
The same way `-timeout 1s` will make the integration tests fail; or `go test -tags=integration` will also fail if the binaries are not built or the dependencies are not up and running.

> `-timeout=2m` wouldn't do much because we have timeout in context https://github.com/src-d/lookout/blob/master/util/cmdtest/cmds.go#L39

Then we can choose to not use the `-timeout` flag. Or use it as a complimentary global timeout. Or remove this context timeout and use only the `go test` timeout. I don't see any problem there.

> btw default timeout in go test is 10m. When we will have more tests we can easily exceed it if we continue cloning repositories. 

Then it's a good thing that the default timeout can be set with the standard `-timeout` flag.



I don't think those are real concerns because the integration tests will only run if `-tags=integration` is used, which means that any normal `go test` workflow or automation will not be affected. 
The main takeaway I want to stress is that there is a testing framework, and we don't need to reinvent a new one.

smacker:
I'm okay with refactoring as long as we hide implementation details by targets in Makefile.

.travis.yml
bzz: 👍 thank you!

bzz: Shall we also document `make test-sdk-short` in https://github.com/src-d/lookout/blob/master/CONTRIBUTING.md#testing ?

PR #141, opened 2018-08-14
Configure min interval

smacker:
Fix: #78 

- read the value from config file
- renamed config variable

### DIFF
--- a/provider/github/client.go
+++ b/provider/github/client.go
@@ -34,12 +34,13 @@ func (p *ClientPool) Client(username, repo string) (*Client, bool) {
 // Client is a wrapper for github.Client that supports cache and provides rate limit information
 type Client struct {
 	*github.Client
-	cache   *cache.ValidableCache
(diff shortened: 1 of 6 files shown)


provider/github/client.go
```diff
@@ -49,10 +50,21 @@ func NewClient(t http.RoundTripper, cache *cache.ValidableCache, l log.Logger) *
 	cachedT.MarkCachedResponses = true
 	cachedT.Transport = limitRT
 
+	interval := minInterval
+	if watchMinInterval != "" {
+		d, err := time.ParseDuration(watchMinInterval)
+		if err != nil {
+			l.Errorf(err, "can't parse min interval")
```
bzz: Things like `"can't parse min interval:'%v'"` can be very useful for debugging

smacker: we already include the value: https://play.golang.org/p/EI1ZBdJ801i

PR #142, opened 2018-08-14
fix new races in tests

smacker:
Fix: #116

As I mentioned in the issue `go test -race` reports one race from `go-git`. But after reading the code it looks like false positive. We can live with it.


bzz:
For some reason `The command "make test-json" exited with 2.` fails on CI.

smacker:
green now.

Issue #149, opened 2018-08-16
DataService: changes in a tree that is a sub-module


bzz:
For now, to avoid failures as in #144 in DataService we just skip all requested files that are not part of the repository. One case when this is not an error is when the changes were done in some sub-module.

Rationale is: most probably that would not be a place user will want to see the feedback on his PR anyway.

Supporting getting files/changes from sub-modules with DataService _may_ make sense in some cases, so this issue is about:
 1. finding such cases
 2. if found, implementing sub-modules support
 3. if not found, may be we should not event detect these as a change and just skip inside `DiffTreeScanner`

bzz:
Using src-d/style-analyzer@c99dcdf...src-d:3a9d78 example from https://github.com/src-d/lookout/issues/144#issuecomment-413528797 it's not immediately clear why changes related to submodules are detected by `DiffTreeScanner`.

After they are detected in a Trees, `File/ChangeBlogScanners` fail to get blobs for such paths.

On manual inspection of those hashes, there are no trees related to a path `lookout/core/server`, that is a sub-module.

```
From Tree

$ git cat-file -p 5e636bfd1263b4669e2973282294dcc1606b78d9
100644 blob 4fd343b5cd43471af35aaf62466e0c6070630f8b	__init__.py
100644 blob 35561554e4a074d279c5d1eb0cd68c10821fb580	__main__.py
040000 tree 2e1d2fee0927fc5c4a8f5562efb5d4033d3b6f59	core
040000 tree 11f294824d517bc746f0fb00a1a4e10943b8fda4	style

To Tree

$ git cat-file -p 581c5c0e21e619095a87a7e953af3474d2c7ee36
100644 blob 4fd343b5cd43471af35aaf62466e0c6070630f8b	__init__.py
100644 blob 35561554e4a074d279c5d1eb0cd68c10821fb580	__main__.py
040000 tree 3007fdce337099fb6d2a3ed1cc00765aaea89c6d	core
040000 tree 7476473b904019468d313869ddd31c11185934cc	style
```

bzz:
1. and 2. is prelimenary done - we do not want to get submodules changes as changes in lookout.

This is only about 3.

bzz:
After a bit more investigation - indeed, a submodule seems to have a Tree entry, and it's `core/server`

```
git diff-tree --raw -r --find-renames 5e636bfd1263b4669e2973282294dcc1606b78d9 581c5c0e21e619095a87a7e953af3474d2c7ee36
[...]

Issue #144, opened 2018-08-16
Review event chokes on git submodules


vmarkovtsev:
Looks like something goes wrong if the commit includes a git submodule update. Steps to reproduce:

1. Checkout the latest master of src-d/style-analyzer
2. `./lookout review -v ipv4://localhost:2000 --from c99dcdff172f1cb5505603a45d054998cb4dd606 --to 3a9d78bdd1139c929903885ecb8f811931b8aa70` (this includes the submodule update)

You get

```
[2018-08-16T11:32:01.355326446+02:00]  INFO streaming server call finished app=lookout duration=670.461043ms error=file not found grpc.code=Unknown grpc.method=GetChanges grpc.service=pb.Data grpc.start_time=2018-08-16T11:32:00+02:00 source=grpchelper/logger.go:19 span.kind=server system=grpc
[2018-08-16T11:32:01.364247864+02:00]  INFO streaming client call finished app=lookout duration=734.359432ms error=rpc error: code = Internal desc = <class 'grpc._channel._Rendezvous'>: <_Rendezvous of RPC that terminated with:
	status = StatusCode.UNKNOWN
	details = "file not found"
	debug_error_string = "{"created":"@1534411921.355470703","description":"Error received from peer","file":"src/core/lib/surface/call.cc","file_line":1083,"grpc_message":"file not found","grpc_status":2}"
> grpc.code=Internal grpc.method=NotifyReviewEvent grpc.service=pb.Analyzer grpc.start_time=2018-08-16T11:32:00+02:00 source=grpchelper/logger.go:19 span.kind=client system=grpc
[2018-08-16T11:32:01.36433733+02:00] ERROR analysis failed analyzer=test-analyzes app=lookout error=rpc error: code = Internal desc = <class 'grpc._channel._Rendezvous'>: <_Rendezvous of RPC that terminated with:
	status = StatusCode.UNKNOWN
	details = "file not found"
	debug_error_string = "{"created":"@1534411921.355470703","description":"Error received from peer","file":"src/core/lib/surface/call.cc","file_line":1083,"grpc_message":"file not found","grpc_status":2}"
> head=3a9d78bdd1139c929903885ecb8f811931b8aa70 provider= repository=file:///home/vadim/Projects/style-analyzer source=server/server.go:256
[2018-08-16T11:32:01.364385434+02:00]  INFO status: success app=lookout source=lookout/serve.go:351
```

bzz:
Thank you for reproducible example! Looking into it

[...]

PR #150, opened 2018-08-16
Fix: skip missing files

bzz:
This is workaround for #144 :
 - log & skip files that do not have Blobs
 - default password for easier reproduction using `docker-compose`
 - a bit more verbose error messages

Better analysis and a long-term solution expected be proposed under #149

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -41,7 +41,7 @@ services:
     ports:
       - "5432:5432"
     environment:
-      POSTGRES_PASSWORD: example
+      POSTGRES_PASSWORD: postgres
       POSTGRES_DB: lookout
 volumes:
   drivers:
--- a/service/git/scanner.go
+++ b/service/git/scanner.go
@@ -250,20 +250,20 @@ func (b *blobAdder) Fn(f *lookout.File) (bool, error) {
 
 	of, err := b.tree.File(f.Path)
 	if err != nil {
-		log.Warningf("skipping - can not get file:'%v', %v", f.Path, err)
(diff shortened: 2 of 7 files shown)


bzz:
macOS profile failed with

```
success push...OK!

stopping services

LOG:  could not receive data from client: Connection reset by peer

multiple analyzers...no comments from the first analyzer

stopping services

LOG:  could not receive data from client: Connection reset by peer

exit status 1

make: *** [test-json] Error 1
```
:/

service/git/scanner.go
```diff
@@ -248,19 +250,20 @@ func (b *blobAdder) Fn(f *lookout.File) (bool, error) {
 
 	of, err := b.tree.File(f.Path)
 	if err != nil {
-		return true, err
+		log.Warningf("skipping - can not get file:'%v', %v", f.Path, err)
```
carlosms: can not -> cannot (here and 2 more instances below)

docker-compose.yml
```diff
@@ -41,7 +41,7 @@ services:
     ports:
       - "5432:5432"
     environment:
-      POSTGRES_PASSWORD: example
+      POSTGRES_PASSWORD: postgres
```
carlosms: How does the DB password change help?

bzz: It simplifies reproducing: running lookout with docker docs with style-analyzer quickstart docs does not require any changes.

service/git/scanner.go
bzz: addressed in e3daced

Issue #147, opened 2018-08-16
Forever loop caused by posting failure


smacker:
```
{"app":"lookout","event-id":[58,195,47,66,46,74,194,47,55,130,233,133,148,130,251,78,44,169,207,10],"event-type":2,"head":"refs/pull/69/head","level":"info","msg":"processing pull request","provider":"github","repository":"git://github.com/src-d/lookout.git","time":"2018-08-16T16:02:47.567487442Z"}
{"app":"lookout","event-id":[58,195,47,66,46,74,194,47,55,130,233,133,148,130,251,78,44,169,207,10],"event-type":2,"head":"refs/pull/69/head","level":"info","msg":"fetching references for repository git://github.com/src-d/lookout.git: [refs/pull/69/head:refs/pull/69/head]","provider":"github","repository":"git://github.com/src-d/lookout.git","time":"2018-08-16T16:02:47.567946249Z"}
{"app":"lookout","event-id":[58,195,47,66,46,74,194,47,55,130,233,133,148,130,251,78,44,169,207,10],"event-type":2,"head":"refs/pull/69/head","level":"info","msg":"repository config is not found","provider":"github","repository":"git://github.com/src-d/lookout.git","time":"2018-08-16T16:02:47.834239076Z"}
{"app":"lookout","grpc.method":"NotifyReviewEvent","grpc.service":"pb.Analyzer","level":"info","msg":"unary client call started","span.kind":"client","system":"grpc","time":"2018-08-16T16:02:48.134196717Z"}
{"app":"lookout","grpc.method":"GetChanges","grpc.service":"pb.Data","level":"info","msg":"streaming server call started","span.kind":"server","system":"grpc","time":"2018-08-16T16:02:48.135535472Z"}
{"app":"lookout","level":"info","msg":"fetching references for repository git://github.com/src-d/lookout.git: [refs/heads/master:refs/heads/master refs/pull/69/head:refs/pull/69/head]","time":"2018-08-16T16:02:48.136013635Z"}
{"app":"lookout","duration":237888884,"grpc.code":0,"grpc.method":"GetChanges","grpc.service":"pb.Data","grpc.start_time":"2018-08-16T16:02:48Z","level":"info","msg":"streaming server call finished","span.kind":"server","system":"grpc","time":"2018-08-16T16:02:48.373414549Z"}
[...]

PR #151, opened 2018-08-17
Skip out of range comments

smacker:
Part of #147

### DIFF
--- a/provider/github/poster.go
+++ b/provider/github/poster.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/google/go-github/github"
 	"gopkg.in/src-d/go-errors.v1"
+	log "gopkg.in/src-d/go-log.v1"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


provider/github/poster.go
```diff
@@ -171,6 +171,10 @@ func (p *Poster) createReviewRequest(
 				req.Comments = append(req.Comments, comment)
 			} else {
 				line, err := dl.ConvertLine(c.File, int(c.Line))
+				if ErrLineOutOfDiff.Is(err) {
+					continue
```
carlosms: Should we log this invalid comment?

bzz: 🙏 on debug level

carlosms: Or maybe error level, if the analyzer is providing malformed comments it will be useful in production too, not only while developing.

smacker: I didn't put log here because it's kind of expected behavior. At least for dummy analyzer. It doesn't calculate diff but scans all file. I asked ML team what they do (all file or diff).
Let's log debug/error depends on their response.

smacker: response from ML team:
> As far as I understand changed files are processed fully (analyzer gets full uasts of previous and new version)

I'm adding it as debug level. Because it's not an error (at least for now)

PR #161, opened 2018-08-22
Fix default auth

### LINKED ISSUE #157: Github env vars are not used
If individual `user` & `token` are not set in the `config.yml` file the log says `use default token for repository`.

But what is actually happening is that the requests are made without any authorization, `GITHUB_USER` & `_TOKEN` env vars are not used.

This is the problem blocking https://github.com/src-d/lookout/issues/80 & https://github.com/src-d/issues-infrastructure/issues/213

carlosms:
Fix #157.

The bug was checking zero value with 2 pointers, instead of values.
I also moved the default auth management to `serve.go`, I think it simplifies it and make it easier to understanding.

### DIFF
--- a/cmd/lookout/serve.go
+++ b/cmd/lookout/serve.go
@@ -141,6 +141,12 @@ func (c *ServeCommand) Execute(args []string) error {
 }
 
 func (c *ServeCommand) initProvider(conf Config) error {
+	noDeafultAuth := c.GithubUser == "" || c.GithubToken == ""
+	defaultConfig := github.ClientConfig{
+		User:  c.GithubUser,
+		Token: c.GithubToken,
+	}
+
 	switch c.Provider {
 	case github.Provider:
 		urls := strings.Split(c.Positional.Repository, ",")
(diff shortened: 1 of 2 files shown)


cmd/lookout/serve.go
```diff
@@ -141,6 +141,12 @@ func (c *ServeCommand) Execute(args []string) error {
 }
 
 func (c *ServeCommand) initProvider(conf Config) error {
+	noDeafultAuth := c.GithubUser == "" || c.GithubToken == ""
```
smacker: noDeafultAuth -> noDefaultAuth

cmd/lookout/serve.go
```diff
@@ -155,16 +161,21 @@ func (c *ServeCommand) initProvider(conf Config) error {
 		for _, url := range urls {
 			conf, ok := repoToConfig[url]
 			if !ok {
-				log.Infof("use default token for repository %s", url)
+				if noDeafultAuth {
+					// Empty github auth is only useful for --dry-run,
+					// we may want to enforce this as an error
+					log.Warningf("missing authentication for repository %s, and no default provided", url)
```
smacker: 👍

PR #162, opened 2018-08-22
Add Drone CD config

### LINKED ISSUE #160: CD: configure Drone to deploy to staging
A chart https://github.com/src-d/charts/pull/63 and a https://drone.srcd.host/src-d/lookout were created and we want to use those for CD of the latest master to staging env.

This issue is about adding a `.drone.yml` that would read appropriate staging configuration \w desired `config.yaml` in it from `.helm-staging.yml` and configuring Drone deployment to be triggered on push to master.

carlosms:
Fix #160.
Depends on https://github.com/src-d/charts/pull/63.

Question for @rporres: how do we configure the github username and token? do we need to config these secrets somewhere in `.drone.yml` or `.helm-staging.yml`?

Based on the drone config we use for gitbase-web, staging would be updated with each new commit in master.
The only reason I've moved the docker hub upload from travis to drone is for consistency with gitbase-web.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,81 @@
+workspace:
+  base: /go
+  path: src/github.com/src-d/lookout
+
+branches: [master]
+
+# ANCHORS
+
+build: &build
+  image: golang:1.10-stretch
+  commands:
+    - make dependencies
+    - PKG_OS=linux DOCKER_OS=linux DOCKER_ARCH=amd64 make packages
+  debug: true
+
+docker_image: &docker_image
+  group: docker
+  image: plugins/docker
+  registry: docker.io
+  repo: srcd/lookout
+  secrets: [ docker_username, docker_password ]
+  dockerfile: Dockerfile
+  debug: true
+
+helm_deploy: &helm_deploy
+  image: quay.io/ipedrazas/drone-helm:master-9b37211
+  skip_tls_verify: true
+  helm_repos: srcd-charts=https://src-d.github.io/charts/
+  chart: srcd-charts/lookout
+  release: gp
+  tiller_ns: kube-system
+  wait: true
+
+# PIPELINE STEPS
+
+pipeline:
+
+  clone:
+    image: plugins/git
+    debug: true
+
+  # deployment to staging environment when master is pushed
+
+  build_stg:
+    <<: *build
+    when:
+      branch: [master]
+      event: [push]
+
+  docker_image_stg:
+    <<: *docker_image
+    # workaround for bug https://github.com/kubernetes/helm/issues/1707
+    tag: 'commit-${DRONE_COMMIT_SHA:0:7}'
+    when:
+      branch: [master]
+      event: [push]
+
+  helm_deploy_stg:
+    <<: *helm_deploy
+    prefix: STG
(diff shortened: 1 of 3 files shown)


.drone.yml
```diff
@@ -0,0 +1,81 @@
+workspace:
```
jbeardly: The file has increased in 81 lines.

.helm-staging.yml
```diff
@@ -0,0 +1,4 @@
+app:
```
jbeardly: The file has increased in 4 lines.

rporres:
We cannot pass drone-helm the secrets as it may leak it. We faced the same problem before. The chart is prepared to receive either a secret name or the secrets. I will create the secret and come back to you

.helm-staging.yml
```diff
@@ -0,0 +1,4 @@
...
+  lookout:
+    repository: github.com/src-d/lookout
+    logLevel: debug
```
rporres: add here:

```yaml
app:
  lookout:
    secretName: lookout-github-credentials
    dbConnectionString: postgres://lookout:lookout@lookout-postgresql:5432/lookout?sslmode=disable
    bblfshdConnectionString: ipv4://lookout-bblfshd-bblfshd:9432

nodeSelector:
 lookout:
   srcd.host/type: worker

nodeSelector:
 dummyAnalyzer:
   srcd.host/type: worker

providers:
  github:

Comment (carlosms) [reply .helm-staging.yml]: Thanks, done in 9194dd1

**Comment by rporres** [comment]
Looks fine, but let's wait for https://github.com/src-d/charts/pull/63 to be merged

**Comment by bzz** [comment]
CI is failing here :/ 

Is there anything left before merging the chart then?

**Comment by carlosms** [comment]
CI is failing because of a race condition in an integration test, I detected it and will fix it as part of #139. For now a restart will probably fix it

The list of analyzers will be set in `.helm-staging.yml`, using the values we can override in the chart (https://github.com/src-d/charts/blob/master/lookout/templates/lookout-configmap.yaml#L11)
In this case, it should be 
```yaml
deployDummyAnalyzer: false
analyzers:
  - name: ...
    addr: ...
    feedback: ...
```

(@rporres please correct me if I'm wrong)

bzz:
On
> Could you help me understand, where do we set a list of analyzers in here?

from https://github.com/src-d/charts/pull/63/files#diff-23c8315edad45f74027628eb86b32347R17 it seems we just need to add yml section called `analyzers`. 

+HT @rporres

.travis.yml
```diff
@@ -101,11 +101,6 @@ jobs:
       before_install: skip
       script: PKG_OS="darwin" make packages-sdk
       deploy: *deploy_anchor
-    - name: "push image to Docker Hub"
-      stage: release
-      script:
-        - PKG_OS=linux make build
-        - DOCKER_PUSH_LATEST=true make docker-push
```
bzz: Is `docker_image` in Drone pre-configured somewhere to run a `make docker-push`?

carlosms: the push to docker hub from drone is not done with our repo `make`, we use a drone plugin.
- https://github.com/src-d/lookout/pull/162/files#diff-3216dfff0ed3e301453e6799e8c367e2R16
- http://plugins.drone.io/drone-plugins/drone-docker/

PR #164, opened 2018-08-22
Integration tests with `go test`

### LINKED ISSUE #139: Refactor integration test to be run with go test
See https://github.com/src-d/lookout/pull/135 for context.

carlosms:
Fix #139.

Integration tests use `go test`, and most of the error checks have been refactored to use testing assertions.


cmd/server-test/dummy_analyzer_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/stretchr/testify/suite"
+)
+
+type DummyIntegrationSuite struct {
+	suite.Suite
+	ctx  context.Context
+	stop func()
+	r    io.Reader
+	w    io.WriteCloser
+}
```
smacker: lets:
```go
type IntegrationSuite struct {
	suite.Suite
	ctx  context.Context
	stop func()
	r    io.Reader
	w    io.WriteCloser
}

func (suite *IntegrationSuite) sendEvent(json string) {
	_, err := fmt.Fprintln(suite.w, json)
	suite.Require().NoError(err)
}

type DummyIntegrationSuite struct { 
    IntegrationSuite
}

type ErrorAnalyzerIntegrationSuite struct { 
    IntegrationSuite
}

type MultiDummyIntegrationSuite struct {
    IntegrationSuite
}
```

util/cmdtest/cmds.go
```diff
@@ -50,7 +51,7 @@ func StoppableCtx() (context.Context, func()) {
 }
 
 // StartDummy starts dummy analyzer with context and optional arguments
-func StartDummy(ctx context.Context, args ...string) io.Reader {
+func StartDummy(ctx context.Context, require *require.Assertions, args ...string) io.Reader {
```
smacker: maybe make this functions part of `IntegrationSuite` too? Then we won't need to pass `require.Assertions` everywhere

cmd/server-test/dummy_analyzer_test.go
carlosms: :+1: done in e6afbb2.

util/cmdtest/cmds.go
carlosms: We also use these functions in `cmd/sdk-test/`, I don't see a clean way to make it part of both packages :thinking:  Any suggestion?

smacker: define `IntegrationSuite` struct in `cmdtest` package?

Like
```go
type IntegrationSuite struct {
 	suite.Suite
 	ctx  context.Context
 	stop func()
}
```

Then both `ServerIntegrationSuite` and `BblfshIntegrationSuite` can inherit it.

Just make sure it's not overengineering 😅

bzz:
Feedback addressed, but moving back to "In Progress" as rebase seems to be in order and CI fails

util/cmdtest/cmds.go
carlosms: done in bf58993

carlosms:
Rebased on master. It was not easy and I had to squash everything into one commit.

util/cmdtest/cmds.go
```diff
@@ -35,147 +42,118 @@ func init() {
...
-		os.Exit(1)
-	}
+	require.NoError(err, "can't connect to DB")
 
 	_, err = db.Exec("DROP SCHEMA public CASCADE;")
-	noDBErr(err)
+	require.NoError(err, "can't execute query")
 	_, err = db.Exec("CREATE SCHEMA public;")
-	noDBErr(err)
+	require.NoError(err, "can't execute query")
```
bzz: Do you think making it something like "can't execute SQL: '%s'" would increase ability to debug the failure?

PR #172, opened 2018-08-23
Improve DB connection error message

### LINKED ISSUE #159: Add better error message to `initDB` failures
Starting the server without the DB will result in a generic error that doesn't mention the DB:
`dial tcp 127.0.0.1:5432: getsockopt: connection refused`

carlosms:
Fix #159.

I also moved the DB initialization up. Because if we are going to abort on DB error, it does not make sense to establish connections with the analyzers before we know the DB is ok.

### DIFF
--- a/cmd/lookout/serve.go
+++ b/cmd/lookout/serve.go
@@ -91,6 +91,21 @@ func (c *ServeCommand) Execute(args []string) error {
 		return err
 	}
 
+	db, err := c.initDB()
+	if err != nil {
+		return fmt.Errorf("Can't connect to the DB: %s", err)
+	}
+
+	reviewStore := models.NewReviewEventStore(db)
+	eventOp := store.NewDBEventOperator(
+		reviewStore,
+		models.NewPushEventStore(db),
+	)
+	commentsOp := store.NewDBCommentOperator(
+		models.NewCommentStore(db),
+		reviewStore,
+	)
+
 	analyzers := make(map[string]lookout.Analyzer)
 	for _, aConf := range conf.Analyzers {
 		if aConf.Disabled {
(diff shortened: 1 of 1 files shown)


cmd/lookout/serve.go
```diff
@@ -121,21 +136,6 @@ func (c *ServeCommand) Execute(args []string) error {
 		return err
 	}
 
-	db, err := c.initDB()
-	if err != nil {
```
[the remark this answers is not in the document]
bzz: Does not seem like that, no? @carlosms

cmd/lookout/serve.go
```diff
@@ -91,6 +91,21 @@ func (c *ServeCommand) Execute(args []string) error {
 		return err
 	}
 
+	db, err := c.initDB()
+	if err != nil {
+		return fmt.Errorf("Can't connect to the DB: %s", err)
```
bzz: Could you check if that should follow https://github.com/golang/go/wiki/CodeReviewComments#error-strings ?

carlosms: I usually follow that convention, but in this case I capitalized it to make it consistent with the previous messages, like "Can't open configuration file".
Do you think I should I change all 3 messages?

cmd/lookout/serve.go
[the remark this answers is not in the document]
carlosms: ask @lookout-bot, not me!

what I'm more surprised about is that we are posting comments on _the left_ :thinking:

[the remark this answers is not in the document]
smacker: #173

cmd/lookout/serve.go
bzz: IDK, up to you - but consistency wins and this also might be top-level error

carlosms: merged, we can change this later

PR #177, opened 2018-08-23
add style analyzer configuration

rporres:
Signed-off-by: Rafael Porres Molina <rafa@sourced.tech>

### DIFF
--- a/.helm-staging.yml
+++ b/.helm-staging.yml
@@ -1,6 +1,6 @@
 app:
   lookout:
-    repository: github.com/src-d/lookout
+    repository: github.com/src-d/lookout,github.com/bblfsh/uast-viewer
     logLevel: debug
     secretName: lookout-github-credentials
     dbConnectionString: postgres://lookout:lookout@lookout-postgresql:5432/lookout?sslmode=disable
@@ -21,6 +21,3 @@ providers:
 analyzers:
   - name: style-analyzer
     addr: ipv4://lookout-style-analyzer:2000
-
-repositories:
-  - url: github.com/bblfsh/uast-viewer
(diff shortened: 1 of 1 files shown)


bzz:
Enabled it on `bblfsh/uast-viewer`

.helm-staging.yml
```diff
@@ -17,3 +17,10 @@ nodeSelector:
 providers:
   github:
     comment_footer: "_If you have feedback about this comment, please, [tell us](%s)._"
+
+analyzers:
+  - name: style-analyzer
+    addr: ipv4://lookout-style-analyzer:2000
+
+repositories:
```
rporres: This setting is not going to have any effect as this key is not taken into account. What are you trying to achieve here?

bzz: Thank you for notice! That is a way of adding more repositories for lookout to watch ([example](https://github.com/src-d/lookout/blob/master/config.yml.tpl))

rporres: Well, the chart does not currently support that... I will add it to the chart.

carlosms: Actually for some reason the repositories are defined with the positional argument, and this section of the config file is only used to optionally provide the credentials for those repos.
[Looking at the chart](https://github.com/src-d/charts/blob/master/lookout/templates/lookout-deployment.yaml#L38) I believe should be done with `repository` (already up in this file)

```
app:
  lookout:
    repository: github.com/src-d/lookout,github.com/bblfsh/uast-viewer
```

.helm-staging.yml
```diff
@@ -17,3 +17,7 @@ nodeSelector:
 providers:
   github:
     comment_footer: "_If you have feedback about this comment, please, [tell us](%s)._"
+
+analyzers:
```
rporres: should we remove this? Chart is going to apply this configuration now

PR #179, opened 2018-08-23
Fix canonical comment in event.proto

carlosms:
Apply @vmarkovtsev's changes from #174.


smacker:
yes. We definitely need to stop downloading lookout from github in integration tests

PR #182, opened 2018-08-24
SDK CLI: add relevant logs

bzz:
While developing/debugging a new analyzer I found these logs were super helpful in SDK binary.

### DIFF
--- a/cmd/lookout/event.go
+++ b/cmd/lookout/event.go
@@ -46,6 +46,7 @@ func (c *EventCommand) openRepository() error {
 }
 
 func (c *EventCommand) resolveRefs() (*lookout.ReferencePointer, *lookout.ReferencePointer, error) {
+	log.Infof("Resolving to/from references")
 	baseHash, err := getCommitHashByRev(c.repo, c.RevFrom)
 	if err != nil {
 		return nil, nil, fmt.Errorf("base revision error: %s", err)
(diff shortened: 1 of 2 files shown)


cmd/lookout/event.go
```diff
@@ -46,6 +46,7 @@ func (c *EventCommand) openRepository() error {
 }
 
 func (c *EventCommand) resolveRefs() (*lookout.ReferencePointer, *lookout.ReferencePointer, error) {
+	log.Infof("Resolving to/from references")
```
carlosms: should this be not capitalized?

cmd/lookout/event.go
```diff
@@ -118,6 +119,7 @@ func (c *EventCommand) makeDataServerHandler() (*lookout.DataServerHandler, erro
 }
 
 func (c *EventCommand) bindDataServer(srv *lookout.DataServerHandler, serveResult chan error) (*grpc.Server, error) {
+	log.Infof("starging a DataServer at %s", c.DataServer)
```
carlosms: starting*

cmd/lookout/event.go
```diff
@@ -133,6 +135,7 @@ func (c *EventCommand) bindDataServer(srv *lookout.DataServerHandler, serveResul
 
 func (c *EventCommand) analyzerClient() (lookout.AnalyzerClient, error) {
 	var err error
+	log.Infof("start looking for Analyzer at %s", c.Args.Analyzer)
```
carlosms: should this be **starting**, to be consistent with the other logs?

cmd/lookout/event.go
bzz: seems to be fine, as this is a line-oriented logging, no?

bzz:
Feedback addressed in 9245fc4, ready for another round.

cmd/lookout/event.go
carlosms: both are fine to me, I was proposing lowercase for consistency with the other ones

Issue #173, opened 2018-08-23
Comments are posted on incorrect line


smacker:
https://github.com/src-d/lookout/pull/172/files#r212242695

Currently, bot shouldn't post comments on deleted lines at all.

PR #185, opened 2018-08-24
Comments are posted on incorrect line

smacker:
Fix: #173

The logic to calculate position was incorrect. Documentation: https://developer.github.com/v3/pulls/comments/#input

### DIFF
--- a/provider/github/diff.go
+++ b/provider/github/diff.go
@@ -1,9 +1,11 @@
 package github
 
 import (
+	"bufio"
 	"fmt"
 	"regexp"
 	"strconv"
+	"strings"
 
 	"github.com/google/go-github/github"
 	"gopkg.in/src-d/go-errors.v1"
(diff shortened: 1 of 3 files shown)


smacker:
One note: it's still posting comments on context lines. If we want to change it we can do it in a separate commit (is should be easy now).

provider/github/diff.go
```diff
@@ -92,23 +108,101 @@ func (d *diffLines) hunks(file string) ([]*hunk, error) {
...
-var hunkPattern = regexp.MustCompile(`^(@@ -(\d+)(?:,(\d+))? \+(\d+)(?:,(\d+))? @@[^@]*)(?:@@.*|$)`)
+func parseHunkHeader(line string) (*hunk, error) {
+	var (
+		err error
+		h   = &hunk{}
+	)
+
+	matches := hunkPattern.FindStringSubmatch(line)
+	if len(matches) == 0 {
+		return nil, fmt.Errorf("bad hunk format")
```
carlosms: Maybe we could do `fmt.Errorf("bad hunk format: %s", line)`? Here and the next ones

smacker: it will be solved during the resolving of merge conflict. I don't think I should change it in this PR.

PR #186, opened 2018-08-24
Use fixed revisions for integration tests

carlosms:
Integration tests are failing erratically, and I realized some SDK tests use default revisions, HEAD and HEAD^. This means that sometimes the dummy analyzer returns 0 comments, and the tests fail.


carlosms:
Closed in favor of #187

Issue #176, opened 2018-08-23
Empty language attributes while filing a PR


m09:
## Problem

The `language` attribute of the `File`-s objects returned by `lookout` are empty.

## Steps to reproduce

I followed the instructions at https://github.com/src-d/style-analyzer, with the latest binary or the one matching the version of the `server` submodule and in both cases using the `style-analyzer` repo as the repo to analyze. Same result for both binaries.

## Expected behavior

The `File`-s returned by the `lookout` binary have all their fields set.

## Observed behavior

The `File`-s `language` fields are empty as shown in the logs below:

This logging statement of the `FormatAnalyzer`:

    for file in files:
        cls.log.info("%s %s %d", file.path, file.language, len(file.uast.children))

produced the following log:

```
WARNING:9024:SQLAlchemyModelRepository:created a new database at sqlite:////tmp/lookout.sqlite
INFO:9024:run:Created SQLAlchemyModelRepository(db=sqlite:////tmp/lookout.sqlite, fs=/tmp)
INFO:9024:run:Created DataService(0.0.0.0:10301)
/home/mog/.virtualenvs/style-analyzer/lib/python3.6/site-packages/sklearn/ensemble/weight_boosting.py:29: DeprecationWarning: numpy.core.umath_tests is an internal NumPy module and should not be imported. It will be removed in a future NumPy release.
  from numpy.core.umath_tests import inner1d
INFO:9024:run:Created AnalyzerManager(FormatAnalyzer/1)
INFO:9024:run:Created EventListener(0.0.0.0:2000, 1 workers)
INFO:9024:run:Listening 0.0.0.0:2000
INFO:691d:EventListener:new ReviewEvent
INFO:691d:AnalyzerManager:cache miss: FormatAnalyzer
INFO:691d:AnalyzerManager:training: FormatAnalyzer
INFO:691d:FormatAnalyzer:train file:///home/mog/work/style-analyzer 432e45f88fd386eae4c3608beb83b57880ef30ab {'files': <_Rendezvous object of in-flight RPC>}
INFO:691d:FormatAnalyzer:.flake8  0
INFO:691d:FormatAnalyzer:.travis.yml  0
INFO:691d:FormatAnalyzer:Dockerfile  0
INFO:691d:FormatAnalyzer:LICENSE.md  0
INFO:691d:FormatAnalyzer:README.md  0
INFO:691d:FormatAnalyzer:lookout/__init__.py  3
[...]

PR #187, opened 2018-08-24
Tests for GetFiles language

carlosms:
Tests for #176.

The changes are:
**fixed revisions**: Integration tests are failing erratically, and I realized some SDK tests use default revisions, HEAD and HEAD^. This means that sometimes the dummy analyzer returns 0 comments, and the tests fail.
**don't return comments for push events**: The dummy analyzer was returning one fixed comment for each push event, but this is not expected by the server, and we had lots of errors in the logs. Now the dummy analyzer accepts `--files`. When it's enabled, we call `GetFiles` (we were only testing `GetChanges` before), and return comments similar to the ones in the review event.
**add comments about uast & language**: if `--uast` is used, there are comments informing of the response uast and file language

### DIFF
--- a/cmd/server-test/main.go
+++ b/cmd/server-test/main.go
@@ -23,7 +23,7 @@ func main() {
 	cmdtest.ResetDB()
 
 	ctx, stop := cmdtest.StoppableCtx()
-	cmdtest.StartDummy(ctx)
+	cmdtest.StartDummy(ctx, "--files")
 	r, w := cmdtest.StartServe(ctx, "--provider", "json", "dummy-repo-url")
 
 	// make sure server started correctly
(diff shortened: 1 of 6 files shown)


cmd/server-test/main.go
```diff
@@ -59,7 +59,6 @@ func main() {
 		successPushJSON := `{"event":"push", "internal_id": "1", "commit_revision":{"base":{"internal_repository_url":"https://github.com/src-d/lookout.git","reference_name":"refs/heads/master","hash":"4eebef102d7979570aadf69ff54ae1ffcca7ce00"},"head":{"internal_repository_url":"https://github.com/src-d/lookout.git","reference_name":"refs/heads/master","hash":"d304499cb2a9cad3ea260f06ad59c1658db4763d"}}}`
 		sendEvent(w, successPushJSON)
 		grepTrue(r, "processing push")
-		grepTrue(r, "comments can belong only to review event but 1 is given")
```
smacker: hm. We don't have a test for it anymore?
What if an analyzer returns comments for push event? (it can according to api contract)

smacker:
all good but:
- I think we still need a test for push with comments
- CI has failed (because of travis, not code) let's make sure it's green before merge

carlosms:
All these tests check that we are getting comments for push events:

bblfsh_test.go:
- TestPushNoUASTWarning
- TestPushUAST
- TestPushLanguage

dummy_test.go:
- TestPush

What we are not testing anymore is the error message we log when the analyzer returns those comments ("comments can belong only to review event but 1 is given"). Do you think we need a test for that?

smacker:
yes, please. I want to see that we don't try to post such comments anywhere.

carlosms:
Done in 77a3f1f.
I had to change the revisions we use there because the other ones where not producing any comments on push.

PR #188, opened 2018-08-24
Better error message for wrong SDK push revisions

carlosms:
While doing tests for #187 I got `EOF` for `lookout push`.
This PR replaces that with a more helpful message.

### DIFF
--- a/cmd/lookout/push.go
+++ b/cmd/lookout/push.go
@@ -2,6 +2,8 @@ package main
 
 import (
 	"context"
+	"fmt"
+	"io"
 	"time"
 
 	uuid "github.com/satori/go.uuid"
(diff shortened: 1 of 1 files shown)


cmd/lookout/push.go
```diff
@@ -61,6 +62,11 @@ func (c *PushCommand) Execute(args []string) error {
 	for {
 		commit, err := log.Next()
 		if err != nil {
+			if err.Error() == "EOF" {
```
smacker: please, when you check for error don't compare text value of it.
In this case it's `io.EOF` error. So the check should be `if err == io.EOF {`.

carlosms: :+1: done in 77dedb8.

Issue #134, opened 2018-08-13
Create test repository


smacker:
currently, we run tests against src-d/lookout.
it's better to have a special repository for tests that expose simply addressable corner cases (so I don't need to lookup hash in log every time and run diff manually to understand what this test expects).
We also need to test configuration. For that, we need `.lookout.yml` in the repository.

Most probably we can update tests in this task too.

bzz:
Having something similar to https://github.com/git-fixtures/basic  but for the lookout, that is not analyzer-specific would be nice.

>simply addressable corner cases 

Could you please list a few here, just as examples?

smacker:
> Could you please list a few here, just as examples?

from the top of my head:

- PR with delete of a file
- PR with file rename (in separate commit to see if we duplicate comments)
- Empty PR
- PR with unknown language by enry
- PR with unknown language by bblfsh
- PR with empty `.lookout.yml`
- PR with `.lookout.yml` containing unknown analyzer
- PR with `.lookout.yml` and all analyzers disabled
- PR with merge conflict

bzz:
Submodules should be included to mitigate #144 

Tarfile \w bare repository can be included in lookout, to avoid downloading.

carlosms:
One more scenario to test:
- What happens if the github repository gets renamed. We should make sure nothing breaks, and that we don't trigger analysis on already processed events.

bzz:
https://github.com/src-d/lookout-test-fixtures was created, let's use it for now and transfer if necessary later.

smacker:
@carlosms what is the point to have branches from a fork?

here follow-up issues:
https://github.com/src-d/lookout/issues/226
https://github.com/src-d/lookout/issues/228
https://github.com/src-d/lookout/issues/230
https://github.com/src-d/lookout/issues/231

carlosms:
> what is the point to have branches from a fork?

To avoid having only test for PRs that come from the same repo. Maybe we start using a new field from `github.PullRequest` that has some subtle difference. Or maybe it's exactly the same, but then we have fixtures to confirm it.

smacker:
Makes sense. Here is an issue: https://github.com/src-d/lookout/issues/232

smacker:
all basic fixtures are created. Simple go api is implemented.
Complex test cases and more features will be tracked in separate issues.

PR #196, opened 2018-08-27
Allow to disable analyzers with .lookout.yml

### LINKED ISSUE #191: Could not disable lookout for a repository
After adding a `.lookout.yml` https://github.com/bblfsh/uast-viewer/blob/master/.lookout.yml there are still comments from `lookout-bot` https://github.com/bblfsh/uast-viewer/pull/60

Changing `Dummy` to `dummy` also results in unexpected
```
{"app":"lookout","event-id":"4cac6ace4897c8c6ba65b6c1fa7e1da8c3f37d37","event-type":2,"head":"refs/pull/59/head","level":"warning","msg":"analyzer 'dummy' required by local config isn't enabled on server","provider":"github","repository":"git://github.com/bblfsh/uast-viewer.git","source":"server/server.go:234","time":"2018-08-24T17:46:20.333795614Z"}
```

carlosms:
Fix #191.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -245,7 +245,8 @@ func (s *Server) concurrentRequest(ctx context.Context, conf map[string]lookout.
 
 	var wg sync.WaitGroup
 	for name, a := range s.analyzers {
-		if a.Config.Disabled {
+		if a.Config.Disabled || conf[name].Disabled {
+			ctxlog.Get(ctx).Infof("analyzer %s disabled by local .lookout.yml", name)
 			continue
 		}
 


bzz:
👍 do you think it could be worth also having an integration test for this? 

BTW `lookout — There was an error during the analysis` did you have a chance to look into the reason?

carlosms:
> 👍 do you think it could be worth also having an integration test for this? 

Definitely this should have integration tests.
I didn't add any because I think this will be done better when we have the test repository #134. We already have [a comment](https://github.com/src-d/lookout/issues/134#issuecomment-412847756) for PRs with `.lookout.yml` and disabled analyzers.

> BTW `lookout — There was an error during the analysis` did you have a chance to look into the reason?

I see this in the logs:
`posting analysis failed: file not found: cmd/server-test/dummy_analyzer_test.go`

It seems to be failing because this merge is not a _fast-forward_, there are commits in master that are not in this branch. And `dummy_analyzer_test.go` is a file changed in one of those commits in master.
```
lookout fix-disable-analyzers= $ git diff --stat HEAD^
 server/server.go | 3 ++-
 1 file changed, 2 insertions(+), 1 deletion(-)

lookout fix-disable-analyzers= $ git diff --stat origin/master 
 cmd/server-test/dummy_analyzer_test.go |  6 ++----
 cmd/server-test/multi_analyzer_test.go | 15 +++++----------
 dummy/dummy.go                         | 17 +++++++----------
 server/server.go                       |  3 ++-
 4 files changed, 16 insertions(+), 25 deletions(-)
```

I'm not sure if the problem is in the lookout server or the dummy analyzer.
We can create a new issue, or add this to #134. What do you think is better?

bzz:
Makes sense for having a separate issue, otherwise that one will grow infinity. 
It will depend on such repository beeing created, but will include updating it to have a reproducible case WDY?

`"error":"posting analysis failed: file not found: cmd/server-test/dummy_analyzer_test.go","event-id":"0cff484868305255034c987578415cfb1db20869` but if I grep for this envent-id occurs 3077 times 😕 - is that expected

carlosms:
New issue: https://github.com/src-d/lookout/issues/197

As for 
> but if I grep for this envent-id occurs 3077 times 😕 - is that expected?

Yes, we log `event processing failed, skipping...` every few seconds.

PR #211, opened 2018-08-31
Initial implementation for GitHub Apps authentication

### LINKED ISSUE #156: Initial PoC of the Github Application installations
This should help to understand the complexity of task and impact of this changes better.

 - CLI "feature switch" to support Github App token
 - no db changes yet, no tests are expected

this would serve as an input to a better design.

carlosms:
Fix #156.

The server can work with the previous username & token, or with the github application ID and private key. This PR could be merged without affecting current staging.

Things to consider:
- If the initial requests done by `NewClientPoolInstallations` fail lookout will exit. We may want to change this, and keep retrying indefinitely.
- `NewClientPoolInstallations` contains `context.TODO`, we probably want to set timeouts.
- The list of repositories that have the app installed is retrieved on startup only. We may want to have it dynamically updated.
- The positional argument with the list of repositories is ignored when using this auth. We could remove this argument completely, and use the repos in config.yml for user/token.
- Tested manually, see for example https://github.com/carlosms/lookout-test/pull/11

### DIFF
--- a/cmd/lookoutd/serve.go
+++ b/cmd/lookoutd/serve.go
@@ -169,48 +169,78 @@ func (c *ServeCommand) logConfig(conf Config) {
 }
 
 func (c *ServeCommand) initProvider(conf Config) error {
+	switch c.Provider {
+	case github.Provider:
+		if conf.Providers.Github.PrivateKey != "" || conf.Providers.Github.AppID != 0 {
+			return c.initProviderGithubApp(conf)
+		}
+
+		return c.initProviderGithubToken(conf)
+	}
+
+	return nil
+}
+
+func (c *ServeCommand) initProviderGithubToken(conf Config) error {
 	noDefaultAuth := c.GithubUser == "" || c.GithubToken == ""
 	defaultConfig := github.ClientConfig{
 		User:  c.GithubUser,
 		Token: c.GithubToken,
 	}
 
-	switch c.Provider {
-	case github.Provider:
-		urls := strings.Split(c.Positional.Repository, ",")
-		urlToConfig := make(map[string]github.ClientConfig, len(urls))
-		repoToConfig := make(map[string]github.ClientConfig, len(conf.Repositories))
-		for _, repo := range conf.Repositories {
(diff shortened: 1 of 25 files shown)


config.yml.tpl
```diff
@@ -6,6 +6,9 @@ analyzers:
 providers:
   github:
     comment_footer: "_If you have feedback about this comment, please, [tell us](%s)._"
+    # See https://developer.github.com/apps/building-github-apps/authenticating-with-github-apps/
+    # app_id: 1234
+    # private_key: ./key.pem
```
bzz: shall we mention this, and it's ability to serve as a feature-switch, in https://github.com/src-d/lookout#configuration-file ?

bzz:
Rebase seem to be in order

config.yml.tpl
carlosms: I wasn't sure of the documentation for this, since the #156 description calls for a proof of concept.

Do we want to have like an extra section in case you want to test this auth, or do we want to document it as a supported auth method?

carlosms:
Rebased on top of master

config.yml.tpl
bzz: As you are right, it's more of the PoC, I would propose to add only a minimal example of the config that uses this functionality to that section.

Better user-facing doc on supported auth methods, etc could be added later

carlosms: Done in 26681a4.

cmd/lookoutd/serve.go
```diff
@@ -141,48 +141,81 @@ func (c *ServeCommand) Execute(args []string) error {
...
+	if err != nil {
+		return err
+	}
+
+	c.pool = pool
+	return nil
+}
+
+func (c *ServeCommand) initProviderGithubApp(conf Config) error {
+	var pool *github.ClientPool
```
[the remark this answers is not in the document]
carlosms: leftover from code that had the declaration here, and both possible pools in an if/else.
Fixed in a76b577.

carlosms:
I'll create a new issue before merging :+1:

Issue #65, opened 2018-07-26
Change binaries format from tar.gz to gz


vmarkovtsev:
I suggest to change the format of released archives from tar.gz to just gz since we pack single files and using tar seems to be an overkill.

vmarkovtsev:
Oh, I get it, we include the proto files inside... I would release the `lookout` binary separately them - the proto files are not going to be frequently used.

bzz:
Thank you for the feedback!
I'm curious to understand your use-case better - why would you need one, but not the other?

Rationale for having a single archive right now is: Analyzer SDK consists of both, Proto files that define API and a binary that has code, generated from those Proto's.

So in case of changes in API, both need to be release and updated simultaniosly.

Later we plan to add more release artifacts:
 - SDK (proto, binaries)
 - Lookout server binary
 - Lookout Docker image
 - Dummy Analyzer Docker image

But we are totally open to changing this ofc.

vmarkovtsev:
My use case is having other people in the team who download the `lookout` binary and don't care about proto because they are already using the Python SDK.

bzz:
I see! Ok, IDK what is `Python SDK`, but just wanted to make it clear that the binary is coupled with .proto definitions.

If you are ok with situation when after download a new version team would not be able to run lookout binary sometimes as i.e Protos might have changed changed - then we can jsut separate lookout analyzer SDK release artefacts and have 2 archives instead of 1: 
  - binary
  - .protos

Please let us know if that would be ok.

vmarkovtsev:
Python SDK is https://github.com/src-d/style-analyzer/tree/master/lookout/core

After downloading a new version they will not be able to run anything anyway because somebody has to update the Python SDK first.

Your proposal is 👍 to me.

vmarkovtsev:
Ping. It is very annoying by now.

vmarkovtsev:
![](http://4.bp.blogspot.com/-W2JmtfG4Adw/U8KHqPhocHI/AAAAAAAAAMo/KLyqUO79rrw/s1600/puss-in-boots-00-645-75.jpg)

bzz:
sorry for delay, we did not drop the 🏈 ! 

it's just that the plan is to do this right after we finish lookout#200 and lookout#201 as those would require other changes in release artifacts that we want to batch together.

🙏 stay tuned, it's coming soon!

bzz:
Lookout is going to be separated on 2 binaries
[...]

PR #212, opened 2018-08-31
Split lookout binary to deamon and sdk on

bzz:
As a preparation for SDK release we should have 2 binaries, to avoid user confusion:
 - `lookoutd`
 - `lookout-sdk`

TODO:
 - [x] split binaries
 - [ ] update integration tests
 - [ ] change Makefile to build both
 - [ ] update Docker
 - [ ] check if Chart update needed
 - [ ] update docs
 - [ ] change release packaging, with #65


bzz:
@carlosms although it's WIP, putting in code-reivew as it's open for early feedback

carlosms:
LGTM so far.

Maybe we can take this opportunity to remove all the sdk stuff from the main makefile. The same way we have `Makefile.dummy`.

bzz:
>remove all the sdk stuff from the main makefile

this is nice suggestion, but as soon as there is alpha version of SDK - there is going to be a separate task of migrating lookout to the SDK library and I think it shall be take care of under that one.

I.e we would not need to have protoc and generate code at all, so it should just be removed.

bzz:
A preview of the release using this branch is under https://github.com/bzz/lookout/releases/tag/v0.1.4

bzz:
@carlosms ready for review

Makefile
```diff
@@ -80,61 +83,47 @@ GOTEST_INTEGRATION = $(GOTEST) -tags=integration
...
+test-json: clean build-all
 	DUMMY_BIN=$(PWD)/$(DUMMY_BIN) \
 	LOOKOUT_BIN=$(PWD)/$(LOOKOUT_BIN) \
 	$(GOTEST_INTEGRATION) github.com/src-d/lookout/cmd/server-test
 
 # Build sdk client and dummy analyzer
-.PHONY: build-sdk
-build-sdk: $(DUMMY_BIN) $(LOOKOUT_BIN)
+.PHONY: build-all
+build-all: $(DUMMY_BIN) $(LOOKOUT_BIN) $(LOOKOUT_SDK_BIN)
```
carlosms: maybe here we could call `.build`, `make -f Makefile.dummy build`, and `make -f Makefile.sdk build`

Makefile
```diff
@@ -80,61 +83,47 @@ GOTEST_INTEGRATION = $(GOTEST) -tags=integration
 
 # Integration test for sdk client
 .PHONY: test-sdk
-test-sdk: clean-sdk build-sdk
+test-sdk: clean-all build-all
```
carlosms: should this (and `-short`) be part of Makefile.sdk?

bzz: I was thinking about it.. but may be this can be done while migrating to SDK?

I.e I still feel that there is still value in running all tests from short `make test` rather then not forgetting to add `-f ...`. It can be solved by include, etc,  but it's a bit more work than I would like to include here.

So decided to postpone, WDYT?

bzz:
Would you be ok with a8112b9 here for fixing #215 ?

Makefile
bzz: that is the thing: calling `build` here would brake i.e server tests non-linux OSes - default binaries in `./build/bin` are ALWAYS for linux arch, so it will fail to exec with wired error.

Same for dummy. That another reason why only packaging is moved to separate makefile for dummy and sdk binaries, but this is kept for tests.

Makefile
carlosms: :+1: we can rethink this later

carlosms:
Yes, fix for #215 also looks ok

bzz:
rebased after #217 merged

Issue #181, opened 2018-08-24
Add event-id to every log line in Server


bzz:
To assists debugging e.g cases like https://github.com/src-d/lookout/pull/178#issuecomment-415676512 we need to make sure same event-id is set for as many places we log things, as possible.

This issue is only about server.

Example of logs without event-id
 - gRPCs `connection state changed to 'CONNECTING'`, etc
 - request to https://api.github.com/repos/src-d/lookout/pulls cached"
 - http request to /repos/src-d/lookout/pulls","rate-limit":3985
 - etc.

Related to previous work in https://github.com/src-d/lookout/pull/120 as will require refactoring of some parts to propagate the Context.

carlosms:
Instead of opening a new issue, I'll add here two other things discussed via chat:
- Add the build version. Either to all logs, or at least once when the server starts
- Log the config file read, and environment variables. Probably we want to filter out fields with passwords or tokens

smacker:
some parts of the code don't accept context (git service is a good example). That's why there is no `event-id`. We need to refactor it to support context first.

smacker:
maybe we want to add PR id also. Without it, it's hard to find out on which PR lookout failed. As an example lookout failed here https://github.com/src-d/lookout/pull/221#issuecomment-418671220 but I'm still not sure if it's correct error.

carlosms:
Reopening because I have realized we are still not logging the build version.

Issue #199, opened 2018-08-27
Do not attempt to post comments if there were no relevant ones


bzz:
Right now `poster.postPR()` tries to create a comment though GH API event if there were no applicable ones found by an analyzer for a given changes. 

This results in a log below e.g with style-analyzer and https://github.com/bblfsh/uast-viewer/pull/61/files

```
"level":"debug","msg":"comment is out the diff range. analyzer: style-analyzer, file CONTRIBUTING.md, line 1"

"msg":"event processing failed", "error":"posting analysis failed: github api error: POST https://api.github.com/repos/bblfsh/uast-viewer/pulls/61/reviews: 422 Validation Failed [{Resource: Field: Code: Message:}]"

"msg":"event processing failed, skipping..."
```

and leads to setting a user-facing status of the analysis to "failed" in server, when it is actually not.

Expected behavior: such comments are just skipped, analysis succeeds.

Steps to reproduce:

```
git clone git@github.com:bblfsh/uast-viewer.git && cd uast-viewer
git fetch origin pull/61/head:long-line && git checkout long-line
cp <path to lookout>/lookout .
./lookout review --log-level=debug \
  "ipv4://localhost:2000"

# follow https://github.com/src-d/style-analyzer#how-to-run-the-format-analyzer
python3 -m lookout run lookout.style.format -c lookout.yaml --log-level DEBUG
```

Full log.
<details>

```
{"app":"lookout","level":"debug","msg":"comment is out the diff range. analyzer: style-analyzer, file CONTRIBUTING.md, line 1","source":"github/poster.go:180","time":"2018-08-27T18:33:40.440786638Z"}

{"app":"lookout","level":"debug","msg":"http request","source":"github/token_transport.go:77","time":"2018-08-27T18:33:40.441003352Z","url":"https://api.github.com/repos/bblfsh/uast-viewer/pulls/61/reviews","user":"lookout-bot"}
{"app":"lookout","level":"debug","msg":"http request to /repos/bblfsh/uast-viewer/pulls/61/reviews","rate-limit":4045,"reset-at":"2018-08-27T18:59:34Z","source":"github/client.go:200","time":"2018-08-27T18:33:40.65467891Z"}
[...]

PR #220, opened 2018-09-04
Log config on lookoutd serve startup

carlosms:
Part of #181.

DB config left intentionally out, as it may contain the password.
I'm initializing the structs this way to avoid changing a pointer reference accidentally, and to avoid leaking by mistake any future fields that may contain credentials.

The output will be similar to this:
```
[2018-09-04T14:57:52.179461203+01:00]  INFO starting lookoutd conf=main.Config{
  Config: server.Config{
    Analyzers: []lookout.AnalyzerConfig{
      lookout.AnalyzerConfig{
        Name: "Dummy",
        Addr: "ipv4://localhost:10302",
        Disabled: false,
        Feedback: "https://github.com/carlosms-bot/analyzer-comments/issues/new?template=feedback.md",
        Settings: map[string]interface {}{
        },
      },
    },
  },
  Providers: struct { Github github.ProviderConfig }{
    Github: github.ProviderConfig{
      CommentFooter: "_If you have feedback about this comment please, [tell us](%s)._",
    },
  },
  Repositories: []main.RepoConfig{
    main.RepoConfig{
      URL: "https://github.com/carlosms/lookout-test",
      Client: github.ClientConfig{
        User: "",
        Token: "****",
        MinInterval: "",
      },
    },
  },
} options=main.ServeCommand{
  CommonOptions: cli.CommonOptions{
    LogOptions: cli.LogOptions{
      LogLevel: "DEBUG",
      LogFormat: "",
      LogFields: "",
      LogForceFormat: false,
      Verbose: false,
    },
    GrpcMaxMsgSize: 100,
  },
  DBOptions: cli.DBOptions{
    DB: "",
  },
  ConfigFile: "config.yml",
  GithubUser: "",
  GithubToken: "****",
  DataServer: "ipv4://localhost:10301",
  Bblfshd: "ipv4://localhost:9432",
  DryRun: true,
  Library: "/tmp/lookout",
  Provider: "github",
  Positional: struct { Repository string "positional-arg-name:\"repository\"" }{
    Repository: "https://github.com/carlosms/lookout-test",
  },
} source=lookoutd/serve.go:175

```

### DIFF
--- a/cmd/lookoutd/serve.go
+++ b/cmd/lookoutd/serve.go
@@ -24,7 +24,9 @@ import (
 
 	"github.com/golang-migrate/migrate"
 	"github.com/gregjones/httpcache/diskcache"
+	"github.com/jinzhu/copier"
 	_ "github.com/lib/pq"
+	"github.com/sanity-io/litter"
 	"google.golang.org/grpc"
 	"gopkg.in/src-d/go-billy.v4/osfs"
 	"gopkg.in/src-d/go-log.v1"
@@ -82,6 +84,8 @@ func (c *ServeCommand) Execute(args []string) error {
 		return fmt.Errorf("Can't parse configuration file: %s", err)
 	}
 
+	c.logConfig(conf)
+
 	dataHandler, err := c.initDataHandler()
 	if err != nil {
 		return err
@@ -138,6 +142,30 @@ func (c *ServeCommand) Execute(args []string) error {
 
 	ctx := context.Background()
 	return server.NewServer(watcher, poster, dataHandler.FileGetter, analyzers, eventOp, commentsOp).Run(ctx)
+}
+
+func (c *ServeCommand) logConfig(conf Config) {
+	var cCp ServeCommand
+	copier.Copy(&cCp, c)
+
+	cCp.DBOptions.DB = "****"
+	cCp.GithubToken = "****"
+
+	var confCp Config
+	copier.Copy(&confCp, conf)
+
+	for i := range confCp.Repositories {
+		confCp.Repositories[i].Client.Token = "****"
+	}
+
+	lt := litter.Options{
+		Compact: true,
+	}
+
+	log.With(log.Fields{
+		"options": lt.Sdump(cCp),
+		"conf":    lt.Sdump(confCp),
+	}).Infof("starting %s", name)
 }
 
 func (c *ServeCommand) initProvider(conf Config) error {
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -276,6 +276,13 @@
   version = "v1.0.0"
 
 [[projects]]
+  digest = "1:5cb64ae11c6690416028ba76726f101ffc669cbfacf05cdb5607c52bcfff11a0"
+  name = "github.com/sanity-io/litter"
+  packages = ["."]
+  pruneopts = "NUT"
+  revision = "09e3a73d5b65eac2f14578aec8fb74ee9b4f1a5e"
+
+[[projects]]
   digest = "1:6bc0652ea6e39e22ccd522458b8bdd8665bf23bdc5a20eec90056e4dc7e273ca"
   name = "github.com/satori/go.uuid"
   packages = ["."]
@@ -669,6 +676,7 @@
     "github.com/grpc-ecosystem/go-grpc-middleware/logging",
     "github.com/jessevdk/go-flags",
     "github.com/lib/pq",
+    "github.com/sanity-io/litter",
     "github.com/satori/go.uuid",
     "github.com/stretchr/testify/assert",
     "github.com/stretchr/testify/mock",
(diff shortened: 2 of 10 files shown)


carlosms:
Lookout failed with this error. It looks to me like it's because of #199.
 
```json
{
  "app":"lookoutd",
  "error":"posting analysis failed: github api error: POST https://api.github.com/repos/src-d/lookout/pulls/220/reviews: 422 Validation Failed [{Resource: Field: Code: Message:}]",
  "event-id":"fe231036afe456a7be94d814a1ccd03325d0cabc",
  "event-type":2,
  "level":"error",
  "msg":"event processing failed",
  "source":"server/server.go:101",
  "time":"2018-09-04T14:05:11.99550047Z"
}
```

bzz:
SGTM, but would not these logs contain a plenty of newlines?

If we are to us  `master` of litter instead, we should be able to use `litter.Config.Compact = true` to make it line-oriented again.

cmd/lookoutd/serve.go
```diff
@@ -140,6 +143,39 @@ func (c *ServeCommand) Execute(args []string) error {
 	return server.NewServer(watcher, poster, dataHandler.FileGetter, analyzers, eventOp, commentsOp).Run(ctx)
 }
 
+func (c *ServeCommand) logConfig(conf Config) {
+	cCp := ServeCommand{
+		ConfigFile:    c.ConfigFile,
```
smacker: I'm pretty sure sooner or later somebody (read: me) will forget to update the struct here when changing ServeCommand.

How about just to deep-copy it and rewrite necessary fields?
You can use this lib for it for example: https://github.com/jinzhu/copier

carlosms: I think we have to make a compromise here. We either risk forgetting to add new fields, or we risk forgetting to remove new fields with sensitive info.

I thought I'd go with the safest way. But I could change it for the deep copy if you guys think that would be better. We probably won't be adding many fields with credentials or secrets.

smacker: I can see only 2 such new field: private key and queue connection string. But we will definetely have more configuration options. The thing is: it's easy to spot sensetive fields in logs and fix it. Not that easy to spot missing fields.

carlosms:
It's true we output a lot of lines you may not be always interested on, but on the other hand that makes it readable. I personally prefer that verbosity to this compact output:

```
[2018-09-05T13:20:48.486148115+01:00]  INFO starting lookoutd conf=main.Config{C
onfig:server.Config{Analyzers:[]lookout.AnalyzerConfig{lookout.AnalyzerConfig{Na
me:"Dummy",Addr:"ipv4://localhost:10302",Disabled:false,Feedback:"https://github
.com/carlosms-bot/analyzer-comments/issues/new?template=feedback.md",Settings:ma
p[string]interface{}{}}}},Providers:struct{Github github.ProviderConfig}{Github:
github.ProviderConfig{CommentFooter:"_If you have feedback about this comment pl
ease, [tell us](%s)._"}},Repositories:[]main.RepoConfig{main.RepoConfig{URL:"htt
ps://github.com/carlosms/lookout-test",Client:github.ClientConfig{User:"",Token:
"****",MinInterval:""}}}} options=main.ServeCommand{CommonOptions:cli.CommonOpti
ons{LogOptions:cli.LogOptions{LogLevel:"info",LogFormat:"",LogFields:"",LogForce
Format:false,Verbose:false},GrpcMaxMsgSize:100},DBOptions:cli.DBOptions{DB:""},C
onfigFile:"config.yml",GithubUser:"",GithubToken:"****",DataServer:"ipv4://local
host:10301",Bblfshd:"ipv4://localhost:9432",DryRun:true,Library:"/tmp/lookout",P
rovider:"github",Positional:struct{Repository string "positional-arg-name:\"repo
sitory\""}{Repository:"https://github.com/carlosms/lookout-test"},analyzers:map[
string]pb.AnalyzerClient{},pool:nil}
```

But I could also go with it if you feel strongly about it.

bzz:
No strong preference. Do you think multiline output will not confuse the log collection then? That's the only thing I had in mind

carlosms:
With the log format set to json, as we do for staging, it will just be a string with a bunch of "\n".
But just in case we end up with a system that needs each log in one line, I changed it in dc61fff.

cmd/lookoutd/serve.go
carlosms: That's the problem, sensitive fields should not be easy to spot in the logs :D
Anyway, I made the change in cbe937d, let's keep an eye on new fields to make sure we don't leak anything.

smacker: haha. I mean it's easy to spot during development. You would see it when you run the server locally and kill it.

carlosms:
rebased, squashed, and merged as soon as CI is green

Issue #218, opened 2018-09-04
Migrate lookout to SDK


bzz:
lookout-sdk has been released https://github.com/src-d/lookout-sdk/releases/tag/v0.0.3

Lookout should remove protos, generated code and automation to do so in favor of using `gopkg.in/src-d/lookout-sdk.v0`.

This might also requite to update https://github.com/src-d/lookout-gometalint-analyzer to make sure it works after the migration.

smacker:
We also need to update documentation. (move it to sdk repo? partially?)

But I would better track it in the separate issue because I'm not sure which parts of it should be still present in lookout and which in sdk.

Here is an issue for that: https://github.com/src-d/lookout/issues/242

PR #221, opened 2018-09-05
Use sdk

smacker:
Lookout part of #218

Gopkg constraint uses revision instead of version. We will need to update it as soon as we release v0.0.4. (I believe we need to fix python first)


smacker:
lookout failed. not sure if it's the right error because there is no PR id. But most probably it is:
```
{
"app":"lookoutd",
"error":"posting analysis failed: file not found: provider/github/poster.go",
"event-id":"73adba9b793078abb3e887bfc70d1dab4d4ea1dd",
"event-type":2,
"level":"error",
"msg":"event processing failed",
"source":"server/server.go:101",
"time":"2018-09-05T09:40:29.442824782Z"
}
```

.travis.yml
```diff
@@ -10,13 +10,13 @@ git:
...
+      - ubuntu-toolchain-r-test
     packages:
-    - gcc-6
-    - g++-6
+      - gcc-6
+      - g++-6
 
 go:
-  - "1.10.x"
+  - '1.10.x'
```
bzz: Why ' ?

bzz:
Looking at vendor update made me wonder - is there a reason we do not prune non-go files?

.travis.yml
smacker: actually, my editor automatically reformatted the file and I didn't notice.
But in general, the difference between single quotes and double quotes is you don't need to escape `\`.
I can revert if you wish.

smacker:
It wasn't pruned in prototype and we just continued using it like that. Maybe it's related to the need of proto files but even so we can enable pruning globally and disable for `google/protobuf` and `gogo` if we need. (but most probably we don't)

.travis.yml
bzz: 👍

smacker:
Issue for prune: https://github.com/src-d/lookout/issues/222

PR #238, opened 2018-09-06
Remove CLI positional argument

### LINKED ISSUE #229: Consider removing the CLI positional argument
The positional argument is not needed at all for the GitHub App auth, we take the list of repos using the API. It should at least be optional.

For the user & token auth we also have a list of repositories in `config.yml`. This is used to configure individual options for each repo, but the list of repos to watch is taken from the positional argument.

This is confusing, and we should probably remove the CLI argument completely and take the list of repos from `config.yml` in the case of user&token auth.

Context in #211.

smacker:
Fix #229 

It looks like chart should work without `repository` value.
Chart would use default one but it will be just ignored by lookout now.

### DIFF
--- a/cmd/lookoutd/serve.go
+++ b/cmd/lookoutd/serve.go
@@ -256,21 +256,14 @@ func (c *ServeCommand) initPoster(conf Config) (lookout.Poster, error) {
 func (c *ServeCommand) initWatcher(conf Config) (lookout.Watcher, error) {
 	switch c.Provider {
 	case github.Provider:
-		urls := make([]string, len(conf.Repositories))
-		for i, repo := range conf.Repositories {
-			urls[i] = repo.URL
-		}
-
-		watcher, err := github.NewWatcher(c.pool, &lookout.WatchOptions{
-			URLs: urls,
-		})
+		watcher, err := github.NewWatcher(c.pool)
 		if err != nil {
 			return nil, err
 		}
 
 		return watcher, nil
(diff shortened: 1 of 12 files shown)


cmd/lookoutd/serve.go
```diff
@@ -262,8 +256,13 @@ func (c *ServeCommand) initPoster(conf Config) (lookout.Poster, error) {
...
 	switch c.Provider {
 	case github.Provider:
+		urls := make([]string, len(conf.Repositories))
+		for i, repo := range conf.Repositories {
+			urls[i] = repo.URL
+		}
+
 		watcher, err := github.NewWatcher(c.pool, &lookout.WatchOptions{
-			URLs: strings.Split(c.Positional.Repository, ","),
+			URLs: urls,
```
carlosms: Doing a quick read it looks like we don't really use this `WatchOptions.URLs`, only for one log line in `NewWatcher`.
And this `URLs` are taken from the config file, so it will not match the actual URLs we will watch if we use the GitHub App auth, right?

Can we just get rid of it?
The list of repos to watch will be logged as part of the config file for user & token; and with GitHub App auth the logs we already have in `NewClientPoolInstallations`.

smacker: I removed `WatchOptions` completely

bzz:
README.md still contains the mentions of `<repository>` CLI arg

Issue #234, opened 2018-09-06
switch to lookout-test-fixtures in tests


smacker:
- [x] server integration test
- [ ] sdk integration test (need to clone repo first and later use repo from tar file)
- [ ] unit tests
- [ ] remove `git: depth: false` from travis

smacker:
looks like we don't use lookout repo in unit tests

PR #243, opened 2018-09-07
Use fixtures for server tests

smacker:
First part of #234


carlosms:
LGTM, but the code needs to be updated after the merge of https://github.com/src-d/lookout-test-fixtures/pull/15. Well, I guess because of `Gopkg.lock` the PR would work as it is, but it would be good to use the current master.
Hopefully we just need to change `longLineFixture.CommitRevision` to `GetCommitRevision()`.

smacker:
- rebased on master
- updated fixtures
- updated code to use new api

Tests were failing locally due to broken cache. But there is a bug in cache remover. It doesn't remove directory recursively. I fixed it in this commit: https://github.com/src-d/lookout/pull/243/commits/906127f8f5f51aeeb19877a64600fa64d2df2809

Please let me know if you think it's better to extract this commit in separate PR.

smacker:
👍 Issue: https://github.com/src-d/lookout/issues/249

PR #244, opened 2018-09-07
Don't walk to parent dirs looking for repository in lookout-sdk

smacker:
I had an error with resolving revision and spent quite much time figuring out why.

Currently lookout-sdk binary would walk through parents until it finds the root of a repository.
And it does that even if a passed path doesn't exist.

It means any incorrect path with git repo somewhere in the middle of it will be resolved to that repo. It's super confusing.

In my case, I forgot to add `/` at the front of the `--git-dir` so it was resolving to the current repo.

I also added a path to the repo in error message.

### DIFF
--- a/cmd/lookout-sdk/event.go
+++ b/cmd/lookout-sdk/event.go
@@ -39,10 +39,16 @@ func (c *EventCommand) openRepository() error {
 	var err error
 
 	c.repo, err = gogit.PlainOpenWithOptions(c.GitDir, &gogit.PlainOpenOptions{
-		DetectDotGit: true,
+		// it's useful to walk to parent in case --git-dir is default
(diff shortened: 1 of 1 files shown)


carlosms:
I see your point, but also I think it's nice to be able to run the binary from anywhere in the target repo.
Could we disable the parent dir search only if the `--git-dir` option is used?

smacker:
huh?
There is already check for that: https://github.com/src-d/lookout/pull/244/files#diff-e48fe7868eacfe2f466d798dd862fe57R42

If you don't pass `--git-dir` option it works as before.

carlosms:
You are right, I don't know how I managed to misunderstand that :D

PR #245, opened 2018-09-10
Improve log fields

### LINKED ISSUE #181: Add event-id to every log line in Server

carlosms:
Fix #181.

The main change in this PR is that the callbacks with `EventHandler` accept the context, and keep the log fields like the PR number.

The `service/git` methods now get the log from the context, but the problem here is that the gRPC requests come from the analyzer, and we don't have the `event-id` or any other field.
I didn't look much into it, so I don't know if there is a way to pass some ID fields to the analyzer and get them back when they call `GetFiles/GetChanges` without modifying the `proto` messages.

### DIFF
--- a/cmd/lookoutd/serve.go
+++ b/cmd/lookoutd/serve.go
@@ -371,7 +371,7 @@ func (c *ServeCommand) initDB() (*sql.DB, error) {
 				"Use '%s migrate' to upgrade your database", dbVersion, version, build, maxVersion, name)
 	}
 
-	log.Debugf("the DB version is up to date, %v", dbVersion)
+	log.With(log.Fields{"db-version": dbVersion}).Debugf("the DB version is up to date")
 	log.Infof("connection with the DB established")
 	return db, nil
 }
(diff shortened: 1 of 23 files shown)


smacker:
> The service/git methods now get the log from the context, but the problem here is that the gRPC requests come from the analyzer, and we don't have the event-id or any other field.
> I didn't look much into it, so I don't know if there is a way to pass some ID fields to the analyzer and get them back when they call GetFiles/GetChanges without modifying the proto messages.

Grpc allows to send metadata in [header](https://godoc.org/google.golang.org/grpc#Header) or [trailer](https://godoc.org/google.golang.org/grpc#Trailer). There are some helpers in [go-grpc-middleware](https://github.com/grpc-ecosystem/go-grpc-middleware) for that.
We can include it in our sdk later to pass `event-id` to an analyzer and send it back.

carlosms:
That's great. I created an issue with your suggestion so we don't forget about it: https://github.com/src-d/lookout/issues/248

Issue #227, opened 2018-09-06
Update the list of GitHub App installations dynamically


carlosms:
For the GitHub App authentication.
The list of repositories that have the app installed is retrieved on startup only. We should make it to update dynamically.

Context in #211.

PR #246, opened 2018-09-10
Dynamic repos

smacker:
Fix: #227 

- `ClientPool` isn't static anymore. Now it has `Update` and `RemoveClient` methods.
- `Watcher` now listens to the `ClientPool` changes and starts/stops goroutines.
- `Watcher` was refactored a little to remove `getClient` methods which would cause race conditions.
- New struct `Installations` holds a list of clients and syncs it with github

### DIFF
--- a/cmd/lookoutd/serve.go
+++ b/cmd/lookoutd/serve.go
@@ -33,8 +33,6 @@ import (
 	yaml "gopkg.in/yaml.v2"
 )
 
-var installationsSyncInterval = time.Hour
-
 func init() {
 	if _, err := app.AddCommand("serve", "run server", "",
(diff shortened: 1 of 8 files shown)


cmd/lookoutd/serve.go
```diff
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"time"
```
smacker: test

smacker: test

smacker:
Yes. We should move to webhooks soon. Or it will become a monster.

Issue #180, opened 2018-08-23
Add health-check endpoints


bzz:
Server and Analyzer should expose an HTTP-level endpoints that can be used as kubernetes life-probes.

As [discussed](https://github.com/src-d/minutes/pull/309/files#diff-d707b85ec18906cb2d5bedb7373d595aR47), at this point probes could check accessibility of it's dependencies like DB, Bblfsh, etc.

Eventually, it would be really nice for such checks to become more complex and execute some basic scenarios and verify them pass (like a simple live unit-tests).

For now:
 - [ ] check what kind of life probes k8s supports/recomends
 - [ ] add endpoints to Server and Dummy analyzer
 - [ ] update charts \w life probes

smacker:
@carlosms We started a discussion in 2 different places: PR and slack. I think it's better to move it here.

The discussions touch different topics, I'll try to summarize them here and give my opinion. I'll start from the "end", specific issues, and move to more general topics.

### Proper handling of restarts. Especially posting comments on github.

The main argument about it I noticed: restarting on the middle of results posting would cause problems.

Restart is a very common problem in software development. No matter do they caused intentionally or not. That's why almost all systems already have solutions for it. For databases it's transaction. For a queue it's publish confirmations. Those concepts allow to handle restarts in the middle of something be safe. Rollback would kick and an application can retry the same operation after the restart.

HTTP APIs are usually designed with this issue in mind also. Retry would cause specific error which can be easily handled by the application. Or API would provide some kind of transaction/confirmation also.

In our particular case "posting comments" github provides such API which allows retry without any undesired effects for an end user (like duplicated comments).
If we create [review](https://developer.github.com/v3/pulls/reviews/#create-a-pull-request-review) without `body` attribute, it will be visible to an user only after the [submit call](https://developer.github.com/v3/pulls/reviews/#submit-a-pull-request-review).

It's safe to call `submit` multiple times.

To make my point more clear that restart in the middle of posting isn't a problem here is a list of changes how we might improve our poster in the future:
- Call `create` without body if don't have ID of review in the state already
- Save returned ID to the state
- In case of failure, we can just retry later without any issues for an end user
- Send confirm
- Update status in DB
- In case of failure, we can just safely retry later

### Should we consider DB as a critical component of the system or not

The current argument:
> Maybe a restart is the better option from our point of view, but from management it's uptime... maybe we prefer to keep posting analysis while the DB is down, and deal with the inconsistent state somehow later when the DB comes up again

[...]

PR #252, opened 2018-09-13
Add http health probes

carlosms:
Part of #180. The other half will be done in a PR to https://github.com/src-d/charts/.

I consider this PR a skeleton to make it easy to add more complex logic later, if needed.

For now the health checks are basically empty.
The reason is that for any initialization problem we already exit, which is picked by kubernetes as a trigger to restart. And we probably don't want kubernetes to restart `lookoutd` just because a dependency, like bbblfsh or one of the analyzers, is not reachable, as that will not solve the problem.

The code for the serve and dummy analyzer is very similar now, but I decided not to extract it to a common place because that would probably make it harder to customize individually later.

`lookoutd` listens on 8090, and the dummy analyzer on 8091 to make it easy to start both servers while developing.

### DIFF
--- a/cmd/dummy/main.go
+++ b/cmd/dummy/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"net/http"
 
 	"github.com/src-d/lookout"
 	"github.com/src-d/lookout/dummy"
@@ -22,9 +23,12 @@ type ServeCommand struct {
 	DataServer       string `long:"data-server" default:"ipv4://localhost:10301" env:"LOOKOUT_DATA_SERVER" description:"gRPC URL of the data server"`
 	RequestUAST      bool   `long:"uast" env:"LOOKOUT_REQUEST_UAST" description:"analyzer will request UAST from the data server"`
 	RequestFilesPush bool   `long:"files" env:"LOOKOUT_REQUEST_FILES" description:"on push events the analyzer will request files from HEAD, and return comments"`
+	ProbesAddr       string `long:"probes-addr" default:"0.0.0.0:8091" env:"LOOKOUT_ANALYZER_PROBES_ADDRESS" description:"TCP address to bind the health probe endpoints"`
 }
 
 func (c *ServeCommand) Execute(args []string) error {
+	c.initHealthProbes()
+
 	var err error
 	c.DataServer, err = grpchelper.ToGoGrpcAddress(c.DataServer)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


carlosms:
> I didn't know we exit on any major error.

I was referring the to the obvious blocking initialization errors like parsing the config file https://github.com/src-d/lookout/blob/535f62f1df3eafe56819d8c3d3c5db1be8468af9/cmd/lookoutd/serve.go#L80
or error connecting to the DB https://github.com/src-d/lookout/blob/535f62f1df3eafe56819d8c3d3c5db1be8468af9/cmd/lookoutd/serve.go#L343

There might be other errors that we want to include for the readiness/liveness, but for now I just wanted to submit a PR with a basic foundation.

> > And we probably don't want kubernetes to restart lookoutd just because a dependency, like bbblfsh
> 
> I believe we want. On restart as long as deps aren't reachable lookout will be stay in crash loop backoff which is much safer in terms of consistency of data and unexpected behavior than working in unhealthy state.

:+1: I can see arguments for both options. For example we may want to keep working without bblfsh, have the analysis that need bblfsh fail gracefully, and then use whatever retry policy we define to process the PRs that failed later. This would allow to keep working for repos/analyzers that do not require bblfsh uast, and would probably avoid problems like restarting on the middle of results posting (and missing the chance to write to the DB that the PR was processed).
Definitely it's something to discuss IRL or slack.

cmd/dummy/main.go
```diff
@@ -22,9 +23,12 @@ type ServeCommand struct {
 	DataServer       string `long:"data-server" default:"ipv4://localhost:10301" env:"LOOKOUT_DATA_SERVER" description:"gRPC URL of the data server"`
 	RequestUAST      bool   `long:"uast" env:"LOOKOUT_REQUEST_UAST" description:"analyzer will request UAST from the data server"`
 	RequestFilesPush bool   `long:"files" env:"LOOKOUT_REQUEST_FILES" description:"on push events the analyzer will request files from HEAD, and return comments"`
+	ProbesAddr       string `long:"probes-addr" default:":8091" env:"LOOKOUT_ANALYZER_PROBES" description:"TCP address to bind the health probe endpoints"`
```
rporres: This env name should have a more descriptive name. If it expects and address, then I would use something like `LOOKOUT_ANALYZER_PROBES_ADDRESS`

Another thing I don't understand well is why default is in a different format as `LOOKOUT_DATA_SERVER`

cmd/lookoutd/serve.go
```diff
@@ -50,9 +51,11 @@ type ServeCommand struct {
 	DryRun      bool   `long:"dry-run" env:"LOOKOUT_DRY_RUN" description:"analyze repositories and log the result without posting code reviews to GitHub"`
 	Library     string `long:"library" default:"/tmp/lookout" env:"LOOKOUT_LIBRARY" description:"path to the lookout library"`
 	Provider    string `long:"provider" default:"github" env:"LOOKOUT_PROVIDER" description:"provider name: github, json"`
+	ProbesAddr  string `long:"probes-addr" default:":8090" env:"LOOKOUT_PROBES" description:"TCP address to bind the health probe endpoints"`
```
rporres: ditto

cmd/dummy/main.go
carlosms: The different format is because `DATA_SERVER` is for a grpc url, this one is the address to bind the http server.

I'll change it to `LOOKOUT_ANALYZER_PROBES_ADDRESS` and `0.0.0.0:8091` to make it more clear.

carlosms: Done in 80c65d3.

PR #271, opened 2018-09-19
Correctly convert interfaces to pb.Struct

### LINKED ISSUE #258: Max length option for go analyzer doesn't work
Comment from the bot: https://github.com/src-d/lookout/pull/257#discussion_r217738347
Configuration: https://github.com/src-d/lookout/blob/master/.lookout.yml#L7

We need to debug why and fix.

smacker:
Should fix #258


bzz:
There type of issues was exactly the reason behind initial proposal of having a very simple KV structure for configuration.

Issue #53, opened 2018-07-24
Propose advanced schema for HistoryDB


bzz:
#35 introduces a simple storage for input (events) and output (comments) of the analysis.

This issue is about proposing additional data model, needed for further operations of lookout server on Github.
 
Things that one might want to persist (not necessarily all in this DB)
 - installations: users, enabling/disabling lookout service on their repositories
 -  [GH App access tokens](https://developer.github.com/apps/building-github-apps/authenticating-with-github-apps): YOUR_APP_ID, private key, JWT tokens (10 min expiration), `YOUR_INSTALLATION_ACCESS_TOKEN` (GH API access token, got with JWT, 1h expiration)
 - m.b some state of analyzers?

bzz:
SQL DB changes to integrate:
  - which analyzer posted a comment (so 2 analyzers can post comment on the same line)
  - save PRs/Events separately

smacker:
### Discarded changes

GH installations: it would be nice to store them somewhere later but for now we are fine with keeping it in memory and sync with GH from time to time.

GH tokens: the library we use [keeps tokens in the transport](https://github.com/bradleyfalzon/ghinstallation/blob/master/transport.go#L35) and doesn't allow to change it. It's fine for now. Later when we move to distributed mode we might want to fork it and keep token in some distributed cache to reduce number of calls to github. Similar to [travis](https://github.com/travis-ci/travis-github_apps/blob/master/lib/travis/github_apps.rb#L46)

Analyzers: we don't have any information about analyzer for now beside the config file. I don't think we need to save to db our own configuration.

### Changes

I propose to make as small changes as possible to address current needs:

- extract information about pull request to separate table. We need it to be able to check should we process PR or post comment (they will use different ids). Because we want to be provider agnostic to call the table `review_target`.
- add analyzer name to comment table. I checked the code and we don't really need this field for our code but I see how valuable this field can be in retrospective.

Here is SQL diff. If we are ok with it I'll proceed with PR.

```diff
 CREATE TABLE review_event (
        id uuid NOT NULL PRIMARY KEY,
        status text NOT NULL,
-       provider text NOT NULL,
        internal_id text NOT NULL,
        created_at timestamptz NOT NULL,
        updated_at timestamptz NOT NULL,
@@ -11,20 +10,20 @@ CREATE TABLE review_event (
        source jsonb NOT NULL,
        merge jsonb NOT NULL,
        configuration jsonb NOT NULL,
[...]

PR #275, opened 2018-09-20
New db schema

smacker:
Second PR for #53

- data migration
- use only new fields
- kallax has problems with joins. Though it's a temporary workaround. After we switch to new eventID queries will become simple again.


smacker:
Travis fails on generation migration as expected (first we change code then apply migration).
Next PR will introduce migration with deleted columns.

PR #281, opened 2018-09-20
Docs

### LINKED ISSUE #229: Consider removing the CLI positional argument

dpordomingo:
fix #229

I took some time to give some love to the docs,
let me know if it's ok/meh/no-please; comments, suggestions and/or rants are **very welcomed** !!!!

here is the rationale I followed:

## README.md
- as our guidelines says [about the readme](https://github.com/src-d/guide/blob/master/engineering/documentation.md#readmemd), it should be as short as possible, and use it as an index to other docs; to do so, I moved some parts to the `docs/configuration.md` file
- I focused in install and ussage sections; the readme could be shorter if we move the "without docker" section to a new place, ie. `docs/usage.md` (would you prefer it? it could explain different ways to run it, and mention some parameters needed to run each piece separatelly)

## other docs
- in order to have our docs in one place, I moved the `CONTRIBUTING.md` into `docs` dir (as GitHub also uses it, see: [setting-guidelines-for-repository-contributors](https://help.github.com/articles/setting-guidelines-for-repository-contributors/#adding-a-contributing-file))
- `docs/configuration.md` provides all the info about how to configure lookout with `config.yml` and the ways to configure the Authentication
- `docs/sdk` -> **WIP**: I'll merge the sdk docs we have now, and we decide later where to store them.


docs/configuration.md
```diff
@@ -0,0 +1,139 @@
...
+  github:
+    comment_footer: "_If you have feedback about this comment, please, [tell us](%s)._"
+    # app_id: 1234
+    # private_key: ./key.pem
+    # installation_sync_interval: 1h
+```
+
+`comment_footer` key defines a format-string that will be used for custom messages for every message posted on GitHub; see how to [add a custom message to the posted comments](#custom-footer)
+
+<a id=basic-auth></a>
```
smacker: `<a id=basic-auth></a>` maybe it shouldn't be here...

dpordomingo: I do not like to use anchors using the autogenerated ids because they can be easily broken if the title is eventually changed.
But if you really prefer to go with autogenerated ones, just confirm it and I'll use them :+1:

smacker: Nah. Up to you. It was just a suggestion.

README.md
```diff
@@ -1,125 +1,106 @@
...
-Clone the repository, or download [`./docker-compose.yml`](./docker-compose.yml)
 
 # Usage
 
-To trigger the analysis on an actual pull request of a GitHub repository you will need [GitHub access token](https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/).
+The **lookout** configuration is defined by the `config.yml` file; you can use the template [`config.yml.tpl`](config.yml.tpl) to create your own config file. You will find more information about it in the [docs about how to configure lookout](docs/configuration.md)
+
+
+To trigger the analysis on any pull request of a GitHub repository you will need a GitHub authentication as it is described in the [docs about how to authenticate with GitHub](docs/configuration.md#basic-auth)
+
```
bzz: Probably a nit-pick, but 2 sentences above does not end with period.

carlosms:
One more thing that would be good to document (maybe in another PR, I'm not saying you should do it): a table with the options and the env vars. Even if this is visible with `--help`, it's good to have the reference in the docs, especially when using docker.

README.md
dpordomingo: wow!! that was neat!
thanks for it.

smacker:
+1 to both points made by Carlos

docs/configuration.md
```diff
@@ -0,0 +1,139 @@
...
+  - url: github.com/<user>/<repo2>
+    client:
+      # user: github-user
+      # token: github-user-token
+      # minInterval: 1m
+```
+
+If you're using [Authentication as a GitHub App](#github-app), the list of repositories to be watched will be taken from the GitHub installations.
+
+Due to the [bug#277](https://github.com/src-d/lookout/issues/277) **lookout** can parse only public repositories.
```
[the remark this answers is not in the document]
dpordomingo: Would you agree on keeping it, and adding to the existing issue a subtask to update this docs?
That way we won't forget updating it :)

docs/configuration.md
```diff
@@ -0,0 +1,139 @@
...
+<a id=custom-footer></a>
+## Add a custom message to the posted comments
+
+You can configure **lookout** to add a custom message to every comment that each analyzer returns. This custom message will be created following the rule:
+```
+sprinf(providers.github.comment_footer, feedback)
+```
+If any of both is not defined, the custom message won't be added.
+
+example:
```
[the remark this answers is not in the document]
dpordomingo: Woow, I need a Title capitalization checker.
Hmmmm...
I could even create an Analyzer for doing it

dpordomingo:
@carlos, about the table with the params, defaults and env values... But then I realized that is quite good detailed with the `--help` option. Do you think it would be also useful here? If all of you think so, I'd transpile the whole `--help`. Or wouldn't it be necessary all the `lookoutd` options?

carlosms:
The commits from this PR were used (squashed as one commit) in #294.
Let's close this one, and continue the improvements and discussion in new issues.

Thanks for the effort @dpordomingo!

Issue #259, opened 2018-09-14
Add support for incremental update of PR


smacker:
We should re-process PRs if they have new commits.
Blocked by https://github.com/src-d/lookout/issues/53

PR #285, opened 2018-09-21
Add support for incremental update of PR

smacker:
Part of #259

- update sdk
- update fixtures
- migration with new column and data migration (it's not safe to have them in the same PR if we would have production but OK for staging. Redeployment process is fast and the chance of new PR during it is very low)
- change logic to use new column
- update unit and integration tests

After merge and deployment, we will need 2 more PR to delete `OldInternalID` (unused now). Just delete it from code, re-run kallax code generation, merge. And then create new migration and merge.
Though there is no hurry with it.


Gopkg.toml
```diff
@@ -74,7 +74,9 @@
 
 [[constraint]]
   name = "gopkg.in/src-d/lookout-sdk.v0"
-  version = "0.0.4"
+  # there is no release of sdk with new ID for review events yet
+  # version = "0.0.4"
+  revision = "e06a088af4fe40561700c057c5dacd46e23b2abe"
```
bzz: @carlosms @dpordomingo would it make sense to release SDK first, as we discussed couple of days ago, and then put 0.0.5 here?

carlosms: makes perfect sense, yes.

bzz: https://github.com/src-d/lookout-sdk/releases/tag/v0.1.0 🚀

smacker:
sdk version updated. I'm merging after CI is green.

Issue #267, opened 2018-09-18
go-queue for incoming events


bzz:
Queue implementation https://github.com/src-d/go-queue should be used for decoupling review creating and it's processing inside the lookout.

Default configuration should work as is, without extra configuration so `lookoutd serve` runs using in-memory implementation. Other environments will use RabbitMQ.

Deployment envs would use something like Borges does
 - `lookoutd enqueue` responsible only for interacting \w Github and putting new Review events in queue
 - `lookout dequeue` process events from the queue sequentially, transactionally updating DB with event state, posting results back, etc

Most probably, should include few queues: 
 - queue for new events
 - queue for re-processing failures

carlosms:
Current status of the steps we defined in the initial discussions:

- [x] go-queue: `serve` uses a memory queue, new `watch` and `work` commands use RabbitMQ
- [ ] Add a memory cache in watcher to avoid sending repeated jobs; current polling retrieves all the github events and sends them to the queue again and again.
- [ ] parallel processing of jobs in worker, with config parameter to control how many goroutines to use. Right now each event is processed sequentially.
- [ ] docs (#315)
- [ ] open issue to request changes in the staging deployment charts.

carlosms:
All tasks initially planed are merged, I think we can close this one.
Future improvements will have new issues.

PR #290, opened 2018-09-25
Distributed watcher and worker using a queue

carlosms:
Part of #267.

First version of the queue integration.
The `serve` command should work as before. It uses the memory queue implementation, while 2 new sub-commands, `watcher` and `worker`, use RabbitMQ.

`cmd/lookoutd/serve.go` has been refactored to `cmd/lookoutd/common.go`, and the watcher is managed outside of `server/server.go`.

Next steps, that can be done later after this PR is accepted:
- Cache in watcher to avoid sending repeated jobs; current polling retrieves all the github events and sends them to the queue again and again.
- parallel processing of jobs in `worker`, with config parameter to control how many goroutines to use. Right now each event is processed sequentially.
- docs (#315)
- open issue to request changes in deployment charts.

Quickstart to test:
- Run dependencies:
`docker-compose up -d --force-recreate postgres  bblfsh`
- Run RabbitMQ. To monitor it, go to http://localhost:8081, enter with `guest/guest`
`docker run -d --hostname rabbitmq --name rabbitmq -p 8081:15672 -p 5672:5672 rabbitmq:3-management`
- Run `lookoutd watcher` and `lookoutd worker`.


bzz:
Re-started UnitTests on CI

cmd/server-test/dummy_analyzer_test.go
```diff
@@ -20,11 +20,27 @@ func (suite *DummyIntegrationSuite) SetupTest() {
 
 	suite.StoppableCtx()
 	suite.StartDummy("--files")
-	suite.r, suite.w = suite.StartServe("--provider", "json",
-		"-c", dummyConfigFile)
 
-	// make sure server started correctly
-	suite.GrepTrue(suite.r, "Starting watcher")
+	if suite.IsQueueTested() {
```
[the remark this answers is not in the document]
bzz: This is actually a valid concern - first useful comment from a 🤖! 🎉 

@carlosms let's may be refactor these: 
 - https://github.com/src-d/lookout/pull/290/files#r220130326
 - https://github.com/src-d/lookout/pull/290/files#r220130325

util/cli/queue.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/src-d/lookout/util/queuehelper"
+	"gopkg.in/src-d/go-queue.v1"
+	_ "gopkg.in/src-d/go-queue.v1/amqp"
+	_ "gopkg.in/src-d/go-queue.v1/memory"
+)
+
+// queueWindow defines the maximum number of unacknowledged jobs that the
+// queue iterator will allow to retrieve. A number of 1 will mean that each
+// job is processed sequentially
+// TODO: because the eventHandler is called synchronously, a value greater than
```
bzz: nit pick, but it's usually a good idea to save reader some time blaming the changes and put `TODO(<name>) `

util/cli/recoverer.go
```diff
@@ -0,0 +1,31 @@
...
+			err := watcher.Watch(ctx, eventHandler)
+			errCh <- err
+		}()
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case err := <-errCh:
+			if err != nil {
+				ctxlog.Get(ctx).Errorf(err, "error during watcher.Watch, restarting watcher")
```
bzz: let's make sure this log will include an event-id, similar to what we have in https://github.com/src-d/lookout/pull/290/files#diff-91bbeda7eb98a7adc57b9e47e2cf5c2bR42

util/cli/recoverer.go
```diff
@@ -0,0 +1,31 @@
+package cli
```
bzz: `recoverer.go` file name seems quite confusing, why not `run_watcher.go` instead?

util/queuehelper/queue.go
```diff
@@ -0,0 +1,48 @@
+package queuehelper
```
bzz: Usually having lots of "helpers" is an anti-pattern of the design. Is there a reason not to put it inside top level `queue.go` ?
 :pray: `queuehelper` -> `queue`, so `queuehelper.QueueJob` -> `queue.Job`

Gopkg.toml
```diff
@@ -64,9 +64,10 @@
   name = "gopkg.in/src-d/enry.v1"
   version = "1.6.4"
 
+# go-queue does not work with the latest release, v1.2.0
 [[constraint]]
   name = "github.com/satori/go.uuid"
-  version = "1.2.0"
+  branch = "master"
```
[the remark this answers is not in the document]
carlosms: it's already there, please see line 67

util/cli/queue.go
carlosms: I agree, but that seems to me that applies to when you have some half-baked code that you plan to add later.
In this case, however, I wanted to mark future work to be done by whoever, with a detailed explanation of what is missing and how to proceed. I'd prefer not to mark this as reserved for me, when anybody could implement the missing functionality.

bzz: I see the confusion you are having - it's quite common. This comment *does not claim* the issue ownership, it just attributes it to particular author who has the best context over it.
 
Same rules of code ownership are applied as anywhere else, so anyone can do that.

Please read more about it - it's very well explained in https://google.github.io/styleguide/cppguide.html#TODO_Comments

That is a normal engineering best-practice and it's up to you of course to follow it or not.
I just suggest to follow.

Gopkg.toml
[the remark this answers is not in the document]
bzz: [Gopkg.toml:67](https://github.com/src-d/lookout/pull/290/files/b656cca62b815b931d44e59b12d9d4776fc74c75#diff-836546cc53507f6b2d581088903b1785R67) 👍

.travis.yml
```diff
@@ -47,16 +47,14 @@ jobs:
       script:
         - make prepare-services
         - psql -c 'create database lookout;' -U postgres
-        - make build
-        - ./build/bin/lookoutd migrate
```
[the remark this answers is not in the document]
carlosms: no, we already reset the DB and run `migrate` in the tests, https://github.com/src-d/lookout/blob/master/util/cmdtest/cmds.go#L138

util/cli/recoverer.go
carlosms: We don't have an event here, and the error may even happen outside of the context of an individual event.
Could you please be more specific how this could be improved?

carlosms:
quoting @bzz:

> 1. What will happen when connection to queue fails? Let's make sure we have some retries with back-off

Right now the `lookoutd watch/work` commands will exit immediately. I think it makes sense to handle it this way, if we assume a distributed env. will be run with kubernetes, which will handle the restart policy.

>  2. On the API side 
> ```go
> go qOpt.RunEventDequeuer(ctx, server.HandleEvent)
> return cli.RunWatcher(ctx, watcher, qOpt.EventEnqueuer(ctx)
> ```
> 
> looks a bit confusing, do you think we could have something a bit more consistent at the high-level, so a client could do
> ```go
> go RunEventDequeuer(ctx, qOpt, server.HandleEvent)
> go RunEventEnquerere(ctx, watcher, qOpt.EventEnqueuer(ctx))
> ```
> and wait? Or a similar one with Watcher/Worker nouns instead.
> 

:+1: Changed in cf4f75d

> ## Consistency
> Possible confustion: `lookoutd serve` starts ... a server (cmd/server-test)? a worker?
> 
> 3 possible processes we have:
>  - watcher (producing changes to Queue from GH/mailing list
>  - worker  (consuming/processing changes from Queue)
>  - server (both+inmem queue)
> 
> 3 commands to start them
>  - lookoutd watch
>  - lookoutd work
>  - lookoutd serve
> 
> Implications:
>  => may be we could rename `processingCommand` -> queueConsumingCommand? That would be consistent with above
>  => let's keep CLI consistent as well: either all verbs (watch, server, work), or all nouns (worker,server,watcher).
>     I vote for verbs.

Done in 8bf1494, 1177c90.

> ## Documentation
> Let's update the README to explain and reflect the structure of commands from above.
> 
> Configuration: LOOKOUT_BROKER, LOOKOUT_QUEUE should be included in docs.

:+1: Already defined as next steps in the description. I will create a new PR for that.

> ## Infra
> Let's create an issue with request of Helm chart changes for Infra: multiple process + cfg update.

:+1: Already defined as next steps in the description. I'd prefer to wait until we merge, this way we can point them to the latest docker image uploaded by drone, so they can test the new commands as they create the new charts.

> ## Last but not least 
> 
> this PR is huge `+28,183 −579` 
> 
> It would really simplify the review if refactoring with extraction of common interfaces is separated from actual new code implementing new features to 2 different PRs. 

Yes it's a big one. I tried to keep a commit history that makes sense to help with the review. I'll try to create those 2 new PRs from this one, but I don't see clearly how to do it cleanly, so that may take a while.

util/queuehelper/queue.go
carlosms: Top level `queue.go`, in package `lookout`, created an import cycle.
I've moved it to `./queue/queue.go` in c6397e4.

util/cli/queue.go
carlosms: done in d668151

cmd/server-test/dummy_analyzer_test.go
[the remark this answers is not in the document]
carlosms: :+1: done in f98f7c3

Gopkg.lock
```diff
@@ -480,6 +504,24 @@
   revision = "f21a4dfb5e38f5895301dc265a8def02365cc3d0"
   version = "v0.3.0"
 
+[[projects]]
+  digest = "1:d905cf8d320c391e840d90ed7f545470cba5de75fc543b19b1193cca52acecbf"
+  name = "google.golang.org/appengine"
```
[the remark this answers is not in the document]
bzz: @carlosms please investigate and if that one is not necessary for our runtime operations - let's not bring it in.

bzz:
Awesome! Shall it be rebase on latest master as well?

Few things that I think are still left before merging this are:

1. [Queue connection](https://github.com/src-d/lookout/pull/290#issuecomment-425474996)
   Sounds good. It would be nice if you could confirm that this behavior is consistent between queue, DB and Bblfsh services.
2. [Error logging](https://github.com/src-d/lookout/pull/290#discussion_r220896795)
   We need to make sure Watcher error logs are easily attributed back to GH events that caused an error.
3. [Dependencies on AppEngine](https://github.com/src-d/lookout/pull/290#discussion_r221597303)
   Remove un-used dependency after identifying how did it get in/

carlosms:
> Shall it be rebase on latest master as well?

The conflict is minimal, I'd rather keep the history to keep track of the review comments, and I'll rebase and squash before merging.

> 1. [Queue connection](https://github.com/src-d/lookout/pull/290#issuecomment-425474996)
>    Sounds good. It would be nice if you could confirm that this behavior is consistent between queue, DB and Bblfsh services.

It is consistent with the DB, we exit immediately if we can't connect:
https://github.com/carlosms/lookout/blob/91fdd8c89688cbeef9b29c1aeb8e20557645fa16/cmd/lookoutd/work.go#L42

But for bblfsh we don't try to connect until we get a `GetChanges` with `wantUAST`. I guess it might make sense to try to ping it also on startup. I propose to open an issue to discuss this further.

> 2. [Error logging](https://github.com/src-d/lookout/pull/290#discussion_r220896795)
>    We need to make sure Watcher error logs are easily attributed back to GH events that caused an error.

As I put in a comment there, we don't have event details here, and the error may even happen outside of the context of an individual event. For example, here:
https://github.com/carlosms/lookout/blob/91fdd8c89688cbeef9b29c1aeb8e20557645fa16/provider/github/watcher.go#L178

When the error is caused in the context of an individual event, the error is logged here before we return it:
https://github.com/carlosms/lookout/blob/91fdd8c89688cbeef9b29c1aeb8e20557645fa16/server/server.go#L41

What I've done is move some of the log fields to this context logger, so it is more detailed, in 2ce24d4

> 3. [Dependencies on AppEngine](https://github.com/src-d/lookout/pull/290#discussion_r221597303)
>    Remove un-used dependency after identifying how did it get in/

After looking a bit into it, I think we don't have any choice here. From the [dep docs](https://golang.github.io/dep/docs/Gopkg.lock.html), we should not edit the `.lock` directly:
> Gopkg.lock is autogenerated; editing it manually is generally an antipattern

And we already have enabled all the available [prune](https://golang.github.io/dep/docs/Gopkg.toml.html#prune) options:
https://github.com/carlosms/lookout/blob/91fdd8c89688cbeef9b29c1aeb8e20557645fa16/Gopkg.toml#L1

util/cli/queue.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// EventEnqueuer returns an event handler that pushes event to the queue.
+// InitQueue must be called before.
+func (c *QueueOptions) EventEnqueuer(ctx context.Context) lookout.EventHandler {
```
smacker: I would better put this function and `RunEventDequeuer` in `queue/queue.go`.

What do you think if we keep `cli` package responsible only for handling command line arguments and move the actual logic to proper packages inside lookout?

carlosms: :+1: change done in 73f0f7f

PR #294, opened 2018-09-26
Doc improvements

### LINKED ISSUE #286: Clarify what scopes are needed for GitHub token
We currently point to the instructions on how to create a GitHub token, but are any scopes necessary?
Which ones should be set for the bot to work correctly?

carlosms:
This PR includes @dpordomingo's commits from #281 plus a few mine on top.
They cover the feedback from https://github.com/src-d/lookout/pull/281#pullrequestreview-157617865, and the docs to fix #286.


docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,17 @@
+# Contribution Guidelines
```
bzz: Allthough it's not mentioned anywhere in https://github.com/src-d/guide/blob/master/engineering/documentation.md#contributingmd

I belive it's customary to put this in the root dir of the project.

README.md
```diff
@@ -1,125 +1,162 @@
...
+[![GoDoc](https://godoc.org/gopkg.in/src-d/lookout?status.svg)](https://godoc.org/github.com/src-d/lookout)
+[![Code Coverage](https://codecov.io/github/src-d/lookout/coverage.svg)](https://codecov.io/github/src-d/lookout)
+[![Go Report Card](https://goreportcard.com/badge/github.com/src-d/lookout)](https://goreportcard.com/report/github.com/src-d/lookout)
+![development](https://svg-badge.appspot.com/badge/stability/development?color=D6604A)
+=======
 
 A service for assisted code review, that allows running custom code Analyzers on pull requests.
 
-# SDK
+# Configuring lookout
```
bzz: I suggest moving this section down in ToC, right before `# SDK for Analyzer Developers`

docs/configuration.md
```diff
@@ -0,0 +1,144 @@
...
+For a production environment you can use **lookout** as a [GitHub App](https://developer.github.com/apps/about-apps/).
+
+To do so, you must also unset any environment variable or option for the GitHub username and token authentication.
+
+You need to create a GitHub App following the [documentation about creating a GitHub app](https://developer.github.com/apps/building-github-apps/creating-a-github-app/). The permissions that must be set are:
+
+- Repository metadata: Read-only
+- Pull requests: Read & write
+- Single file: Read-only
+- Commit statuses: Read & write
```
bzz: I belive this will solve #286 👍

README.md
carlosms: But in that case, you go to the easier use case, running with docker compose, and it mentions `config.yml`.

And it's actually a requirement, there is not default for this file. I think it makes sense to have it right in the beginning of the document, so you know there is a reference for this somewhere.

docs/CONTRIBUTING.md
carlosms: Apparently in docs it will also be [picked up by github](https://help.github.com/articles/setting-guidelines-for-repository-contributors/).

But I don't care one way or another, I'll move it back to the root folder.

carlosms:
Thanks for the review, feedback applied in 59c33d7.

README.md
bzz: Up to you, but the reason for suggestion is - a first time user rarely wants to configure something, he usually needs to understand how to run it first, and then configure

carlosms: I agree with the reasoning. In this case since it's just a link to the config guide, I think it makes sense to keep it so as a first-time user you know that the reference exists, and can come back later.

zurk:
Yes, now it is evidently better, thank you!
just wonder where is `# Authenticate as a GitHub App` section?

carlosms:
We moved it to the configuration guide, https://github.com/src-d/lookout/blob/master/docs/configuration.md#authentication-as-a-github-app

Issue #304, opened 2018-09-28
panic in git scanner


smola:
Executing on gitbase-web repository:

```
$ lookout-sdk review -v ipv4://localhost:2000 --from=HEAD^^^^^^^^^^^  --to=HEAD
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xa9eda6]

goroutine 75 [running]:
github.com/src-d/lookout/service/git.(*regexpFilter).Fn(0xc420c7b4a0, 0x0, 0x0, 0x0, 0xc420c8ed80)
        /home/travis/gopath/src/github.com/src-d/lookout/service/git/scanner.go:172 +0x26
github.com/src-d/lookout/service/git.NewChangeFilterScanner.func1(0xc420aaafa0, 0xc420aaafa0, 0xc420db76d0, 0xc420bc9a20)
        /home/travis/gopath/src/github.com/src-d/lookout/service/git/scanner.go:215 +0x38
github.com/src-d/lookout.(*FnChangeScanner).Next(0xc42142d500, 0xc420e8c000)
        /home/travis/gopath/src/github.com/src-d/lookout/data.go:288 +0x99
github.com/src-d/lookout.(*FnChangeScanner).Next(0xc42142d540, 0x0)
        /home/travis/gopath/src/github.com/src-d/lookout/data.go:286 +0x5a
github.com/src-d/lookout.(*FnChangeScanner).Next(0xc42142d580, 0xc420205300)
        /home/travis/gopath/src/github.com/src-d/lookout/data.go:286 +0x5a
github.com/src-d/lookout.(*FnChangeScanner).Next(0xc42142d5c0, 0xc420205300)
        /home/travis/gopath/src/github.com/src-d/lookout/data.go:286 +0x5a
github.com/src-d/lookout/service/bblfsh.(*ChangeScanner).Next(0xc420246a00, 0xc420205330)
        /home/travis/gopath/src/github.com/src-d/lookout/service/bblfsh/bblfsh.go:141 +0x45
github.com/src-d/lookout.(*DataServerHandler).GetChanges(0xc420c0d1e0, 0xc42142d140, 0xf77720, 0xc421002ab0, 0x0, 0x0)
        /home/travis/gopath/src/github.com/src-d/lookout/data.go:86 +0x151
[...]

Issue #261, opened 2018-09-18
exclude_pattern crashes lookout server


m09:
## Context

I had some problems when trying to run an analyzer that uses `exclude_pattern` in `FilesRequest` and `ChangesRequest`.

## Steps to reproduce

- clone the lookout-sdk fork

        git clone git@github.com:m09/lookout-sdk.git
- install lookout-sdk (in a venv probably)

        pip install lookout-sdk
- run lookout-sdk server

        python3 language-analyzer.py
- run lookout server

        lookout-sdk review -v ipv4://localhost:2021

## Expected behavior

Analyzer should run with default behavior (comment language and number of functions for each file)

## Observed behavior

Crashes with following log

```
$ lookout-sdk review -v ipv4://localhost:2021
[2018-09-18T10:20:47.88145688+02:00]  INFO resolving to/from references app=lookout-sdk source=lookout-sdk/event.go:55
[2018-09-18T10:20:47.923477623+02:00]  INFO starting a DataServer at ipv4://localhost:10301 app=lookout-sdk source=lookout-sdk/event.go:128
[2018-09-18T10:20:47.923596478+02:00]  INFO starting looking for Analyzer at ipv4://localhost:2021 app=lookout-sdk source=lookout-sdk/event.go:144
[2018-09-18T10:20:47.925695782+02:00]  INFO processing pull request app=lookout-sdk head=HEAD provider= repository=file:///home/mog/work/js/freeCodeCamp source=server/server.go:117
[2018-09-18T10:20:47.92580276+02:00]  INFO getting .lookout.yml app=lookout-sdk head=HEAD provider= repository=file:///home/mog/work/js/freeCodeCamp source=server/server.go:197
[2018-09-18T10:20:47.925897331+02:00]  INFO validating refs: [internal_repository_url:"file:///home/mog/work/js/freeCodeCamp" reference_name:"HEAD" hash:"5cfa1cd78c4db237d050ea70fa41406b343ed430" ], validateRefName: false app=lookout-sdk head=HEAD provider= repository=file:///home/mog/work/js/freeCodeCamp source=git/service.go:36
[2018-09-18T10:20:47.925950972+02:00] DEBUG load trees for references: [{file:///home/mog/work/js/freeCodeCamp HEAD 5cfa1cd78c4db237d050ea70fa41406b343ed430}] app=lookout-sdk head=HEAD provider= repository=file:///home/mog/work/js/freeCodeCamp source=git/service.go:137
[...]

PR #310, opened 2018-10-03
Include regexp patterns in dummy requests

carlosms:
While testing #304 I added regexp to each requests made by the dummy analyzer.
These should not have any impact on the results, but it will help to uncover big problems like #261 in our integration tests.


smola:
Why not a test case instead?

carlosms:
Regexp are already covered by the unit tests, and we included one specifically for #261:
https://github.com/src-d/lookout/blob/master/service/git/scanner_test.go#L151

I see this more like an extra step, and a convenience to more comprehensive manual tests while developing.

PR #311, opened 2018-10-09
remove old internal id for review event

smacker:
part of #259

CI should fail on code generation check (code has been changed but migration will be created only in the next PR for correct deployment)


carlosms:
Please help me understand what will be in that next PR you mention?
If it's just a commit updating the generated kallax migrations, I'd rather add it to this PR to avoid having failing CI on master.

smacker:
Next commit will contain migration.

If we include it here, migration will be applied before new code deployed and the server would start crashing trying to read from column that doesn't exist anymore.

Issue #274, opened 2018-09-20
Add the ability to specify configuration in CLI


vmarkovtsev:
We have the ability to send the configuration for an analyzer in our protocol.
However, the CLI for `review` and `push` does not provide a way to specify it.
I request the feature to send arbitrary JSON configuration in those two commands. `--config` I guess.

m09:
Duplicate of #269 I think

vmarkovtsev:
This one is about an ad-hoc CLI argument, not about reading `.config.yml`.

m09:
Right, I thought you meant reading JSON from a file :)

Issue #269, opened 2018-09-18
lookout-sdk binary doesn't pick up on config.yml


m09:
It seems that this config file is `lookoutd` specific.

smacker:
why do you need `config.yml` in sdk? It contains configuration specific for `lookoutd`. Or are you talking about `.lookout.yml`?

m09:
I need a `config.yml` in the sdk to have a single source of configuration. Atm I need to copy configuration to all the repositories I analyze which is sub-optimal workflow. I need in particular the `settings` section of `config.yml` when I use `lookout-sdk` and those settings don't change between repos.

smacker:
Got it. But the `config.yml` defines multiple analyzers. Sdk works only with one defined by positional argument. 

Maybe we can add a flag to sdk to read file with `settings` section only. Would it work for you?

@carlosms @bzz wdyt?

m09:
It would be great yes! Just a clarification though, a single analyzer argument in `lookout-sdk` can still ask for several configurations (we have analyzers that are lists of analyzers in `style-analyzer` and each one of them will request its config even though it's exposed as one single analyzer).

dpordomingo:
@m09 was your requirement fully satisfied with the current `--config-json` option?
Do you still miss being able to use the same `config.yml` than when running `lookoutd`?

m09:
It's good enough config-json for me.

dpordomingo:
Since there is no special necessity on passing config in any other way than though `--config-json` we'll close this as agreed by [minutes/419::R87](https://github.com/src-d/minutes/pull/419/files#diff-82137befbf83aa65593b6378ccb848cfR87)

PR #312, opened 2018-10-09
Add config-json flag to lookout-sdk

smacker:
Fix: #274

We might also implement reading config file as requested in #269 but this option feels more "curl-like" as "-d" flag. And the original idea was to make lookout-sdk similar to curl.


carlosms:
LGTM, but this needs some docs.

smacker:
I added this flag to sdk README. Where else should I add it?

PR #315, opened 2018-10-09
Add distributed lookout docs

carlosms:
Part of #267.

First version of the docs. Still pretty basic, later we can expand, add some diagrams.
We may also want to use the docker images to run the `watch/work` commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ A service for assisted code review, that allows running custom code Analyzers on
          * [Installation](#installation)
          * [Dependencies](#dependencies)
          * [Quickstart](#quickstart)
+      * [Running in Distributed Mode](#running-in-distributed-mode)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -108,6 +109,12 @@ Following these steps you will be able to run separately the **lookout** depende
     $ lookoutd serve --dry-run
     ```
 
+## Running a Distributed lookout
```
smacker: maybe in distributed mode? but I'm fine with current word header too.

smola: `Running in Distributed Mode` would be slightly better, even if it's only because capitalization here looks incorrect (even if it is correct because it's using English heading capitalization + lookout in lowercase as it is in branding, but looks inconsistent).

carlosms: :+1: Change made; squashed and rebased.

PR #317, opened 2018-10-10
Queue memory cache for watcher

carlosms:
Part of #267.

On the initial discussion for the go-queue integration we mentioned the idea of adding a memory cache to the watcher.

The motivation was that the watcher is polling github, and enqueuing the same events over and over again. These events are then consumed by the worker, which goes to the DB and sees that this event was processed and can be discarded. Thus, a totally unnecessary  enqueue & dequeue.

The main commit 5583dda adds a `sync.Map` cache for event IDs. After implementing it I didn't like the idea of it growing indefinitely, so I added a cleanup in 5246e04.
With the cleanup an event will be enqueued at most once every 24 hours, until that event is not seen by the watcher anymore (i.e. the PR was closed).

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -712,7 +712,7 @@
   version = "v1.0.1"
 
 [[projects]]
-  digest = "1:2b78a7efa35fc4bba7c099e3a19b54b9c62bfd3dd040786bef0bd90f1a2f0f71"
+  digest = "1:81e67057a8a099050b28dae9eaa390d29cdb5e266d84271366c30314c0e93e6a"
   name = "gopkg.in/src-d/go-queue.v1"
   packages = [
     ".",
@@ -720,8 +720,8 @@
     "memory",
   ]
   pruneopts = "NUT"
-  revision = "3eb0eac8758069e896ea90ae8cf8afb41f1dedc8"
-  version = "v1.0.3"
+  revision = "6214d25507193fb3f3f011d29ed75cee98bd84b0"
+  version = "v1.0.4"
 
 [[projects]]
   digest = "1:e532b733db09e0d561bd3d10a3c50198234aee2ff289211f3c1f6f44d9aeaf1f"
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -153,6 +153,17 @@
   version = "v1.0.0"
 
 [[projects]]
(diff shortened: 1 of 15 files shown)


queue/queue.go
```diff
@@ -81,10 +98,22 @@ func EventEnqueuer(ctx context.Context, q queue.Queue) lookout.EventHandler {
...
+			cache.Store(e.ID(), true)
+			delayeCacheCleanup(cache, e.ID(), cleanupDelay)
+		}
+
 		return nil
 	}
 }
 
+func delayeCacheCleanup(cache *sync.Map, id lookout.EventID, cleanupDelay time.Duration) {
+	go func() {
```
smacker: for a big number of events it would create lots of goroutines. Which isn't really a big problem (goroutines are cheap) but still better don't abuse this fact.

Standard approach for this task is to create one goroutine that cleans old values with some interval. In our case you can store timestamp instead of the `true` in map.

Example in one of the most popular cache lib in go: https://github.com/patrickmn/go-cache/blob/master/cache.go#L1071

smacker:
I would also better use some library that implements LRU cache instead of handling it manually and send duplicated events "sometimes".

smola:
FYI: we use https://github.com/hashicorp/golang-lru at gitbase, but feel free to pick whatever suits the needs here.

cmd/lookoutd/serve.go
```diff
@@ -87,7 +87,7 @@ func (c *ServeCommand) Execute(args []string) error {
 	}()
 
 	go func() {
-		enqErrCh <- c.runEventEnqueuer(ctx, qOpt, watcher)
+		enqErrCh <- c.runEventEnqueuer(ctx, qOpt, watcher, false)
```
smacker: why don't you want to apply cache here too?

carlosms: I figured we could simplify and skip this step for the memory queue.
But, at the same time, it could save us from going to the DB constantly to check for already processed events, so... should we enable it too?

smacker: Up to you, but I don't like having lots of log with `skipping` that we have right now. The cache will remove it and will stop calling DB all the time. We may also remove this flag from `runEventEnqueuer` and always use decorator inside (less code!).

carlosms: OK, done in fa55b9a

carlosms:
friendly ping @smola :)

PR #318, opened 2018-10-10
Update to bblfsh 2.9, use -drivers image

carlosms:
Use bblfsh `-drivers` image in docker-compose, update travis bblfsh to 2.9.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,8 +33,8 @@ jobs:
       script: make test-coverage codecov
     - name: 'SDK Integration Tests Linux'
       script:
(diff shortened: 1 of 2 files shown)


smacker:
why don't we use the same `drivers` image in travis?

smacker:
Travis downloads cache too :-D Though 480mb is a lot. Thanks for the clarification!

smacker:
I mean when you use travis cache it gets uploaded to a cloud and on each build travis downloads it. So there is no much difference if you download from dockerhub or use travis cache. It's going to be downloaded from amazon s3/google cloud storage anyway.

carlosms:
I see what you mean now :+1:

PR #321, opened 2018-10-12
Concurrent queue job processing

carlosms:
Part of #267.

Done on top of #317, the relevant commit is 69e9ae2.

The semaphore was needed because we use the `memory` queue for `lookoutd serve`, and it [ignores the advertised window](https://godoc.org/github.com/src-d/go-queue/memory#Queue.Consume).

If you think the concurrency management should not be handled here, I see 2 options to simplify the code:
- Make the memory queue in src-d/go-queue respect the advertised window. Probably something similar to the semaphore code here.
- Stop using the memory queue for `lookoutd serve`, and go back to call `Server.HandleEvent` directly as the handler for the watcher.

If we didn't use the memory queue the code could be simpler, something like:
```go
	iter, _ := q.Consume(concurrent)
	for {
		consumedJob, err := iter.Next()

		go func(consumedJob *queue.Job) {
			var qJob QueueJob
			consumedJob.Decode(&qJob)
			event, _ := qJob.Event()
			eventHandler(ctx, event)

			consumedJob.Ack()
		}(consumedJob)
	}
```

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -712,7 +712,7 @@
   version = "v1.0.1"
 
 [[projects]]
-  digest = "1:2b78a7efa35fc4bba7c099e3a19b54b9c62bfd3dd040786bef0bd90f1a2f0f71"
+  digest = "1:81e67057a8a099050b28dae9eaa390d29cdb5e266d84271366c30314c0e93e6a"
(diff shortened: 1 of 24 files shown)


carlosms:
Closing in favor of #322.

Issue #287, opened 2018-09-25
Improve observability / error messages


campoy:
Whenever something fails it's very hard to tell why.

For instance, if an analyzer is missing its address the error message is simply "scheme not supported" without any references to what scheme we're talking about.

Similarly logs are not always useful, since there's a very large repetition of pretty useless continuous logs:

```bash
lookout_1   | time="2018-09-24T21:37:33.3683477Z" level=info msg="event successfully processed, skipping..." app=lookoutd event-id=7c712b2e295cda99f96d93a07c009f07e471a28f event-type=2
```

Let's log only things that are meaningful, this line should be `level=debug` if anything.

carlosms:
Note: include the comments about log level from this PR: https://github.com/src-d/lookout/pull/326

PR #326, opened 2018-10-19
Use github auth to fetch git contents

### LINKED ISSUE #277: Lookout can not work with private repositories
**lookout** can not process private repositories.

It seems to fail at clone phase when it's trying to fetch the repository config from `.lookout.yml`

I ensured that the provided user/token is valid for that user/repo, and the whole app works if the repo is made public again.

```text
ERROR event processing failed app=lookoutd
error=Can't get .lookout.yml in revision: repository not found
{git://github.com/dpordomingo/testing-repo.git refs/pull/54/head 2873cc62558ef48ea506d1c7b87c3a70bcf02989}
event-id=85b1edcc40dc34bdd6e901d513100cb7192f451f
event-type=2 pr-id=216969605 pr-number=54
repo=https://github.com/dpordomingo/testing-repo
```

carlosms:
Fixes #277.

A couple of notes:
- In `syncer.go` I change the clone url from `git://` to `https://` because the token auth is http based only. I left a TODO to move this change to `lookout-sdk`, it can be easily done later if this PR is approved.
- The Client `gitAuthFn` needs to be a function because for GH Apps the tokens expire, so it needs to be refreshed every time we do a fetch. A cleaner approach would be creating a new type of `http.AuthMethod`, but it's not possible because it has [unexported methods](https://github.com/src-d/go-git/blob/master/plumbing/transport/http/common.go#L163)


service/git/syncer.go
```diff
@@ -60,35 +80,44 @@ func (s *Syncer) Sync(ctx context.Context,
...
 	}
 
-	return s.fetch(ctx, frp.InternalRepositoryURL, r, refspecs)
+	return s.fetch(ctx, repoInfo, gitRepo, refspecs)
 }
 
-func (s *Syncer) fetch(ctx context.Context, repoURL string, r *git.Repository, refspecs []config.RefSpec) (err error) {
-	ctxlog.Get(ctx).Infof("fetching references for repository %s: %v", repoURL, refspecs)
+func (s *Syncer) fetch(ctx context.Context, repoInfo *lookout.RepositoryInfo, r *git.Repository, refspecs []config.RefSpec) (err error) {
+	ctxlog.Get(ctx).Infof("fetching references for repository %s: %v", repoInfo.CloneURL, refspecs)
```
smacker: maybe debug here?

carlosms:
Since the log levels are not really new code from this PR, I'll merge and then take care of better log messages in general in #287

Issue #324, opened 2018-10-18
sdk README inconsistency


vmarkovtsev:
sdk/README.md says:

> All the needed .proto files are located in the /sdk directory.

However, they are not.

carlosms:
Thanks, closing this one as it will get fixed as part of #330.
We were in the process of fixing this in #242 when Davis had to take the leave.

Issue #330, opened 2018-10-24
Move documentation to GitBook


carlosms:
This issue is to fulfill the OKR "Get lookout production ready", which calls for:
- Extend lookout end user documentation in gitbook
- Extend lookout developer documentation in gitbook

PR #334, opened 2018-10-29
Update README.md

cclulu:
clear indication that the .proto files are located in a different repository. This is to address issue #324


carlosms:
Thanks, but we'll make a big revamp of the docs on #330 and this will be fixed as part of that.

PR #336, opened 2018-11-01
Clean docker

gy741:
Hello,

This prevents extra packages from being installed, which helps reduce

the change we'll take an unknown dependency, such as `ca-certificates`. 

And, The use of COPY is preferred over ADD when additional functionalities of ADD are not being used (like spraying the contents of a tar in a folder).

Ref:https://docs.docker.com/engine/articles/dockerfile_best-practices/#add-or-copy

Thanks.


smacker:
Thank you @gy741 !

PR #338, opened 2018-11-05
adapted CD files to distributed mode

rporres:
Signed-off-by: Rafael Porres Molina <rafa@sourced.tech>

### DIFF
--- a/.helm-staging.yml
+++ b/.helm-staging.yml
@@ -13,8 +13,6 @@ databases:
 nodeSelector:
  lookout:
    srcd.host/app: lookout
- dummyAnalyzer:
-   srcd.host/app: lookout
 
 repositories:
   - url: github.com/src-d/lookout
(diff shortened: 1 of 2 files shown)


.helm-staging.yml
```diff
@@ -38,3 +36,11 @@ analyzers:
   - name: sonarcheck-analyzer
     addr: ipv4://lookout-sonarcheck-analyzer:10304
     feedback: https://github.com/src-d/lookout-sonarcheck-analyzer/issues/new
+
+rabbitmq-ha:
+  replicaCount: 1
+  rbac:
+    create: false
```
meyskens: Why is this set to false? The lookout-tiller service account should allow RBAC creation which is used by the chart to list service endpoints.

rporres: It doesn't work with RBAC. See https://github.com/src-d/infrastructure/issues/174#issuecomment-435835131

meyskens: Thanks for the info! It's not a problem :)

Issue #342, opened 2018-11-06
Add timeouts


smacker:
During the deployment to a new cluster, I witnessed that requests to analyzers just hang forever.

We need to add timeouts for that and maybe to some other places (like communication with GH api, downloading repositories) if we don't have timeouts there yet.

PR #343, opened 2018-11-07
Add timeouts

smacker:
Fix: #342 

Added (hardcoded) timeouts to:
- analyzers calls
- bblfsh calls
- git syncer
- github client

### DIFF
--- a/cmd/lookout-sdk/event.go
+++ b/cmd/lookout-sdk/event.go
@@ -114,7 +114,7 @@ func (c *EventCommand) makeDataServerHandler() (*lookout.DataServerHandler, erro
 			files:   dataService,
 		}
 	} else {
-		dataService = bblfsh.NewService(dataService, dataService, bblfshConn)
+		dataService = bblfsh.NewService(dataService, dataService, bblfshConn, 0)
 	}
 
 	dataService = purge.NewService(dataService, dataService)
--- a/cmd/lookout-sdk/push.go
+++ b/cmd/lookout-sdk/push.go
@@ -57,11 +57,15 @@ func (c *PushCommand) Execute(args []string) error {
 		return err
 	}
 
-	srv := server.NewServer(&server.LogPoster{log.DefaultLogger}, dataSrv.FileGetter, map[string]lookout.Analyzer{
-		"test-analyzes": lookout.Analyzer{
-			Client: client,
+	srv := server.NewServer(
+		&server.LogPoster{log.DefaultLogger}, dataSrv.FileGetter,
+		map[string]lookout.Analyzer{
+			"test-analyzes": lookout.Analyzer{
+				Client: client,
+			},
 		},
-	}, &store.NoopEventOperator{}, &store.NoopCommentOperator{})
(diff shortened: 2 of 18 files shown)


carlosms:
Code looks OK :+1:

I'm assuming the goal is to avoid getting hung forever in rare cases, and that we don't gain much from failing fast. In that case, maybe we should choose higher timeouts?

- github request 3s. Could be something like 30s
- analyzer 5m. In this case, if the analyzer is doing some ML training after a push event it may take longer. What about something like 20m?
- bblfsh parse 10s. Could be higher, like 5m.

Anyway, I'm thinking of higher timeouts for installations that may have high load, and the ones you proposed may be fine in most cases.
I think the timeouts should go in the config, as you proposed in a comment. Otherwise we limit the creation of new analyzers that for whatever reason may take longer than what we predict is reasonable.

smacker:
Yes. The goal is to avoid hanging.

One broken analyzer would block posting comments from all others without the timeout.
Or hanged call to github in event polling can stop pulling completely for the installation.

Also it can lead to resource exhaustion:
- number of goroutines will grow unlimited and never finish
- open connections to github/bblfsh

For analyzers, I think it's better to have different timeouts for review/push events.

I don't mind to increase default timeouts but prefer to have reasonable ones so user can get feedback in a reasonable time instead of just handing for 20m if something went wrong with analyzer or 5 minutes if bblfsh doesn't work (even very large file in the slowest js-driver takes 1 second to parse, x10 timeout is a good default imo). 

The case of installations with high load will be covered by configuration. For our production cluster, we can increase it if needed.

carlosms:
That makes sense. In the end it boils down to each installation details, so let's:
- Add timeouts to config
- Add new config fields to docs
- Increase a little bit github request and bblfsh parse timeouts if you agree, they feel low to me.

What do you think?

smacker:
Now timeouts are configurable.
Push&review have different values.
Documentation added.

Current default timeouts:
- review = 5m all our current analyzers fit in it without a problem, same as style-analyzer. It isn't too big to provide quick enough feedback.
- push = 30m value from the design doc
- bblfsh = 10s according to benchmarks large file with the slowest driver takes 1s to parse. The value is x10 which should be enough for giant files.
- github = 3s. Normal response takes a dozen milliseconds. But it's external network and external service, so it's x300.

I'm sure they are not perfect. So please provide some reasoning behind numbers you want to change and we will update them.

carlosms:
OK, my main concern is with the 3s github timeout.

Failing fast is not going to provide us any advantage, because if something is broken in github's side, the next calls are also going to fail so there is no rush.

But those 3s may prevent lookout from working if the network is slow today, or maybe github is suffering a problem that makes it slower to reply, or lookout is installed on premises in a place with not so great internet, etc. In all these cases we are transforming a situation where lookout works slower, to one where lookout does not work at all.

I just feel 3s are too close to common response times, with no apparent advantage to push it as low as possible.

smacker:
Increased github timeout to 30s and bblfsh timeout to 60s.

PR #345, opened 2018-11-08
Revert "Remove style analyzer while we deploy it in staging"

rporres:
Now that we've been able to deploy style analyzer, let's place this configuration back

Reverts src-d/lookout#340

### DIFF
--- a/.helm-staging.yml
+++ b/.helm-staging.yml
@@ -27,6 +27,9 @@ providers:
     installation_sync_interval: 5m
 
 analyzers:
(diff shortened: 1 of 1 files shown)


rporres:
D'oh, not again... Sure, let's wait. That appeared after I created this. I wanted to test it with some traffic from lookout but it appears there's (even) more work to do to just make it work.

Thanks for noticing, @smacker

rporres:
fixed, @smacker. Lookout endpoint was wrong

PR #347, opened 2018-11-13
Fix http cache with authorized requests

### LINKED ISSUE #124: Cache is bypassed when requests use authorization
The http cache we have in use is bypassed because the request header `Authorization` does not match the cached response header (empty for `Authorization`). It is working fine when we make unauthorized requests (leaving the github env vars unset).

More exactly, this call to `varyMatches` returns false because of the `Authorization` header: https://github.com/gregjones/httpcache/blob/9cad4c3443a7200dd6400aef47183728de563a38/httpcache.go#L160

carlosms:
Fix #124.

The problem was that we were chaining `RoundTrippers` like this:
`httpcache` -> `limit` -> `github Auth`.
So the cache was getting requests that did not have any auth headers added yet, but when the responses got back from github, they did have those headers.
When comparing requests with the stored responses the headers did not match, and the cache was never used.

With this PR the `RoundTrippers` are now chained like this: `limit` -> `github Auth` -> `httpcache`.

### DIFF
--- a/provider/github/client.go
+++ b/provider/github/client.go
@@ -14,7 +14,6 @@ import (
 	"github.com/src-d/lookout/util/ctxlog"
 
 	"github.com/google/go-github/github"
-	"github.com/gregjones/httpcache"
 	"gopkg.in/src-d/go-git.v4/plumbing/transport"
 	log "gopkg.in/src-d/go-log.v1"
 )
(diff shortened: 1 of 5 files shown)


carlosms:
:+1: I agree it would be better to have it in a common place. But everything I could think of was too complex for such a small thing and would make the code hard to follow.

PR #354, opened 2018-11-14
support pagination in repositories list

### LINKED ISSUE #346: Lookout doesn't fetch all repos
I am running Lookout on my own server (in distributed mode) with a GitHub app authorized on all my repositories. I happen to own 208... (not all maintained). What I noticed is that Lookout does not fetch all my repos only a subset of them, while i checked that the repo i want to use it on has the app authorized.

smacker:
fix #346


smacker:
rebased. CI is green.

PR #356, opened 2018-11-15
[WIP] Better review comments

### LINKED ISSUE #280: Use draft-review + submit api for posting on github
Currently posting and saving status to DB aren't atomic.
After a successful call to GH if DB update failed review will be duplicated.

Proposed solution:
If we create [review](https://developer.github.com/v3/pulls/reviews/#create-a-pull-request-review) without body attribute, it will be visible to an user only after the [submit call](https://developer.github.com/v3/pulls/reviews/#submit-a-pull-request-review).

It's safe to call submit multiple times.

Steps:

 - Call create without body if it doesn't have ID of review in DB already
 - Save returned ID to DB
 - In case of failure, we can just retry later without any issues for an end user
 - Send confirm (submit)
 - Update status in DB
 - In case of failure, we can just safely retry later

smacker:
Fix #280 but differently from the description in the issue.

Working on the original proposal I realized it would require to complicate interfaces a lot.
Poster must return remoteID, add new methods.
Storage interfaces become much more complicated due to the need for transactions across different storages.

Here is different proposal implementation:
1. Poster should implement "safe" mode if it's applicable
2. I introduced new status for events that reflects posting was started but not finished. In such case server uses "safe" mode.

There are few more fixes:
1. Add sleep when posting many comments or we hit the same error (was submitted too quickly)
2. Transport to fix (unify) non-standard errors from review endpoint to normal once. So when we get an error we can actually see the response from GH and log it. (currently such errors logged as empty structures)


carlosms:
Looks good to me.

On one hand seems a bit complex for an uncommon problem where the worst that can happen is having duplicated, but valid, comments. On the other hand, since this should only happen very rarely, I guess the added complexity does not affect the common code path and it's ok to handle this corner case in the proper way.

I'm not sure about merging comments. If they come from different analyzers, having them as independent comments allows to have 2 conversation threads, or resolve one of them as a bad suggestion but keep discussing the other one.

Maybe it would make sense to merge comments that come only from the same analyzer.

smacker:
> the added complexity does not affect the common code path

true. The only change for normal code path - writing one more status to DB.

> Maybe it would make sense to merge comments that come only from the same analyzer.

good point!

smacker:
ping @carlosms (rebased on master)

Issue #339, opened 2018-11-05
Extend the protocol to provide the ability to parse arbitrary code


vmarkovtsev:
We are developing the format analyzer. We hit the following problem: our analyzer may suggest code which is completely wrong, up to the point that it is not parsed to the same AST or not parseable at all. We don't want to suggest garbage, of course, so we need to parse the suggested variant and compare the old and the new ASTs. However, Lookout protocol does not allow to parse arbitrary code, only by specific Git revision. Our workaround is to invoke Bablefish directly, which breaks the whole abstraction.

I request an extension of the protocol which is basically a thin proxy layer between the Lookout server and the Babelfish server. We need the old, "v1" UASTs in Q4. More specifically: given a file name and it's contents, return the UAST. We are totally fine with batching.

smacker:
@smola this feature request is a little bit controversial, could you please approve it before we start implementing it?

smola:
Analyzers definitely need to access this API, whether it is on lookout itself or not.
I think we recently agreed that it would be exposed in the lookout API, @mcuadros is that right?

vmarkovtsev:
Decision on the meeting today:

- We will proceed with proxying bblfsh API from the lookout data server.
- We will consider later on if lookout should provide a higher level API to do sanity checks on suggested changes from analyzers.

carlosms:
A comment on the solution, from slack conversations:
Lookout will proxy the gRPC methods to bblfsh. The analyzers will be use a regular bblfsh client, which will be initialized with some new helper method in lookout-sdk.

carlosms:
Reopening since I have just realized we don't have any docs for the new feature.

smacker:
There is pull request waiting for release of new version:
https://github.com/src-d/lookout-sdk/pull/29

Or do you think we also need them in this repository?

carlosms:
Let me think about it, but probably that's enough taking into account we are in the process of migrating everything over to GitBook.

carlosms:
Releasing in a few minutes :)

PR #361, opened 2018-11-20
Proxy bblfsh

smacker:
Fix: #339


cmd/lookoutd/common.go
```diff
@@ -375,8 +379,10 @@ func (c *queueConsumerCommand) startServer(srv *lookout.DataServerHandler) error
 	go func() {
 		if err := grpcSrv.Serve(lis); err != nil {
 			log.Errorf(err, "data server failed")
+			os.Exit(1)
```
carlosms: Why is this needed now? Are there any changes that require it, or was it missing from before?

smacker: it was missed before I just noticed and fixed it while doing my task.

Currently, if data-server fails we log it but don't exit.
Worker will continue processing events but all of them will fail because analyzers won't be able to connect to data-server to get changes.

smacker:
all tests are passing successfully locally but fail in travis (with errors like `listen tcp 0.0.0.0:8090: bind: address already in use`). I'm investigating.

cmd/lookoutd/common.go
carlosms: How did you stumble with the problem? Can this happen only on initialization, like an already used port, or at any time?

smacker: I just noticed it while looking at the code around.
I didn't dig too deep in grpc implementation to answer your question. But for http server it can return error in 2 cases on initialization and when done channel is closed. (happens when you call `server.Shutdown()` or `server.Close()`) Most probably grpc works similar.

carlosms: What I was thinking is that maybe this method should be blocking, removing `go func` and directly do `return grpcSrv.Serve`. Then on top handle it like here https://github.com/src-d/lookout/blob/master/cmd/lookoutd/serve.go#L96. WDYT? is it worth to make it consistent with other top level errors, so we have one clear exit point?

smacker: We use this function not only in `serve.go` but in worker too. It doesn't use chan/switch pattern. SDK uses a similar pattern but instead of making the methods blocking it accepts chan as an argument. 

Though I like your idea to make it consistent, I'm not sure which pattern we should use. Do you mind if we move it to another issue then?

carlosms: ok, let's keep it as it is here, and I'll open an issue to look into it later, it's a minor thing and not really urgent.

carlosms: Here: https://github.com/src-d/lookout/issues/362

cmd/lookout-sdk/event.go
```diff
@@ -129,7 +129,12 @@ func (c *EventCommand) makeDataServerHandler() (*lookout.DataServerHandler, erro
 
 func (c *EventCommand) bindDataServer(srv *lookout.DataServerHandler, serveResult chan error) (*grpc.Server, error) {
 	log.Infof("starting a DataServer at %s", c.DataServer)
-	grpcSrv := grpchelper.NewServer()
+	bblfshGrpcAddr, err := grpchelper.ToGoGrpcAddress(c.Bblfshd)
```
dpordomingo: unhandled `err`

dpordomingo:
I couldn't post before merging because the connectivity issues in the office :(

Makefile
```diff
@@ -96,3 +108,16 @@ toc:
...
 	chmod a+x gh-md-toc
 	./gh-md-toc --insert README.md
+
+.PHONY: ci-start-bblfsh
+ifeq ($(OS),Darwin)
+ci-start-bblfsh:
+	@echo "running bblfsh is unsupported on mac os ci"
+else
+ci-start-bblfsh:
+	docker run -d --name bblfshd --privileged -v $(HOME)/bblfshd:/var/lib/bblfshd -p "9432:9432" bblfsh/bblfshd:v2.9.0
```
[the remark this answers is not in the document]
smacker: Please explain why.

I don't see any problem with it. Makefile has `ci-install` and `prepare-services` targets already. bblfsh is just one more service.

cmd/lookout-sdk/event.go
smacker: thanks! fixed: https://github.com/src-d/lookout/commit/650fcdd90a90f34713bcabc198ca091f04eef458

PR #367, opened 2018-11-26
[RFC][WIP] add review-all command

smola:
This command sends all commits in a repository to an analyzer.

This is just a proof of concept that I'm using for local development. It could use some improvements, such as some pretty git-log-style pretty printing with less redundancy.

### DIFF
--- a/cmd/lookout-sdk/review_all.go
+++ b/cmd/lookout-sdk/review_all.go
@@ -1,0 +1,128 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"time"
+
+	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-git.v4/plumbing/object"
+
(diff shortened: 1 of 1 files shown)


smacker:
I wonder what is the use case of it? Make sure analyzer doesn't fail on some commit or ...?

smola:
@smacker I'm testing an analyzer that, quite often, produces no output at all. So when I want to check results, I need to go through a lot of commits. Comments are sparse enough for me to need a few dozen commits analyzed to get more than a couple comments.

smacker:
I see. Yeah, it may make sense for some analyzers. Though I think we definitely need to change logs output for it.

carlosms:
Is it relevant for your tests that each commit is processed individually? I wonder if the need is mostly covered setting `--from` to an older enough commit.

smola:
@carlosms Calling `--from` with an old commit works too. Although including intermediate commits increases the number of results (many more variations in files).

PR #373, opened 2018-11-29
Don't add footer to empty comment. Fix #371

### LINKED ISSUE #371: Lookout posts empty comments
https://github.com/src-d/lookout/pull/370

screenshot 2018-11-28 at 16 08 14

Most probably degradation from recently merged #356

smacker:
Bug was introduced in #356
Review body should also contain footer if it exists.
But addFootnote function didn't have check for empty text which caused
reviews with footers only.


carlosms:
I love that lookout-staging agrees with the bug description :joy:

smacker:
Sorry David I don't wait for your review. The bug is pretty annoying and the fix is super small. Please comment here or create a new issue if you think something should be changed.

PR #375, opened 2018-12-03
Add log fields to queue jobs

### LINKED ISSUE #357: Log fields are lost when events are sent to the queue
The log fields added in https://github.com/src-d/lookout/pull/245 were lost when we introduced the queue. Once the job is dequeued, it does not keep the context with the log Fields.

carlosms:
Fix #357.

WIP because it depends on https://github.com/src-d/go-log/pull/9.

I checked that if an old queue job without the field in this PR is dequeued and deserialized, it works without a problem.
I changed the fields added by the github provider because `pr-number` does not really help to identify, and adds noise. I also added a `github.` prefix to help organize log fields a bit better, the same way we have `grpc.start_time`, `grpc.code`.

### DIFF
--- a/provider/github/watcher.go
+++ b/provider/github/watcher.go
@@ -250,8 +250,7 @@ func (w *Watcher) handlePrs(ctx context.Context,
 
 	for _, e := range prs {
 		ctx, _ := ctxlog.WithLogFields(ctx, log.Fields{
-			"pr-id":     e.GetID(),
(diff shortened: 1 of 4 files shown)


carlosms:
Yes, I created #427 to make the needed changes.

carlosms:
Pushed again using the changes from #439.

PR #378, opened 2018-12-03
Use go-cli lib

carlosms:
Migrate the CLI to use https://github.com/src-d/go-cli.

This PR does not solve any issue, but I think in the long term it's good to have all our projects use the same libraries to be consistent.


cmd/dummy/main.go
```diff
@@ -8,18 +8,24 @@ import (
...
 
 var (
-	version = "local_build_1"
-	app     = cli.New("dummy")
+	name    = "dummy"
+	version = "undefined"
+	build   = "undefined"
 )
 
+var app = gocli.New("dummy", version, build, "Dummy analyzer for testing")
```
smacker:
```
gocli.New(name, version, build, "Dummy analyzer for testing")
```

util/cli/log.go
```diff
@@ -3,24 +3,15 @@ package cli
...
-	LogFormat      string `long:"log-format" env:"LOG_FORMAT" description:"log format (text or json), defaults to text on a terminal and json otherwise"`
-	LogFields      string `long:"log-fields" env:"LOG_FIELDS" description:"default fields for the logger, specified in json"`
-	LogForceFormat bool   `long:"log-force-format" env:"LOG_FORCE_FORMAT" description:"ignore if it is running on a terminal or not"`
-}
-
-var _ initializer = &LogOptions{}
-
-// Init initializes the default logger factory.
-func (c *LogOptions) init(app *App) {
+// InitLog is similar to go-cli LogOptions.Init, but adds the application name
```
smacker: it also applies the configuration to logrus.

cmd/lookoutd/serve.go
```diff
@@ -7,23 +7,25 @@ import (
...
 type ServeCommand struct {
+	gocli.PlainCommand `name:"serve" short-description:"run a standalone server" long-description:"Run a standalone server"`
 	queueConsumerCommand
 }
 
 func (c *ServeCommand) Execute(args []string) error {
 	ctx, stopCtx := context.WithCancel(context.Background())
 	stopCh := make(chan error, 1)
 
+	cli.InitLog(&c.LogOptions, name)
```
smacker: What do you think if we use `cli.Command` instead of `cli.PlainCommand`? It would bring profiling by default now and any other perks in the future. Also I really like that `go-cli` calls init by default. Right now we don't use it.

My suggestion:
```
type LookoutCommand struct {
  cli.Command
}

func (c LookoutCommand) Init(a *App) error {
  c.Command.Init()
  // add additional logic for our logger
}
```

cmd/lookoutd/serve.go
```diff
@@ -7,23 +7,25 @@ import (
...
-	}
+	app.AddCommand(&ServeCommand{})
 }
 
 type ServeCommand struct {
+	gocli.PlainCommand `name:"serve" short-description:"run a standalone server" long-description:"Run a standalone server"`
 	queueConsumerCommand
 }
 
 func (c *ServeCommand) Execute(args []string) error {
```
smacker: Let's use [ExecuteContext](https://github.com/src-d/go-cli/blob/master/context.go#L15) and remove our custom signal handling.

cmd/dummy/main.go
carlosms: Done in 0c2b3ea

util/cli/log.go
carlosms: Done in f800074

cmd/lookoutd/serve.go
carlosms: I'd prefer to leave the full `cli.Command` out of this PR to limit it to migrate what we have to go-cli, as is. We can add profiling later easily if we think we need it.

For the initializer interface, I added it in 2812a43

cmd/lookoutd/serve.go
carlosms: Done in 5efef51

carlosms:
Ready for another round of review.
The merge conflicts are minimal, I'll do it before merging to not mess with the current reviews and comments.

cmd/lookoutd/serve.go
smacker: makes sense 👍

smacker:
Sorry most probably I did something wrong and github doesn't show context for my comments anymore...

carlosms:
Besides the changes requested I've pushed another commit c9dbaa9 to make our `LogOptions` behave like go-cli's `LogOption`.

I didn't do the same for `QueueOptions` and `DBOptions` because they connect to other components, and it feels like this does not belong to an "initialization" phase. It's more about the code structure, it would work the same in the command's `Init` and `Execute`.

cmd/lookoutd/common.go
```diff
@@ -50,12 +50,20 @@ type lookoutdCommand struct {
...
 }
 
 // Init implements the go-cli initializer interface. Initializes logs
 // configuration based on the cli options
 func (c *lookoutdCommand) Init(a *gocli.App) error {
 	cli.InitLog(&c.LogOptions, name)
+
+	var err error
+	c.conf, err = c.initConfig()
+	if err != nil {
```
smacker: nit pick:
```go
var err error
c.conf, err = c.initConfig()
return err
```

sorry 😅

carlosms: Done in 63ae46e

carlosms:
Tests are failing because of #391

PR #380, opened 2018-12-03
Remove timeouts in lookout-sdk

### LINKED ISSUE #376: Add timeout configuration for lookout-sdk
To the best of my knowledge, there is no way to change timeout parameters for `lookout-sdk` the same way it is possible for lookout server (https://github.com/src-d/lookout/blob/master/docs/configuration.md#advanced-fine-tuning)

ML team need to have a way to change it. Some big repositories take too much time to process but we need results for quality estimation. We should solve this issue before we can release `style-analyzer`

cc @vmarkovtsev

carlosms:
Fix #376.

This PR does:
- Moves the timeout defaults to `lookoutd` config initialization. This allows to:
  - Have no timeouts for `lookout-sdk`
  - See the values that apply in the initial config logs (in master the default timeouts are shown as `0` in the config logs)
- Increase the default timeouts. There is not much to gain from lower timeouts, and make the ML team analyzer fail.

The `lookout-sdk` is a tool that runs once, for manual testing, and you can cancel with a `ctrl+c`. It makes sense to remove any timeouts.

### DIFF
--- a/cmd/lookoutd/common.go
+++ b/cmd/lookoutd/common.go
@@ -109,6 +109,15 @@ func (c *lookoutdCommand) initConfig() (Config, error) {
 		return conf, fmt.Errorf("Can't open configuration file: %s", err)
 	}
 
+	// Set default timeouts
+	conf.Timeout = TimeoutConfig{
+		AnalyzerReview: 10 * time.Minute,
+		AnalyzerPush:   60 * time.Minute,
+		GithubRequest:  time.Minute,
+		GitFetch:       20 * time.Minute,
+		BblfshParse:    2 * time.Minute,
+	}
+
 	if err := yaml.Unmarshal([]byte(configData), &conf); err != nil {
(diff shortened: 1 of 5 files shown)


cmd/lookoutd/common.go
```diff
@@ -113,6 +113,23 @@ func (c *lookoutdCommand) initConfig() (Config, error) {
 		return conf, fmt.Errorf("Can't parse configuration file: %s", err)
 	}
 
+	// Set default timeouts
```
smacker: I would suggest to rewrite it like this:
```
conf.Timeout = TimeoutConfig{
    AnalyzerReview: 10 * time.Minute,
    AnalyzerPush: 60 * time.Minute,
    ...
}

if err := yaml.Unmarshal([]byte(configData), &conf); err != nil {
```

carlosms: :+1: done in 9b3c61f

PR #382, opened 2018-12-04
Fix timeout yml keys in conf file

### LINKED ISSUE #381: The documented timeout yml keys do not apply
The documented keys for the timeouts are not the ones we use in the code, so the timeout config does not apply.

carlosms:
Fix #381.

The keys were `analyzerreview` instead of the ones in the docs, `analyzer_review`. I also added the timeouts to the template conf file.

The default values are the ones from #380.


smacker:
But CI is failing

carlosms:
Yes, I submitted https://github.com/src-d/lookout/pull/383 to fix the build

Issue #350, opened 2018-11-13
Add default host and port on lookout-sdk review


smola:
I assume that the general case is that we're testing an analyzer locally and on the default port (`2022`?). So we could just have `ipv4://0.0.0.0:2022` as the default analyzer to check.

smacker:
currently there is no default port for analyzers:

dummy: [10302](https://github.com/src-d/lookout/blob/master/cmd/dummy/main.go#L22)
gometalint: [2001](https://github.com/src-d/lookout-gometalint-analyzer/blob/master/cmd/gometalint-analyzer/main.go#L34)
sonarcheck: [2022](https://github.com/src-d/lookout-sonarcheck-analyzer/blob/master/sonarcheck_analyzer.py)
terraform: [2020](https://github.com/src-d/lookout-terraform-analyzer/blob/master/main.go#L12)
style: [2000](https://github.com/src-d/style-analyzer#how-to-use)

shall we choose one and update all?

smola:
Yes, I think so, it would be in line with other efforts we are making to choose good default ports. If we have to make an arbitrary choice, I propose 9930 (random 4-digit port number without an entry [known to nmap](https://svn.nmap.org/nmap/nmap-services)).

smacker:
Current status:

- [x] lookout-tool
- [x] dummy
- [x] gometalint
- [x] sonarcheck
- [ ] terraform (PR created)
- [ ] style (doesn't have default number, requires explicit number in configuration, PR with documentation update to 9930 as recommended created)

smacker:
I think we can close this issue.

- Default port support in lookout-tool was implemented
- All analyzers with default port were updated
- Style-analyzer doesn't have default port but documentation update was suggested. (style-analyzer doesn't belong to us so it's up to ML team to merge update or not)

PR #389, opened 2018-12-04
Default port for analyzer

smacker:
Part of https://github.com/src-d/lookout/issues/350

### DIFF
--- a/cmd/lookout-sdk/event.go
+++ b/cmd/lookout-sdk/event.go
@@ -20,7 +20,7 @@ import (
 
 	gogit "gopkg.in/src-d/go-git.v4"
 	"gopkg.in/src-d/go-git.v4/plumbing"
-	"gopkg.in/src-d/go-log.v1"
+	log "gopkg.in/src-d/go-log.v1"
 	"gopkg.in/src-d/lookout-sdk.v0/pb"
 )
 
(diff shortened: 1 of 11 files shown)


smacker:
The pr needs to be merged & released before updating analyzers (because they use lookout-sdk in CI)

cmd/lookout-sdk/event.go
```diff
@@ -32,9 +32,7 @@ type EventCommand struct {
 	RevFrom    string `long:"from" default:"HEAD^" description:"name of the base revision for event"`
 	RevTo      string `long:"to" default:"HEAD" description:"name of the head revision for event"`
 	ConfigJSON string `long:"config-json" description:"arbitrary JSON configuration for request to an analyzer"`
-	Args       struct {
-		Analyzer string `positional-arg-name:"analyzer" description:"gRPC URL of the analyzer to use"`
-	} `positional-args:"yes" required:"yes"`
+	Analyzer   string `long:"analyzer" default:"ipv4://localhost:9930" description:"gRPC URL of the analyzer to use"`
```
carlosms: With this change the new release will break the usage from/to
- `lookout-sdk review ipv4://localhost:10302`
- `lookout-sdk review --analyzer ipv4://localhost:10302`

right?
Is there a need for this, or could we just make the argument optional to keep compatibility?

smacker: The reason why it's changed is in the commit message: https://github.com/src-d/lookout/pull/389/commits/3275303db234dd6e4e314c2787683a3bdec48d60

I would prefer to use `--analyze` as a recommended way to change address. But return positional argument for compatibility. WDYT?

carlosms: I see your point. Even if it makes the code ugly, I think we should prioritize compatibility and keep the argument.
And if we keep the argument having also `--analyzer` to do the same thing will lead to unnecessary confusion.

I would just go with the optional argument, checking in our code if it's missing to use the default, and adding to the argument description `  (default: ipv4://localhost:9930)`

smacker:
@carlosms updated. I squashed all commits together and pushed force.

carlosms:
The tests are failing, but I don't think it's related to the changes in this PR (please double check in case I'm wrong).
I created an issue to track it, https://github.com/src-d/lookout/issues/391

smacker:
True. This test became unstable (only in travis on linux though, I can't reproduce locally on mac or I have never seen it failing on travis macos). Restart usually helps.

smacker:
Yep. Green now. Merging.

Issue #365, opened 2018-11-22
repository not found


smacker:
Maybe private repositories cloning is broken?

<img width="840" alt="screenshot 2018-11-22 at 20 45 25" src="https://user-images.githubusercontent.com/406916/48920168-a0337780-ee97-11e8-9545-6d8f1b71f721.png">

smacker:
After debugging I couldn't find any problem in lookout.
At the end, I generated token using staging private key. And:
```
$ git clone https://x-access-token:[TOKEN]@github.com/src-d/infrastructure.git
Cloning into 'infrastructure'...
remote: Repository not found.
```

Most probably the app lacks `content` access to the repository. @src-d/infrastructure please confirm.

rporres:
These are the permissions that https://github.com/apps/lookout-staging has over src-d code

![image](https://user-images.githubusercontent.com/2647973/49278108-16ca0980-f484-11e8-85d4-5e4296a5cbe3.png)

This is what you mean when you say the app lacks `content` access, @smacker?

meyskens:
@smacker this were the permissions I applied: https://github.com/src-d/issues-infrastructure/issues/233

smacker:
Yes. We introduced support for private repositories after that ticket was created. 
Lookout needs access to read content of a repository to be able to clone it. Please add it.

rporres:
We need @mcuadros superpowers for this

carlosms:
@smacker please check if we also need to update our docs

smacker:
As proposed by David/Carlos moved to infra repo: https://github.com/src-d/infrastructure/issues/249
I'm closing this one.

PR #390, opened 2018-12-05
Update Github App required permissions

smacker:
Ref: https://github.com/src-d/lookout/issues/365


carlosms:
Note for other reviewers: [Repository contents](https://developer.github.com/v3/apps/permissions/#permission-on-contents) allows the same action as [Single File](https://developer.github.com/v3/apps/permissions/#permission-on-single-file), that's why it can be replaced instead of added.

smacker:
tbh I don't know why we asked for single file at all. I don't see any code that needs it.

PR #392, opened 2018-12-05
Only languages filter scanner

smacker:
~Depends on https://github.com/src-d/lookout-sdk/pull/44~
Depends on a new `lookout-sdk` release containing [lookout-sdk#44](https://github.com/src-d/lookout-sdk/pull/44)

Fix: https://github.com/src-d/lookout-sdk/issues/36

Review commit by commit would be easier than all the changes together.


service/enry/scanner.go
```diff
@@ -0,0 +1,124 @@
...
+func filterVendor(f *lookout.File) (bool, error) {
+	if f == nil {
+		return false, nil
+	}
+
+	return enry.IsVendor(f.Path), nil
+}
+
+// newChangeExcludeVendorScanner creates new FnChangeScanner
+func newChangeExcludeVendorScanner(scanner lookout.ChangeScanner) *lookout.FnChangeScanner {
```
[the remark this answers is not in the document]
dpordomingo: it makes sense, thanks for the clarification !!

dpordomingo:
The dependency on src-d/lookout-sdk#44 was already satisfied.
Is there something missing?

dpordomingo:
Waiting for the new release of `lookout-sdk`

service/enry/scanner_test.go
```diff
@@ -0,0 +1,176 @@
...
+			Head: &lookout.File{Path: "f1new.go"},
+		},
+		{
+			Base: &lookout.File{Path: "f1old.py"},
+		},
+		{
+			Base: &lookout.File{Path: "f2old.py"},
+			Head: &lookout.File{Path: "f2new.js"},
+		},
+		// the change that used to be vendor but isn't anymore should survive
```
carlosms: Is this comment correct? The test case seems to be the exact opposite

smacker: thanks! The comment is correct but the code is wrong, the right one was at the very end of use cases. Fixed.

Issue #391, opened 2018-12-05
Integration tests failing on Travis


carlosms:
Since a couple of days ago I've see the "Lookoutd Integration Tests Linux" fail on Travis. They usually work if you retry it manually.

One of the failing logs:
```

$ LOOKOUT_TEST_QUEUE=true make test-json
rm -rf build build/bin 
go clean .
go build -ldflags "-X main.version=dev-362f0e3 -X main.build=12-05-2018_11_25_07 -X main.commit=362f0e3" -o "build/bin/dummy" ./cmd/dummy
go build -ldflags "-X main.version=dev-362f0e3 -X main.build=12-05-2018_11_25_09 -X main.commit=362f0e3" -o "build/bin/lookoutd" ./cmd/lookoutd
go build -ldflags "-X main.version=dev-362f0e3 -X main.build=12-05-2018_11_25_12 -X main.commit=362f0e3" -o "build/bin/lookout-sdk" ./cmd/lookout-sdk
DUMMY_BIN=/home/travis/gopath/src/github.com/src-d/lookout/build/bin/dummy \
	LOOKOUT_BIN=/home/travis/gopath/src/github.com/src-d/lookout/build/bin/lookoutd \
	go test -v -parallel 1 -tags='integration bblfsh' github.com/src-d/lookout/cmd/server-test
=== RUN   TestDummyIntegrationSuite
=== RUN   TestDummyIntegrationSuite/TestReviewDontPostSameComment
running lookoutd migrate
starting lookoutd watch --provider json -c ../../fixtures/dummy_config.yml
starting lookoutd work --provider json -c ../../fixtures/dummy_config.yml --probes-addr 0.0.0.0:8091
starting dummy serve --files
stopping services
=== RUN   TestDummyIntegrationSuite/TestSuccessPush
running lookoutd migrate
starting lookoutd watch --provider json -c ../../fixtures/dummy_config.yml
starting lookoutd work --provider json -c ../../fixtures/dummy_config.yml --probes-addr 0.0.0.0:8091
starting dummy serve --files
stopping services
=== RUN   TestDummyIntegrationSuite/TestSuccessReview
running lookoutd migrate
starting lookoutd watch --provider json -c ../../fixtures/dummy_config.yml
starting lookoutd work --provider json -c ../../fixtures/dummy_config.yml --probes-addr 0.0.0.0:8091
starting dummy serve --files
stopping services
=== RUN   TestDummyIntegrationSuite/TestWrongRevision
[...]

PR #395, opened 2018-12-07
Skip macOS tests on PR

carlosms:
We are already skipping "SDK Integration Tests macOS" on PR to speed up build times. With this PR we also skip "Lookoutd Integration Tests macOS".

A [recent build](https://travis-ci.org/src-d/lookout/builds/464865586) took 15min for macOS tests, when the linux ones were done in 3min.


carlosms:
Tests are failing because of #391.

Issue #385, opened 2018-12-04
If a repo added in config.yml does not exist, CLI is flooded with 404 requests


erizocosmico:
I added a repo that didn't exist to the config.yml and started lookout. It started quickly flooding the CLI with 404 requests.

PR #408, opened 2018-12-10
Set a fixtures repo in config-template

dpordomingo:
superseded by https://github.com/src-d/lookout/pull/410

When a newcomer creates the `config.yml` with a copy/paste from `config.yml.tpl` and then runs `lookout`, it will spam **lookout** main repo.

If this PR is merged, the watched repo will be `github.com/src-d/lookout-test-fixtures` instead of the main one.

### alternative
Set a placeholder, with a missing repo, so the user will be forced to change it.
- **cons:**
  - it will be needed an extra step to try **lookout**,
  - The user might not have a testing repo in case he don't want to spam their own repos for trying lookout.
- **pros:**
  - the user is forced to think before starting running commands :wink: 

Adopting the alternative without solving #385 (**_If a repo added in config.yml does not exist, CLI is flooded with 404 requests_**) could lead into a bad UX.

### DIFF
--- a/config.yml.tpl
+++ b/config.yml.tpl
@@ -13,7 +13,7 @@ providers:
     # watch_min_interval: 2s
 
 repositories:
-  - url: github.com/src-d/lookout
+  - url: github.com/src-d/lookout-test-fixtures
     client:
       # user:
       # token:


smacker:
I vote for placeholder or for a repo specially designed for showing lookout features with some standard analyzer. This repo exists to reproduce bugs...

carlosms:
I don't like this change to be honest. We are just moving the problem from one repository to another.

For now, I'd prefer to have a placeholder you need to edit to point to the repository where you want to test it.

dpordomingo:
I prefer alternatives that require as less config as possible to let the newcomers try it, but ok.

Superseded by https://github.com/src-d/lookout/pull/410

Issue #409, opened 2018-12-10
Change lookout-sdk return code if analysis failed


zurk:
Right now there is no easy way to check if Analyser fails with an error or it succeeds to process the request. It will be really helpful to add a non-zero return code for such cases.

PR #411, opened 2018-12-11
feature: make lookout-sdk return error code if analysis failed

smacker:
I have rewritten concurrentRequest function using channels to be able to
exit as soon as the first analyzer failed.

commentList is also replaced with channel because it's weird to mix
sync package and channels in one function.

fix: https://github.com/src-d/lookout/issues/409

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -270,6 +270,10 @@ func (s *Server) getConfig(ctx context.Context, e lookout.Event) (map[string]loo
 }
 
 func (s *Server) concurrentRequest(ctx context.Context, conf map[string]lookout.AnalyzerConfig, send reqSent) ([]lookout.AnalyzerComments, error) {
+	var cancel context.CancelFunc
+	ctx, cancel = context.WithCancel(ctx)
+	defer cancel()
+
 	commentsCh := make(chan *lookout.AnalyzerComments, len(s.analyzers))
 	errCh := make(chan error)
 
--- a/server/server.go
+++ b/server/server.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"reflect"
-	"sync"
 	"time"
 
 	"github.com/src-d/lookout"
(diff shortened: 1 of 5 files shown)


server/server.go
```diff
@@ -279,19 +293,37 @@ func (s *Server) concurrentRequest(ctx context.Context, conf map[string]lookout.
...
 		}(name, a)
 	}
-	wg.Wait()
 
-	return comments.Get()
+	var comments []lookout.AnalyzerComments
+	for i := 0; i < len(s.analyzers); i++ {
+		select {
+		case err := <-errCh:
+			return nil, err
```
carlosms: Should we do a context cancellation for all the other `send` still running in goroutines?

smacker: I thought about it and most probably we don't need to cancel it here.
We will propagate the error to the cmd which would catch it and cancel everything.

server/server.go
```diff
@@ -51,7 +54,7 @@ func NewServer(
 	reviewTimeout time.Duration,
 	pushTimeout time.Duration,
 ) *Server {
-	return &Server{p, fileGetter, analyzers, eventOp, commentOp, reviewTimeout, pushTimeout}
+	return &Server{false, p, fileGetter, analyzers, eventOp, commentOp, reviewTimeout, pushTimeout}
```
dpordomingo: I'm late in the discussion... but... I'd prefer not to use positional arguments.

server/server.go
```diff
@@ -279,19 +297,38 @@ func (s *Server) concurrentRequest(ctx context.Context, conf map[string]lookout.
...
 		}(name, a)
 	}
-	wg.Wait()
 
-	return comments.Get()
+	var comments []lookout.AnalyzerComments
+	for i := 0; i < len(s.analyzers); i++ {
+		select {
+		case err := <-errCh:
+			cancel()
```
carlosms: Is this needed if we already do the `defer` above?

smacker: somehow in my head cancellation was meant to be before return from this function. But there is no reason for that. Thanks. Removed.

server/server.go
smacker: sorry David but it's unrelated to this PR. It uses the same style that was used before. We can fix it in another PR later if you want.

smacker:
I update the failing test also.

PR #415, opened 2018-12-13
New lookout-sdk --config option

### LINKED ISSUE #269: lookout-sdk binary doesn't pick up on config.yml

carlosms:
Fix #269.

With this change you can use `--config=./conf.yml`. To avoid confusion on what option has preference over another, what I've done is make it incompatible with `--config-json` and the argument `analyzer`. So if you decide to use a config file, you define everything there: the analyzer address and its settings.

I didn't make any changes to the docs since we are in the middle of the migration to GitBook. If this one gets approved we can open a new issue to add the docs later.

### DIFF
--- a/cmd/lookout-sdk/event.go
+++ b/cmd/lookout-sdk/event.go
@@ -5,10 +5,10 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"path/filepath"
 	"time"
 
-	"github.com/gogo/protobuf/types"
 	"github.com/src-d/lookout"
 	"github.com/src-d/lookout/service/bblfsh"
 	"github.com/src-d/lookout/service/enry"
@@ -16,12 +16,13 @@ import (
 	"github.com/src-d/lookout/service/purge"
 	"github.com/src-d/lookout/util/cli"
 	"github.com/src-d/lookout/util/grpchelper"
-	"google.golang.org/grpc"
 
+	"google.golang.org/grpc"
 	gogit "gopkg.in/src-d/go-git.v4"
 	"gopkg.in/src-d/go-git.v4/plumbing"
 	log "gopkg.in/src-d/go-log.v1"
 	"gopkg.in/src-d/lookout-sdk.v0/pb"
+	yaml "gopkg.in/yaml.v2"
 )
 
 type EventCommand struct {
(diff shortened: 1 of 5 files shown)


carlosms:
quoting @smacker 
> I have 2 analyzers in my config. I would expect both of them to be called. But it looks like the second one will be just ignored without any notification about it.

I tried to replicate exactly what the `argument` and `--config-json` do now. If you define more than one analyzer, you will get an error message, [see here](https://github.com/src-d/lookout/pull/415/files#diff-e48fe7868eacfe2f466d798dd862fe57R268).

Maybe in the future we could allow more than one analyzer if we have a use case that requires it. But for now I think it makes sense to keep `lookout-sdk` as simple as possible.

> It's also confusing for me that feedback field in analyzer and timeouts are also ignored.

This can be clarified in the documentation, we can say `lookout-sdk` takes the `analyzers` part of the config file.
To me it makes sense that anything that does not have to do with the `analyzers` is ignored. `feedback` is ignored because it is used to post comments on github. `timeout` applies to `lookoutd` but not to `lookout-sdk`. The same way github credentials are ignored. 

quoting @dpordomingo 
> I have my own concerns about offering kind of similar features from different access points and naming. I'd vote for a global solution.

This is a valid point. We already have `--config-json`, but I was assuming defining json in a file would be more convenient. I we were doing it from scratch I would propose to have only `lookout-sdk argument` or `lookout-sdk --config`. I could agree to drop this PR if we think it complicates things. Maybe we could ask ML team.

smacker:
TBH if I would be a user of `lookout-sdk` I would be very confused by this parameter.

For me current `--config-json` looks much more straightforward because it works exactly like `-d/--data` parameter of curl and the whole point of this binary to provide simple way to call analyzer as you do with curl and http.

I'm *not* strongly against support of lookoutd config but at least it would need proper documentation (and I would expect to clear most of my concerns raised above when I do `lookout-tool --help` not hidden somewhere in gitbook). Also, let's discuss it with ML team first if it's useful for them.

dpordomingo:
Let me provide two alternatives to be considered.
I'll back to you along before EOD

dpordomingo:
Since there is no special necessity on passing config in any other way than though `--config-json` we'll close this as agreed by [minutes/419](https://github.com/src-d/minutes/pull/419) at [R87](https://github.com/src-d/minutes/pull/419/files#diff-82137befbf83aa65593b6378ccb848cfR87)

Issue #397, opened 2018-12-07
lookoutd web subcommand skeleton


carlosms:
Umbrella issue to create the base of a new `lookoutd web` sub command. It includes the following tasks:
- [x] Initialize a create-react-app
  - Including support for [TypeScript](https://facebook.github.io/create-react-app/docs/adding-typescript)
  - Sample app or empty app is fine
  - Frontend code in a new folder, similar to [gitbase-web](https://github.com/src-d/gitbase-web/tree/master/frontend)
  - Without bootstrap yet. We need to define the library options and which one we want to use.
- [x] New `web` sub command for `lookoutd` that will serve the frontend
  - Totally individual, not integrated in `serve`
  - Code is packaged with bindata. We can use the same mechanism with build tags used in gitbase-web
  - Build with the `Makefile`. If for some reason we don't want to add the bindata and web server to the main build we could have both `make build` and a new `make build-web`.

PR #398, opened 2018-12-07
Enhance docs with GitBook

dpordomingo:
fix #330
blocks #423 and #424

The GitBook docs site is available under my own fork:
https://dpordomingo.gitbook.io/lookout

Once it is approved, it would be made public under:
https://docs.sourced.tech/lookout

I did:
- created the GitBook structure and complete the docs according to the team agreements,
- added the `.github/CODE_OF_CONDUCT.md` file,
- added the `DCO` file,
- fixed the `LICENSE.md` file, and created the `NOTICE.md` file,
- moved the `CONTRIBUTING.md` file into `docs` and enhance it,
- created a `docker-compose-rabbitmq.yml` so RabbitMQ dependency will be runnable as the others,
- move the the sequence diagrams into `docs/assets/` and enhance them,
- avoid downloading the toc generator if already available,

PR #416, opened 2018-12-17
Init frontend

smacker:
fix: #397

Differences with pure create-react-app:
- added typescript deps and js files renamed to tsx as recommended by create-react-app
- removed logo to avoid license
- removed readme
- removed service-worker
- added `yarn lint` command which runs eslint

web command:
- similar to gitbase-web but everything in 1 package web (at least for now) because in lookout web is just a small part of the project, not the main one
- it supports footer html just because of copy-past from gitbase

Makefile:
- multiple new commands with `web` prefix (similar to gitbase-web)
- default `dependencies`, `test` and `build` command are overridden to support frontend
- we don't have lint stage in CI or lint targets for lookout. We may consider add it later and run `web-lint` there too.

Documentation:
- currently none because it is going to be reworked almost completely by #398 
- we will add it later
- the process is similar to our other projects: `yarn start` for dev & `make web-serve` for testing with bindata.


.gitignore
```diff
@@ -16,6 +16,7 @@
 
 # build results
 /build
+/web/assets/asset.go
```
dpordomingo: aren't we pushing the assets?

.travis.yml
```diff
@@ -18,6 +18,10 @@ go:
 before_install:
   - sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-6 90
   - sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-6 90
+  - . $HOME/.nvm/nvm.sh
```
dpordomingo: ouch... we're making the CI slower because of the web :(
I didn't realize about of this during the design meeting :man_facepalming:

frontend/.gitignore
```diff
@@ -0,0 +1,23 @@
...
+# misc
+.DS_Store
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
+
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
```
dpordomingo: gitbase-web added `.eslintcache`

```suggestion
yarn-error.log*

.eslintcache
```

frontend/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "react-dom": "^16.6.3",
+    "react-scripts": "2.1.1",
+    "typescript": "^3.2.2"
+  },
+  "scripts": {
+    "start": "react-scripts start",
+    "build": "react-scripts build",
+    "test": "react-scripts test",
+    "eject": "react-scripts eject",
+    "lint": "eslint 'src/**/*.tsx'"
```
dpordomingo: We have a `format` script in gitbase-web
```suggestion
    "lint": "eslint 'src/**/*.tsx'",
    "format": "eslint --fix 'src/**/*.js'"
```

.travis.yml
```diff
@@ -18,6 +18,10 @@ go:
...
+  - nvm install 8
+  - nvm use 8
+  - npm install -g yarn
```
dpordomingo: Aren't we having a `validate and build` stage as in `gitbase-web`?

example:
```yaml
jobs:
  include:
    - stage: validate and build
      script:
        - make test
        - make lint
        - make coverage
        - make validate-commit
```

frontend/public/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+  "short_name": "React App",
```
dpordomingo: we should use project data instead of example placeholders :wink:

frontend/public/manifest.json
```diff
@@ -0,0 +1,15 @@
...
+  "name": "Create React App Sample",
+  "icons": [
+    {
+      "src": "favicon.ico",
```
dpordomingo: I think we should use the company favicon instead of the react one

web/static_files.go
```diff
@@ -0,0 +1,99 @@
+// +build bindata
```
dpordomingo: I'd put the handlers inside `handler` package as in our other current web apps.

.gitignore
smacker: no we don't: https://github.com/src-d/gitbase-web/blob/master/.gitignore#L9

.travis.yml
smacker: 🤷‍♀️

frontend/.gitignore
smacker: not gitbase-web. Old version of `create-react-app`. New version doesn't do it. Most probably it's not needed anymore with newer eslint.

cmd/lookoutd/web.go
```diff
@@ -0,0 +1,34 @@
...
+
+	cli.LogOptions
+}
+
+func (c *WebCommand) Execute(args []string) error {
+	static := web.NewStatic("build/public", c.ServerURL, c.FooterHTML)
+	server := web.NewHTTPServer(static)
+	addr := fmt.Sprintf("%s:%d", c.Host, c.Port)
+
+	log.Infof("Start http server on %s", addr)
```
carlosms: For consistency with other messages, I would use
```suggestion
 	log.Infof("Starting http server on %s", addr)
```

frontend/package.json
smacker: I decided to leave it for now because of typescript. I'm studing documentation for the language itself right now, then I'll check ecosystem. I'm not sure we need eslint with typescript at all.

.travis.yml
smacker: I mentioned it in the PR description:
> we don't have lint stage in CI or lint targets for lookout. We may consider add it later and run web-lint there too.

frontend/public/manifest.json
smacker: where can I get one?

.travis.yml
carlosms: We could include `yarn lint` in the current 'Generated code' travis job. And maybe rename it to 'Validate'

web/static_files.go
smacker: `handler` is a quite common name and in other apps that definitely makes sense because all the app is http server. In lookout http is just a very small part.

Let's start with only 1 package for now and split it later if we need. WDYT?

.travis.yml
smacker: my point is we don't have `make lint` target for go code.

If you try to run `golint` on our source code it would produce LOTS of errors.
Also do we still want to run vet? In newer go I believe it works by default.

Or do you think we should add linting only for frontend code?

frontend/public/manifest.json
smacker: thanks! fixed!

smacker:
Imo 200 lines Makefile most of which are simple isn't a big problem.

We can create `Makefile.web` and include it in main `Makefile` or we can run `make -f Makefile.web build` inside main Makefile. But that would still require override at least `build` target. I don't mind to move it but not sure it worths it. @dpordomingo wdyt?

.travis.yml
carlosms: If we enforce it from the beginning for the frontend code, it will be easier to keep it nice and formatted.
To avoid the problem you just mentioned with go, if adding a linter later involves too many changes we won't do it.

Up to you to include it here, or later in another PR after researching more about the better way to do it for typescript. But I would try to include it soon.

.gitignore
dpordomingo: wow... I had a reaaaaaly old version of `gitbase-web` in my PC, xD
so ashamed... :man_facepalming:

frontend/public/manifest.json
dpordomingo: whatever is better than the react one; for example:
- from Landing, https://sourced.tech/img/logos/favicon.png, or
- from Gitbase-web, https://gitbase-playground-staging.srcd.run/favicon.ico

.travis.yml
dpordomingo: I'd agree if there is an issue about this, and it is postponed to be done as the next task.

web/static_files.go
dpordomingo: I always prefer standard ways to face similar problems and to lower the differences between similar apps as much as possible; Anyway, it is not a strong opinion, just a personal preference that can be ignored if it does not suit you :wink:

dpordomingo:
You know @smacker that I use to vote for splitting things :wink: ; I'd do it for `Makefile.web` too.
But I'd not add an extra step for the building: it should be done with only one `make build` (it use to be a company requirement).

I'd purpose including `Makefile.web` as it's done with `Makefile.main`, and moving the new web things into that new `Makefile.web`.

frontend/public/manifest.json
smacker: I believe we should use a new purple one like this one: https://avatars1.githubusercontent.com/u/15128793?s=32&v=4
Let me try to ask for it from the design team...

dpordomingo:
Reviewing the current state of this PR, I'd say yes if:
- [ ] `lint` issue is created or added in this PR: https://github.com/src-d/lookout/pull/416#discussion_r242518574
- [ ] proper favicon is used: https://github.com/src-d/lookout/pull/416#discussion_r242518497

The `Makefile.web` https://github.com/src-d/lookout/pull/416#issuecomment-448208476 and `handler` package https://github.com/src-d/lookout/pull/416#discussion_r242519898 are not strong requirements but friendly suggestions :smile:

web/static_files.go
smacker: Okay. Let's keep it for now like this if you don't feel too strongly about it. And we can change it later if needed.

Just to explain my concern better:

The problem with "the standard" way is the context.
For example in gitbase-web top level `server` package means http server. But in lookout it doesn't. Same here, when I read lookout source code `web.Static` is obvious but `handler.Static` not really. Is it http handler? Or is it github event handler?

smacker:
I have removed target for web-lint completely for now.
Typescript support in eslint is experimental: https://eslint.org/docs/user-guide/configuring
And there is another project for linting: https://github.com/palantir/tslint

I have created another issue for that: https://github.com/src-d/lookout/issues/418

frontend/public/manifest.json
dpordomingo: if it's 32x32 I'd just use it here, and change whenever they release it if it has not been yet released.

smacker:
I moved targets to separate file and included it.

smacker:
Thank you guys for the thorough review!

PR #419, opened 2018-12-18
Avoid repeated API error messages

### LINKED ISSUE #385: If a repo added in config.yml does not exist, CLI is flooded with 404 requests

carlosms:
Fix #385.

With the changes in this PR the GitHub api errors are logged on error level every 5 minutes. The rest of the time they are logged on debug level.

I also changed the error logged to include only the error message, the stack trace was huge and it was not useful.

### DIFF
--- a/provider/github/watcher.go
+++ b/provider/github/watcher.go
@@ -40,14 +40,17 @@ var (
 type Watcher struct {
 	pool *ClientPool
 	// maps clients to functions that stop watching the client
-	stopFuncs map[*Client]func()
+	stopFuncs               map[*Client]func()
+	lastErrPR, lastErrEvent map[*lookout.RepositoryInfo]*errThrottlerState
 }
 
 // NewWatcher returns a new
 func NewWatcher(pool *ClientPool) (*Watcher, error) {
 	return &Watcher{
-		pool:      pool,
-		stopFuncs: make(map[*Client]func()),
+		pool:         pool,
+		stopFuncs:    make(map[*Client]func()),
+		lastErrPR:    make(map[*lookout.RepositoryInfo]*errThrottlerState),
+		lastErrEvent: make(map[*lookout.RepositoryInfo]*errThrottlerState),
 	}, nil
 }
 
@@ -177,9 +180,18 @@ func (w *Watcher) processRepoPRs(
 ) (time.Duration, error) {
 	resp, prs, err := w.doPRListRequest(ctx, c, repo.Owner, repo.Name)
 	if ErrGitHubAPI.Is(err) {
-		ctxlog.Get(ctx).With(log.Fields{
+		if w.lastErrPR[repo] == nil {
+			w.lastErrPR[repo] = &errThrottlerState{}
+		}
+
+		// go-errors %+v prints the stack trace. Doing this we create a plain
+		// error with no stack trace for the log
+		err := fmt.Errorf("%s", err)
+
+		newErrThrottlerLogger(ctxlog.Get(ctx), w.lastErrPR[repo]).With(log.Fields{
 			"repository": repo.FullName, "response": resp,
 		}).Errorf(err, "request for PR list failed")
+
 		return c.watchMinInterval, nil
 	}
 
@@ -199,9 +211,18 @@ func (w *Watcher) processRepoEvents(
 ) (time.Duration, error) {
 	resp, events, err := w.doEventRequest(ctx, c, repo.Owner, repo.Name)
 	if ErrGitHubAPI.Is(err) {
-		ctxlog.Get(ctx).With(log.Fields{
+		if w.lastErrEvent[repo] == nil {
+			w.lastErrEvent[repo] = &errThrottlerState{}
+		}
+
+		// go-errors %+v prints the stack trace. Doing this we create a plain
+		// error with no stack trace for the log
+		err := fmt.Errorf("%s", err)
+
+		newErrThrottlerLogger(ctxlog.Get(ctx), w.lastErrEvent[repo]).With(log.Fields{
 			"repository": repo.FullName, "response": resp,
 		}).Errorf(err, "request for events list failed")
+
 		return c.PollInterval(eventsCategory), nil
 	}
 
--- a/provider/github/err_throttler.go
+++ b/provider/github/err_throttler.go
@@ -1,0 +1,48 @@
+package github
+
+import (
+	"time"
+
+	log "gopkg.in/src-d/go-log.v1"
+)
+
+// errThrottlerInterval controls how often repeated ErrGitHubAPI (like 404)
+// are logged with Error level. The rest of the logs will be logged as Debug
+const errThrottlerInterval = 5 * time.Minute
+
+type errThrottlerState struct {
+	lastWhen time.Time
+	lastErr  string
+}
+
+// errThrottlerLogger is a logger that logs repeated error messages as Debug
+// level. An Error level is used once every errThrottlerInterval
+type errThrottlerLogger struct {
+	log.Logger
+	*errThrottlerState
+}
+
+func newErrThrottlerLogger(log log.Logger, state *errThrottlerState) *errThrottlerLogger {
+	return &errThrottlerLogger{log, state}
+}
+
+func (l *errThrottlerLogger) With(f log.Fields) log.Logger {
+	return &errThrottlerLogger{l.Logger.With(f), l.errThrottlerState}
+}
+
+func (l *errThrottlerLogger) Errorf(err error, format string, args ...interface{}) {
+	// log on Error level if the last message was different, or if the last time
+	// was longer than errThrottlerInterval ago
+	asErrLevel := l.lastErr != err.Error() ||
+		time.Now().After(l.lastWhen.Add(errThrottlerInterval))
+
+	if asErrLevel {
+		l.lastErr = err.Error()
+		l.lastWhen = time.Now()
+
+		l.Logger.Errorf(err, format, args...)
+		return
+	}
+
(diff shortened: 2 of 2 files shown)


provider/github/watcher.go
```diff
@@ -169,6 +180,37 @@ func (w *Watcher) watchLoop(
...
+	ctx context.Context,
+	m map[*lookout.RepositoryInfo]lastErrLog,
+	repo *lookout.RepositoryInfo,
+	errMsg string,
+	resp *github.Response,
+	err error,
+) {
+	l := m[repo]
+
+	asErr := l.err != err.Error() || time.Now().After(l.last.Add(errLogInterval))
```
dpordomingo: I thought we want to discard every repeated error on the same repos if they happen in the same time frame, don't we?

If I'm not mistaken, with the current implementation, every new error over the same repo will clean the throttler, won't it?

example:
```go
history := make(map[*lookout.RepositoryInfo]lastErrLog)
repo1 := &lookout.RepositoryInfo{"repo-1"}
w.logGitHubErr(ctx, history, repo1, "error A", resp, "")
w.logGitHubErr(ctx, history, repo1, "error B", resp, "")
w.logGitHubErr(ctx, history, repo1, "error A", resp, "")
w.logGitHubErr(ctx, history, repo1, "error C", resp, "")
```

will log:
```
error A
error B
error A
error C
```

but it should be:
```
error A
error B
error C
```

provider/github/watcher.go
```diff
@@ -169,6 +180,37 @@ func (w *Watcher) watchLoop(
 	}
 }
 
+// logGitHubErr logs an ErrGitHubAPI at debug or error level, depending on the
+// last time it was logged on error level
+func (w *Watcher) logGitHubErr(
```
dpordomingo: I don't like such cryptic signatures and var names :disappointed: 

I'd prefer something like:
```go
const errLogThrottlingInterval = 5 * time.Minute // because we're throttling

type lastErrLog struct {
	err  string
	when time.Time // instead of `last`
}

func (w *Watcher) logGitHubErr(
	ctx context.Context,
	history map[*lookout.RepositoryInfo]lastErrLog, // more meaningful
	repo *lookout.RepositoryInfo,
	resp *github.Response,
	err error,
	msg string, // we're logging a msg; this way there is no duplicate `err` in var names
) {
	lastRepoError := history[repo] // I find this way easier to remember

        var shouldBeLogged bool // I don't understand `asErr` meaning
	shouldBeLogged := lastRepoError.err != err.Error() ||
		time.Now().After(lastRepoError.when.Add(errLogThrottlingInterval))

        ...
```

But whatever you prefer :wink:

dpordomingo: :up: this is a friendly suggestion, not a blocker one ;)
Feel free to mark it as `resolved` if it does not suit you.

provider/github/watcher.go
carlosms: The current PR tries to deal with the scenario described in #385 only.

If the repo does not exist, you will get a lot of 404, and it's unlikely you will get any other error message. That could also happen if a repo you have configured suddenly becomes private and your token does not grant access to it. Or if you are offline.

For a misconfiguration (wrong repo url, wrong repo credentials, you are offline) you don't want to see the same message thousands of times.

For the scenario you describe, I can't think of a use case. If you are getting a different error each time, it's unlikely any of those errors are from a misconfiguration, and I'd say we probably want to log all of them.

provider/github/watcher.go
carlosms: :+1: I added most of your suggestions in da8b256.

carlosms:
> Second I think it's not the best solution for the issue.
> This issue happens only when user using token auth and manual repos configuration. If repo returns 404 it's misconfiguration. I believe in such case lookout should exit with the error, same as we exit when wrong db connection string provided for example.
> 
> Do you have any arguments to do it this way instead of exit?

We could exit if we did some checks before starting the watch loop, as a new initialization step. In that case yes, it makes sense.

But exiting whenever we get a 404 is too drastic. It may happen that you delete a repo you are watching. Or that you make a repo private and the token does not authorize you to list it. There might be other cases.
If we exit we will stop analyzing other repositories that may be working fine.

provider/github/watcher.go
dpordomingo: I find more usual/expected the behavior I described:
> _avoid repeated errors in a time frame_

than the implemented one:
> _avoid two errors being repeated in a row in a time frame_

On the other hand, the implementation of the first one would not be way more complicated; it would only require, for example, to store a deep `map[*lookout.RepositoryInfo]map[string]lastErrLog` instead of the current plain `map[*lookout.RepositoryInfo]lastErrLog`; or to store a plain map indexed by the mix of repo and err.

dpordomingo:
About @smacker [suggestion](https://github.com/src-d/lookout/pull/419#pullrequestreview-186479529)
a) I like more the idea of having a reusable throttler, but I'd not reject the PR for that either;
b) I agree with @carlos about [that](https://github.com/src-d/lookout/pull/419#issuecomment-448556259): it's not a misconfiguration that source{d} Lookout cannot reach the passed repos.

smacker:
> We could exit if we did some checks before starting the watch loop, as a new initialization step. In that case yes, it makes sense.

so that is exactly my proposal: add a check on initialization. That would solve the issue better.

And we still can have throttler for other weird cases.

carlosms:
I've made @smacker's request in 31659d7. To be honest I think the new code is harder to understand, and it's unlikely we will use it for anything else, since this error flooding is a specific problem of the GitHub watcher using polling. But at the same time I don't see it too different as to keep discussing it, so there it is.

About adding extra checks on initialization, I would consider it for later. The same functionality right now is similar, when you run lookout you see the 404 errors inmediately.

provider/github/watcher.go
```diff
@@ -177,9 +180,14 @@ func (w *Watcher) processRepoPRs(
...
-		ctxlog.Get(ctx).With(log.Fields{
+		if w.lastErrPR[repo] == nil {
+			w.lastErrPR[repo] = &errThrottlerState{}
+		}
+
+		newErrThrottlerLogger(ctxlog.Get(ctx), w.lastErrPR[repo]).With(log.Fields{
 			"repository": repo.FullName, "response": resp,
-		}).Errorf(err, "request for PR list failed")
+		}).Errorf(fmt.Errorf(err.Error()), "request for PR list failed")
+
```
smacker:
```go
fmt.Errorf(err.Error())
```
I'm confused. Why do you need it?

carlosms: To remove the stack trace.

With this change we log:
```
[2018-12-26T09:43:08.663549151Z] ERROR request for PR list failed app=lookoutd error=github api error: GET https://api.github.com/repos/nope/nope/pulls: 404 Not Found [] repository=nope/nope response=github.Rate{Limit:5000, Remaining:4998, Reset:github.Timestamp{2018-12-26 10:43:08 +0000 WET}}
```

With `.Errorf(err, `, we log:
```
[2018-12-26T09:44:38.594569159Z] ERROR request for PR list failed app=lookoutd error=github api error: GET https://api.github.com/repos/nope/nope/pulls: 404 Not Found []

github.com/src-d/lookout/provider/github.(*Watcher).doPRListRequest
	/home/cmartin/go/src/github.com/src-d/lookout/provider/github/watcher.go:308
github.com/src-d/lookout/provider/github.(*Watcher).processRepoPRs
	/home/cmartin/go/src/github.com/src-d/lookout/provider/github/watcher.go:181
github.com/src-d/lookout/provider/github.(*Watcher).processRepoPRs-fm
	/home/cmartin/go/src/github.com/src-d/lookout/provider/github/watcher.go:132
github.com/src-d/lookout/provider/github.(*Watcher).watchLoop
	/home/cmartin/go/src/github.com/src-d/lookout/provider/github/watcher.go:151
runtime.goexit
	/usr/lib/go-1.11/src/runtime/asm_amd64.s:1333 repository=nope/nope response=github.Rate{Limit:5000, Remaining:4993, Reset:github.Timestamp{2018-12-26 10:43:08 +0000 WET}}
```

It's an API error in a known place ("request for PR list failed"), so there is not need to flood the logs with extra debug information.

smacker: Maybe use Warn level then? If this error is not "real" error...

carlosms: Why is it not a "real" error?

smacker: Errors should be rendered with stack trace according to our logging library.
You work around it using
```go
fmt.Errorf(err.Error())
```
which is a hack that makes the error "not-really-error" but a more a warning in terms of our logging lib

carlosms: Well, the way I see it, it's not a panic, but it's still a "real" error. Missconfiguration error, connectivity error, or auth error... but something the user will be interested in knowing and fixing. A warning level might be overlooked too easily.

I don't think removing the stack trace makes it a hack. If we didn't log it here and instead returned, it could be `fmt.Errorf("request for events list failed: %s", err)`. Just like you can add relevant information to an error, I don't see a problem removing extra information that we know is not relevant.

smacker:
```go
fmt.Errorf(err.Error())
```
does look like a hack and WTF. I'm pretty sure anybody reading this code will have the same question as I did.

I don't mind to merge it now. But it does NOT look right.

smacker: a cleaner version of the same code (though I have doubts it's should be written like that at all):
```go
UserError(err)
```

smacker: also better than right now 👍

carlosms: done in 994c357

PR #420, opened 2018-12-18
Improve log message levels and error messages

### LINKED ISSUE #287: Improve observability / error messages

carlosms:
Fix #287.

### DIFF
--- a/cmd/dummy/main.go
+++ b/cmd/dummy/main.go
@@ -89,7 +89,7 @@ func (c *ServeCommand) initHealthProbes() {
 		log.With(log.Fields{
 			"addr":  c.ProbesAddr,
 			"paths": []string{livenessPath, readinessPath},
-		}).Debugf("listening health probe HTTP requests")
+		}).Infof("listening to health probe HTTP requests")
 
 		err := http.ListenAndServe(c.ProbesAddr, nil)
 		if err != nil {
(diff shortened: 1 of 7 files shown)


service/git/syncer.go
```diff
@@ -78,7 +78,7 @@ func (s *Syncer) Sync(ctx context.Context,
 }
 
 func (s *Syncer) fetch(ctx context.Context, repoInfo *lookout.RepositoryInfo, r *git.Repository, refspecs []config.RefSpec) (err error) {
-	ctxlog.Get(ctx).Infof("fetching references for repository %s: %v", repoInfo.CloneURL, refspecs)
+	ctxlog.Get(ctx).Debugf("fetching references for repository %s: %v", repoInfo.CloneURL, refspecs)
```
smacker: could you please check if we have any `info` level message somewhere in the callers that explains downloading?
I remember we make this one as `info` because fetching is a slow operation and it looks like lookoutd just hanged without it.

carlosms: We still have `level=info msg="creating local repository for: .....git"`. But reading again I'm realizing now that message is only logged on repo initialization, the next time we fetch a new PR branch you will not see anything.

Fixed in 7359f3a.

Issue #399, opened 2018-12-07
Log in with GitHub accounts


carlosms:
Umbrella issue, depends on #397.
This issue is to implement the basics for GitHub user login, with the following tasks:
- [x] GitHub Oauth login:
  - Serve a `/login` screen with a "login with GitHub" button
  - Endpoint in the server to perform the login using OAuth. This can be similar to what we do in https://github.com/src-d/code-annotation
  - If the login is successful we redirect to `/`; if not to `/unauthorized` (or similar name). For now it can be a blank page with a simple "authenticated"/"authentication error" text.
- [x] Manage access authorization and sessions
  - The server will return a JWT
  - The client will save it in localStorage
  - Unauthorized access to `/` should redirect to `/login`
  - A client should be able to access `/` directly if the token is valid
  - The `/` page will have a "log out" button

smacker:
done

PR #433, opened 2018-12-26
add client-side router

smacker:
Ref: #399


carlosms:
link to issue: #399

carlosms:
Continuing the conversation from #425.
The code looks good and not too complex. And it opens the door for a future UI design with proper routes like `/org-name/settings`, `/org-name/repo-name/settings`.

But still, it adds bloat that I don't know if we want to add at this stage (basic UI was the key requirement). Besides losing the nice `/login` and `/logout` routes in the address bar, are there any downsides to not use the router, @smacker? How would the auth management work?

smacker:
There are no technical limitations to do everything without a router. It works just fine on the previous PR. Most probably it would require only new logout method of App component and pass it down as props to every component that need to do logout (including failed api calls). And of course there are some more hacky solutions like just `Token.delete() && window.reload()`.

I see this advantages of adding the router on the current stage (from the biggest to the smallest in my option):
- It might be much of a headache to add a router to an existing project that didn't have it before. Due to synchronization of the URL and internal state.
- A router naturally introduces abstraction of pages and separation of them from components. I find it useful while building web applications.
- Nice urls

And the disadvantages:
- it requires a bit more code due to better structure / new page & route abstractions

As a conclusion maybe I wouldn't implement it right now if there were no requirements for clean url. But now when it's already here I see more benefits than harm by merging it.

frontend/src/services/auth.ts
```diff
@@ -0,0 +1,57 @@
...
+export interface User {
+  name: string;
+}
+
+const localStorageKey = 'token';
+
+class AuthService {
+  private _user: User | null = null;
+
+  get token() {
```
carlosms: should the getters in this class have return type?

carlosms: I'm not sure right now. I guess if it's good enough for the `strict` setting in `tsconfig` we can keep using the inferred return; we can always add them later if we see we prefer the explicit types.

Issue #421, opened 2018-12-19
create http go-log middleware


smacker:
Currently, we use pressly/lg middleware that is deprecated and uses logrus instead of src-d go-log.
Let's create simple http middleware using go-log directly in lookout and when it works propose to move it to go-log repo or separate one, so we can reuse it in other projects.

PR #435, opened 2018-12-26
Request logger middleware

smacker:
Fix: #421


web/auth.go
```diff
@@ -12,7 +12,7 @@ import (
 	jwt "github.com/dgrijalva/jwt-go"
 	"github.com/dgrijalva/jwt-go/request"
 	"github.com/gorilla/sessions"
-	"github.com/pressly/lg"
+	"github.com/src-d/lookout/util/ctxlog"
```
dpordomingo: Since `ctxlog.RequestLogger` is going to be used in `web` I'd move it under `web` dir.
Is there any plan to use it elsewhere?

smacker: I would keep it here because the middleware belongs to `ctxlog` :)
So most probably it's about personal preferences.
Let's hear one more opinion from @carlosms and move/keep depends on it.

carlosms: To me it feels more like something makes use of `ctxlog`. But I don't see any problem with it in the `ctxlog` package, even if it's not my initial intuition.

smacker:
Blocked by: https://github.com/src-d/lookout/issues/444

smacker:
I rebase on master, made some changes according to the changes in master, moved middleware to `web` package. Please, another pass.

//cc @carlosms @dpordomingo

PR #437, opened 2018-12-27
Remove web build from main makefile

carlosms:
Currently, drone builds are broken because `yarn` is missing in the image we use to build.

Instead of fixing the build, I thought it made sense to remove the web targets from the main makefile for now, until web is more stable and we want to include it in the released binary.
A normal `make build` will still create the `lookoutd web` server, but accessing returns 404.

The web targets can be used like this:
```
make -f Makefile.web build
make -f Makefile.web web-start
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,6 +62,12 @@ jobs:
       os: osx
       osx_image: xcode9.4
       before_install: skip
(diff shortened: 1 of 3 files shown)


carlosms:
go-bindata was missing, fixed in b8426f6.

Issue #248, opened 2018-09-13
Add ID metadata to gRPC calls, and use it in logs


carlosms:
Currently the logs for actions started by a gRPC call initialized by an analyzer do not contain any information to link it to a specific event or PR. Maybe we can include some metadata to the `NotifyReviewEvent` call and have the analyzers send it back when they call `GetChanges`.

Context in #245 and #181:

> The service/git methods now get the log from the context, but the problem here is that the gRPC requests come from the analyzer, and we don't have the event-id or any other field.
> I didn't look much into it, so I don't know if there is a way to pass some ID fields to the analyzer and get them back when they call GetFiles/GetChanges without modifying the proto messages.

> Grpc allows to send metadata in [header](https://godoc.org/google.golang.org/grpc#Header) or [trailer](https://godoc.org/google.golang.org/grpc#Trailer). There are some helpers in [go-grpc-middleware](https://github.com/grpc-ecosystem/go-grpc-middleware) for that.
> We can include it in our sdk later to pass `event-id` to an analyzer and send it back.

bzz:
As discussed today, it may be a good idea to put in backlog  so @dpordomingo could take over this one

PR #439, opened 2018-12-27
Use log Fields in ctxlog instead of logger

### LINKED ISSUE #427: Do not store logger in context, store log fields instead
As discussed in the last sync meeting, we should not store a logger in the context, only log fields from which we can create a logger when we need it.

Related to #375, #359.

carlosms:
Fix #427.

I couldn't find a way to use go-log for the fields merge.

### DIFF
--- a/util/ctxlog/context.go
+++ b/util/ctxlog/context.go
@@ -6,30 +6,42 @@ import (
 	log "gopkg.in/src-d/go-log.v1"
 )
 
-type ctxLogger int
+type ctxKey int
 
-// LoggerKey is the key that holds logger in a context.
-const LoggerKey ctxLogger = 0
+// logFieldsKey is the key that holds log Fields in a context.
+const logFieldsKey ctxKey = 0
 
-// Get returns logger from context or DefaultLogger if there is no logger in context
+// Get returns a logger configured with the context log Fields or the default fields
 func Get(ctx context.Context) log.Logger {
-	if v := ctx.Value(LoggerKey); v != nil {
-		return v.(log.Logger)
+	var fields log.Fields
+
+	if v := ctx.Value(logFieldsKey); v != nil {
+		fields = v.(log.Fields)
 	}
 
-	if log.DefaultLogger == nil {
-		log.DefaultLogger = log.New(nil)
+	return log.New(fields)
+}
+
+// WithLogFields returns a context with new logger Fields added to the current
+// ones, and a logger
+func WithLogFields(ctx context.Context, fields log.Fields) (context.Context, log.Logger) {
+	var newFields log.Fields
+
+	if v := ctx.Value(logFieldsKey); v != nil {
+		newFields = v.(log.Fields)
+
+		for k, v := range fields {
+			newFields[k] = v
+		}
+	} else {
+		newFields = fields
 	}
-	return log.DefaultLogger
+
+	ctx = set(ctx, newFields)
+	return ctx, Get(ctx)
 }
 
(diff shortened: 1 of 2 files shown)


util/ctxlog/context.go
```diff
@@ -3,33 +3,46 @@ package ctxlog
...
-func WithLogFields(ctx context.Context, f log.Fields) (context.Context, log.Logger) {
-	logger := Get(ctx).With(f)
-	return Set(ctx, logger), logger
+// WithLogFields returns a context with new logger Fields added to the current
+// ones, and a logger
+func WithLogFields(ctx context.Context, fields log.Fields) (context.Context, log.Logger) {
+	newFields := fields
+
+	if v := ctx.Value(logFieldsKey); v != nil {
+		// let logrus merge the previous and new fields
```
smacker: huh? Why not just:
```go
for k, v := range prevFields {
    newFields[k] = v
}
```
?

carlosms: Because of things like this, and any other future corner cases
https://github.com/sirupsen/logrus/commit/08e8d6501d300248a75c6375cfd3444b2469bb71#diff-32f22e4789509e4288e8ee93f532fbd0

We could do the `for` you posted and trust that `log.New` will sanitize any problems when we need the logger. But it seemed safer to also store the `log.Fields` already sanitized.

carlosms: Right now it's very localized and the `for` will work fine, but maybe in the future having different log fields in the context value and the logger becomes a problem for whatever reason. For #357 and #248 we'll have to create a method to access to the ctxlog log fields directly instead of the logger.

I don't see any downside to use the available sanitizing. If/when `go-log` changes the backend, we can just as easily change it here.

smacker: I see breaking an abstraction here. Which I'm quite strongly against of.

So I have 2 points against using logrus here:
1. Breaking an abstraction. What is the point of `go-log` lib if we lock ourselves on particular implementation details of it (logrus). Let's use logrus directly then.
2. This code in logrus exists only because of implementation details of JSON formatter. And even with it it doesn't cover all cases. For example, if we occasionally pass struct with a function inside, code of #357 might still fail (I didn't read implementation of `msgpack.Marshal` inside `go-queue`). But default log in dev mode will work.

If you want to have the same sanitazion in this place as in logrus I would prefer if you just copy-past it then using logrus directly.

carlosms: We are breaking abstraction anyway. `go-log` does not allow access to the logger `log.Fields`, so what we are doing here is making assumptions of the internals in any case.

I am assuming `go-log` uses `logrus`. But you are also assuming that the fields merge is done replacing older values. Maybe tomorrow `go-log` or `logrus` makes a change and now `With(fields)`, when it finds a repeated log field key, joins the previous and new values with a `,`.

Even with the small snippet you propose:
```go
for k, v := range prevFields {
    newFields[k] = v
}
```
There is a bug. In case of repeated keys `go-log` (via `logrus`) keeps the new value. With that snippet the older value would be kept instead. So we take the log fields merge problem -even if it's such a small problem-, already solved by a library, and repeat the work ourselves.

Anyway I don't agree with you but I aslo don't see the point of keeping discussing such a small thing. Change done in 0b4202b

smacker: `log.Fields` is a public interface of `go-log`. It's absolutely fine to use it.

There is no bug. It's our code that written in this or that way. `go-log` doesn't provide any public API for merging fields so it's on us. Our code doesn't have to repeat how `go-log` works with `Fields`.

Issue #331, opened 2018-10-25
Failed to obtain reader in the logs


smacker:
```
Failed to obtain reader, Failed to marshal fields to JSON, json: unsupported type: func(io.Reader)
```

this error isn't formatted as json, so I guess the error happens during logging.

carlosms:
What is the context of that error? Is that from staging? Your installation? which version? Is that after performing some action, or on startup?

smacker:
staging. the version is master.

Because the error doesn't contain any info besides above I don't know when exactly it happens. Context around says during fetching events but looks like it doesn't happen always.

the PR should help to debug it.

smacker:
Don't see this error on staging anymore (not exactly this but the one related to logging a function).

Let's wait few days if there is still no error we can close it as "can't reproduce". It's possible that error was fixed in some previous commit (apparently lookout was deploying on different cluster since some days ago)

smacker:
Still nothing related in logs on new staging.

smacker:
I see this error in staging logs again:
```
Failed to obtain reader, Failed to marshal fields to JSON, json: unsupported type: func() (io.ReadCloser, error)
```

not sure why new logrus version doesn't provide correct logs for this error. We need to investigate.

smacker:
When I try to reproduce it locally adding function with the same signature into logs I get:
```
{"app":"lookoutd","level":"info","logrus_error":"can not add field \"func\"","msg":"my-test","time":"2018-12-05T17:13:34.32557+01:00"}
```

PR #440, opened 2018-12-27
fix: Failed to obtain reader in the logs

smacker:
I have finally found it.

Ref: https://github.com/src-d/lookout/issues/331


carlosms:
So the problem was a nil pointer to a struct, not to a function?

smacker:
No. There is a function inside the structure. I couldn't find it easily because it's defined deep inside in the structure (in `http.Request` to be precise).

Issue #413, opened 2018-12-12
Configurable bblfsh & drivers versions


EgorBu:
Hello,
For our analyzers, we need to use specific versions of drivers. It will be great to make configurable.
Versions:
```
bblfshd "bblfsh/bblfshd:v2.10.0"
javascript "docker://bblfsh/javascript-driver:v1.2.0"
```

smacker:
blocked by https://github.com/src-d/charts/issues/128

PR #441, opened 2018-12-28
Lock bblfsh server and js driver versions on staging

smacker:
Style-analyzer requires specific versions to work correctly.

Fix: https://github.com/src-d/lookout/issues/413


carlosms:
should we add the cpp driver?

smacker:
added

PR #442, opened 2018-12-31
Improves logging message for failed analysis and removes duplicated logic

se7entyse7en:

PR #443, opened 2019-01-02
Removes duplicated logic in server module

se7entyse7en:
This depends on #442. Please see [here](https://github.com/src-d/lookout/pull/442#issuecomment-450660906).

### DIFF
--- a/cmd/server-test/dummy_analyzer_test.go
+++ b/cmd/server-test/dummy_analyzer_test.go
@@ -37,7 +37,7 @@ func (suite *DummyIntegrationSuite) TearDownTest() {
 func (suite *DummyIntegrationSuite) TestSuccessReview() {
 	suite.sendEvent(successJSON)
 	suite.GrepAll(suite.r, []string{
-		`processing pull request`,
+		`processing event type 2`,
 		`{"analyzer-name":"Dummy","file":"another.go","line":3,"text":"This line exceeded`,
 		`status=success`,
 	})
@@ -71,7 +71,7 @@ func (suite *DummyIntegrationSuite) TestReviewDontPostSameComment() {
 	suite.sendEvent(rev1Event.String())
 	suite.GrepAndNotAll(suite.r,
 		[]string{
-			`processing pull request`,
+			`processing event type 2`,
 			`{"analyzer-name":"Dummy","file":"dummy.go","text":"The file has increased`,
 			`status=success`,
 		}, []string{
@@ -104,7 +104,7 @@ func (suite *DummyIntegrationSuite) TestSuccessPush() {
 	}
 	suite.sendEvent(pushEvent.String())
 	suite.GrepAll(suite.r, []string{
-		"processing push",
+		"processing event type 1",
 		"comments can belong only to review event but 1 is given",
 		`status=error`,
 	})
--- a/server/server.go
+++ b/server/server.go
@@ -99,14 +99,7 @@ func (s *Server) HandleEvent(ctx context.Context, e lookout.Event) error {
 	// we need to retry analyzis but post only new comments (poster should handle it)
 	safePosting := status == models.EventStatusPosting
 
-	switch ev := e.(type) {
-	case *lookout.ReviewEvent:
-		err = s.HandleReview(ctx, ev, safePosting)
-	case *lookout.PushEvent:
-		err = s.HandlePush(ctx, ev, safePosting)
-	default:
-		logger.Debugf("ignoring unsupported event: %s", ev)
-	}
+	err = s.AnalyzeAndComment(ctx, e, safePosting)
 
 	if err == nil {
 		status = models.EventStatusProcessed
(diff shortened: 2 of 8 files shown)


server/server.go
```diff
@@ -127,113 +120,91 @@ func (s *Server) HandleEvent(ctx context.Context, e lookout.Event) error {
...
-		return err
+	getAnalyzerNotifier := func(ctx context.Context, settings map[string]interface{}) (analyzerNotifier, error) {
+		st := grpchelper.ToPBStruct(settings)
+		switch ev := e.(type) {
+		case *lookout.ReviewEvent:
+			if st != nil {
+				ev.Configuration = *st
+			}
+			return analyzerNotifier{
+				func(ctx context.Context, a lookout.AnalyzerClient) (*pb.EventResponse, error) {
```
dpordomingo: We try to avoid structs initialized with positional arguments because they're harder to read and can cause problems in some cases (like changing the order of properties having the same type, or adding new properties).

imho I'd prefer:
```go
return analyzerNotifier{
  notify: func...
  timeot: s.analyzerReviewTimeout
}, nil
```

server/server.go
```diff
@@ -127,113 +120,91 @@ func (s *Server) HandleEvent(ctx context.Context, e lookout.Event) error {
...
+		case *lookout.PushEvent:
+			if st != nil {
+				ev.Configuration = *st
+			}
+			return analyzerNotifier{
+				func(ctx context.Context, a lookout.AnalyzerClient) (*pb.EventResponse, error) {
+					return a.NotifyPushEvent(ctx, ev)
+				}, s.analyzerPushTimeout}, nil
+		default:
+			ctxlog.Get(ctx).Debugf("ignoring unsupported event: %s", ev)
```
dpordomingo: I think we're moving the logic to validate the event too deep 

```go
s.HandleEvent(ctx, strangeEvent) ->
    s.AnalyzeAndComment(ctx, strangeEvent, isSafePosting) ->
        s.analyze(ctx, strangeEvent, isSafePosting) ->
            s.getSender(strangeEvent) ->
                getAnalyzerNotifier(ctx, errorMsgMap) ->
                    if unsupported return fmt.Errorf("unsupported")
```

instead of how we did initially:

```go
s.HandleEvent(ctx, strangeEvent) ->
    if unsupported fmt.Debug("unsupported")
```

I do not see the benefit of moving that validation so deep.

server/server.go
```diff
@@ -127,113 +120,91 @@ func (s *Server) HandleEvent(ctx context.Context, e lookout.Event) error {
...
+			return analyzerNotifier{}, fmt.Errorf("unsupported event: %s", ev)
```
dpordomingo: I might be mistaken, but this PR is also changing the `s.HandleEvent(ctx, strangeEvent)` behavior when it is sent an event that was not a `ReviewEvent` nor a `PushEvent`;
If I'm rightm the `s.HandleEvent(ctx, strangeEvent)` will fail with an `unsupported` error message, instead of just logging a debug `unsupported` message.

If that's true, I think we should not change the behavior in a refactor PR or without a clear reasons to do so.

server/server.go
```diff
@@ -127,113 +120,91 @@ func (s *Server) HandleEvent(ctx context.Context, e lookout.Event) error {
...
+			err = fmt.Errorf("posting analysis failed: %s", err)
+		} else {
+			s.status(ctx, e, lookout.SuccessAnalysisStatus)
+		}
+	}
+	return err
+}
 
+func (s *Server) analyze(ctx context.Context, e lookout.Event, safePosting bool) ([]lookout.AnalyzerComments, error) {
+	ctxlog.Get(ctx).Infof("processing event type %d", e.Type())
```
dpordomingo: I see more cryptic the new error message: `event type 2` instead of `pull request`

dpordomingo: I think we're also missing info about the event provider (see https://github.com/src-d/lookout/pull/443/commits/13266785477a3562816e9a9bcb19f38136552103#diff-91bbeda7eb98a7adc57b9e47e2cf5c2bL188)

server/server.go
```diff
@@ -127,113 +120,91 @@ func (s *Server) HandleEvent(ctx context.Context, e lookout.Event) error {
...
 
-// HandleReview sends request to analyzers concurrently
-func (s *Server) HandleReview(ctx context.Context, e *lookout.ReviewEvent, safePosting bool) error {
-	ctx, logger := ctxlog.WithLogFields(ctx, log.Fields{
-		"provider": e.Provider,
-	})
-	logger.Infof("processing pull request")
+func (s *Server) AnalyzeAndComment(ctx context.Context, e lookout.Event, safePosting bool) error {
+	comments, err := s.analyze(ctx, e, safePosting)
+	if err == nil {
```
dpordomingo: we can get rid of an identation level if we do:
```go
if err != nil {
    return err
}

if err := s.post(ctx, e, ....
```
On the other hand, it's the usual way to handle errors in go.

server/server.go
```diff
@@ -127,113 +120,91 @@ func (s *Server) HandleEvent(ctx context.Context, e lookout.Event) error {
...
+		if err := s.post(ctx, e, comments, safePosting); err != nil {
+			s.status(ctx, e, lookout.ErrorAnalysisStatus)
+			err = fmt.Errorf("posting analysis failed: %s", err)
```
dpordomingo: This assignment does not apply over [first `err` definition #line-124](https://github.com/src-d/lookout/pull/442/commits/52e1fa8a904ae92161e0dd977e46ed4428d2a723#diff-91bbeda7eb98a7adc57b9e47e2cf5c2bR124), but over [its own scope #line-126](https://github.com/src-d/lookout/pull/442/commits/52e1fa8a904ae92161e0dd977e46ed4428d2a723#diff-91bbeda7eb98a7adc57b9e47e2cf5c2bR126)

We avoid it, doing early returns; it means, that we stop processing as soon as an error condition is met; example:
```go
if err := doSomething(); err != nil {
    return fmt.Errorf("new deep error") // we stop processing
}
// do other things
```

instead of moving errors and reusing variables containing it, like in your example:
```go
var err error
if err := doSomething(); err != nil { // we shadow the upper err definition
    err = fmt.Errorf("new deep error") // this definition does not apply over the upper scope
}
return err // nothing was overriden; we got <nil>
```

You can see an example about it at: https://play.golang.org/p/I75A6xUtJgN

server/server.go
```diff
@@ -127,113 +120,91 @@ func (s *Server) HandleEvent(ctx context.Context, e lookout.Event) error {
...
+		} else {
+			s.status(ctx, e, lookout.SuccessAnalysisStatus)
+		}
+	}
+	return err
```
dpordomingo: we always introduce a new line after every code block

server/server.go
```diff
@@ -127,113 +120,91 @@ func (s *Server) HandleEvent(ctx context.Context, e lookout.Event) error {
...
+}
 
+func (s *Server) analyze(ctx context.Context, e lookout.Event, safePosting bool) ([]lookout.AnalyzerComments, error) {
+	ctxlog.Get(ctx).Infof("processing event type %d", e.Type())
+
+	var comments []lookout.AnalyzerComments
 	if err := e.Validate(); err != nil {
-		return err
+		return comments, err
```
dpordomingo: we use to avoid indirections when doing an early return caused by an error; example:

```go
func doSomething() (string, error) {
    var msg string
    // do things, operate with msg...
    // at  some point, we check an error, and we do:
    if err := checkSomething(); err != nil {
        return "", err // instead of `return msg, err`; relying on a previous value of `msg`
    }
}
```

That way, we do explicit the things that we return

server/server.go
```diff
@@ -127,113 +120,91 @@ func (s *Server) HandleEvent(ctx context.Context, e lookout.Event) error {
...
 	}
 
 	conf, err := s.getConfig(ctx, e)
 	if err != nil {
-		return err
+		return comments, err
```
dpordomingo: _idem._

server/server.go
```diff
@@ -127,113 +120,91 @@ func (s *Server) HandleEvent(ctx context.Context, e lookout.Event) error {
...
 		a lookout.AnalyzerClient,
 		settings map[string]interface{},
 	) ([]*lookout.Comment, error) {
-		st := grpchelper.ToPBStruct(settings)
-		if st != nil {
-			e.Configuration = *st
+		var comments []*lookout.Comment
+		analyzerNotifier, err := getAnalyzerNotifier(ctx, settings)
+		if err != nil {
+			return comments, err
```
dpordomingo: _idem._ about returning errors.

dpordomingo:
I saw some messages from `lookout-staging` bot.
Some of them may be addressed, and others might not. In any case, we should proceed as they were posted by a real user:
- addressing them if possible,
- explaining why they can not be addressed, or
- marking them as `resolved` if they make no sense and no explanation is needed.

se7entyse7en:
> I also wonder if the initial implementation was coded that way (with that dupe code) considering any planned change, that might cause that both event handlers were totally different.

Exactly, I was thinking the same thing that I could have been missing the big picture and future planned change.

server/server.go
se7entyse7en: Agree

server/server.go
se7entyse7en: That's another point against the refactor. The reason for this was to avoid checking twice the type of the event using the `switch`.

server/server.go
se7entyse7en: Agree on this too, in fact an idea could be to add a `stringer` to `EventType` in `lookout-sdk`.

se7entyse7en: You're right, I forgot to add it in to the log ctx.

server/server.go
se7entyse7en: You're right. With these changes the `HandleEvent` method fails in case on supported event. Another thing to note with this refactor is that in case of unsupported event the check is so deep that it even spawns a goroutine for each analyzer. And this is a big no-no in my opionion.

cmd/server-test/dummy_analyzer_test.go
```diff
@@ -37,7 +37,7 @@ func (suite *DummyIntegrationSuite) TearDownTest() {
 func (suite *DummyIntegrationSuite) TestSuccessReview() {
 	suite.sendEvent(successJSON)
 	suite.GrepAll(suite.r, []string{
-		`processing pull request`,
+		`processing event type 2`,
```
smacker: hm. The previous message was much better.

smacker:
Usually, it's a good idea to remove duplicated code, but I see 2 problems with this PR:

1. Before there was a much more clear separation of handling different events. Despite the fact that right now we handle them in a very similar manner it can be not true anymore in the future. (for example when we start processing push event instead of just ignoring it) I believe this refactoring would make such potential changes very difficult. Same applies to new types of events.
2. As it was already mentioned the check is too deep (it runs go-routines for unsupported events!)

Though I agree a current code of this module needs some improvements.

server/server.go
carlosms: In other place we are doing `reflect.TypeOf(e).String()`

se7entyse7en:
I'm gonna close this PR ;)

PR #448, opened 2019-01-04
Improve web config file error messages

carlosms:
With the current error in master, you get this, which is not so obvious:
```console
$ ./build/bin/lookoutd web 
provider github client_id is required
```

### DIFF
--- a/cmd/lookoutd/web.go
+++ b/cmd/lookoutd/web.go
@@ -50,13 +50,13 @@ func (c *WebCommand) Execute(args []string) error {
 
 	ghConfg := conf.Providers.Github
 	if ghConfg.ClientID == "" {
-		return fmt.Errorf("provider github client_id is required")
+		return fmt.Errorf("Missing field in configuration file: provider github client_id is required")
 	}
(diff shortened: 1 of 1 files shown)


cmd/lookoutd/web.go
```diff
@@ -50,13 +50,13 @@ func (c *WebCommand) Execute(args []string) error {
 
 	ghConfg := conf.Providers.Github
 	if ghConfg.ClientID == "" {
-		return fmt.Errorf("provider github client_id is required")
+		return fmt.Errorf("Missing field in configuration file: provider github client_id is required")
```
se7entyse7en: I seem to remember that it's a good practice to start the error message with a lowercase.

smacker: absolutely true. Though in this particular case it's not that important because it's top-level error.

carlosms: Yes. I made it uppercase for consistency with the other top-level error messages above this one. Maybe we could change all of the error messages to lowercase at some point.

PR #438, opened 2018-12-27
Replace go-bindata with esc

smacker:
Fix: #426

PR #450, opened 2019-01-08
WIP docs enhancements

dpordomingo:
WIP PR; will be ready along the day


carlosms:
what is TFM?

docs/architecture.md
```diff
@@ -33,4 +33,25 @@ An analyzer is a gRPC service that will be called by the [Server](#server) to pe
 
 They are not part of **source{d} Lookout** repository so they can be developed by third parties.
 
-source{d} Lookout Server will call all the registered Analyzers to produce comments for the opened Pull Request in the watched repositories. To register new Analyzers in the configuration file, `lookoutd` will need to be restarted.
+**source{d} Lookout** Server will call all the registered Analyzers to produce comments for the opened Pull Request in the watched repositories. To register new Analyzers in the configuration file, `lookoutd` will need to be restarted.
+
+
+# External services
```
carlosms:
```suggestion
# External Services
```

docs/architecture.md
```diff
@@ -33,4 +33,25 @@ An analyzer is a gRPC service that will be called by the [Server](#server) to pe
...
+
+**source{d} Lookout** will call some external services, some of them depending on **source{d} Lookout** configuration or on how it is run.
```
carlosms: **source{d} Lookout** is repeated too closely. How about
```suggestion
**source{d} Lookout** will call some external services, some of them depending on the configuration.
```

docs/architecture.md
```diff
@@ -33,4 +33,25 @@ An analyzer is a gRPC service that will be called by the [Server](#server) to pe
...
+
+## Babelfish
+
+Babelfish is used to parse files into [UAST](https://doc.bblf.sh/uast/uast-specification.html) when the [DataService](#dataservice) `GetChanges` or `GetFiles` methods are asked to include UASTs.
+
+Babelfish can also be directly called through **DataService** proxy.
```
carlosms: Do we have a link for the babelfish proxy methods?

docs/architecture.md
```diff
@@ -33,4 +33,25 @@ An analyzer is a gRPC service that will be called by the [Server](#server) to pe
...
+
+## RabbitMQ
+
+If **source{d} Lookout** is run in [distributed mode](github.com/src-d/lookout/docs/how-to-run.md#distributed-mode), the watcher will enqueue in RabbitMQ all new repository and PR events, and the workers will take from there the events to be processed.
```
carlosms: Why the full link? should this be:
```suggestion
If **source{d} Lookout** is run in [distributed mode](how-to-run.md#distributed-mode), the watcher will enqueue in RabbitMQ all new repository and PR events, and the workers will take from there the events to be processed.
```

docs/architecture.md
```diff
@@ -33,4 +33,25 @@ An analyzer is a gRPC service that will be called by the [Server](#server) to pe
...
+
+## PostgreSQL
+
+**source{d} Lookout** uses PostgreSQL to store application internal state. There will be stored processed events (`ReviewEvent` and `PushEvent`), their targets (`ReviewTarget`), and the comments (`Comment`) returned by the Analyzers.
```
carlosms: My preference is that we should not go into details of what we store and what we use it for. The way I see it if we document it here, in architecture, we are implying you may be interested in the DB contents, or that you may want to manipulate the DB.

I propose
```suggestion
**source{d} Lookout** uses PostgreSQL to store the application internal state.
```

And also to remove below
`It is used to avoid processing the same PR twice for the same push and having duplicate comments.`

docs/CONTRIBUTING.md
```diff
@@ -84,6 +84,48 @@ $ make test-json
 ```
 
 
+## Web Interface
+
+If you want to contribute in **source{d} Lookout** [Web Interface](web.md), you should consider the following:
+
+### Dependencies
+
+It is required [Node.js](https://nodejs.org/es) `>=8`; you can check which version of `node` do you have, running:
```
carlosms:
```suggestion
[Node.js](https://nodejs.org) `>=8` is required; you can check which version of `node` do you have, running:
```

docs/CONTRIBUTING.md
```diff
@@ -84,6 +84,48 @@ $ make test-json
...
+
+```shell
+$ node -v
+v10.15.0
+```
+
+### Running
+
+In case you want to locally run the web interface from sources, you can choose one of the following:
+- using web assets from `bindata`, that will require you restart the server everytime you change any web asset.
```
carlosms: It's not yet merged, but this will not be `bindata` after #438.

docs/lookout-sdk.md
```diff
@@ -40,6 +40,8 @@ To perform a `NotifyPushEvent` call to an analyzer and serve the **source{d} Loo
 $ lookout-sdk push
 ```
 
+`lookout-sdk` will exit with a non `0` exit code if the called analyzer did not responded or if it returned an error.
```
carlosms: I think in case the analyzer does not respond, the execution will not end, because we don't have any timeouts.
What about:
```suggestion
`lookout-sdk` will exit with a non `0` exit code if the gRPC call to the analyzer ends with an error.
```

docs/configuration.md
```diff
@@ -1,6 +1,19 @@
 # Configuring source{d} Lookout
 
-**source{d} Lookout** is configured with the `config.yml` file, you can use the template [`config.yml.tpl`](/config.yml.tpl) to create your own. Use the `lookoutd` option `--config` to set the path to it, or use the default location at `./config.yml`.
+The behavior of **source{d} Lookout** is configured with two different configuration files:
+
+- [`config.yml`](#config.yml), to define the global configuration of the server.
```
carlosms: Link does not work, at least in GitHub, I don't know about Gitbook.
```suggestion
- [`config.yml`](#configyml), to define the global configuration of the server.
```

docs/configuration.md
```diff
@@ -1,6 +1,19 @@
...
+- [`.lookout.yml`](#.lookout.yml), to override the analyzer default behavior in each repository.
```
carlosms: Same,
```suggestion
- [`.lookout.yml`](#lookoutyml), to override the analyzer default behavior in each repository.
```

docs/configuration.md
```diff
@@ -1,6 +1,19 @@
 # Configuring source{d} Lookout
 
-**source{d} Lookout** is configured with the `config.yml` file, you can use the template [`config.yml.tpl`](/config.yml.tpl) to create your own. Use the `lookoutd` option `--config` to set the path to it, or use the default location at `./config.yml`.
+The behavior of **source{d} Lookout** is configured with two different configuration files:
```
carlosms: Too many _configure_
```suggestion
The behavior of **source{d} Lookout** is defined with two different configuration files:
```

docs/configuration.md
```diff
@@ -1,6 +1,19 @@
...
+
+- [`config.yml`](#config.yml), to define the global configuration of the server.
+- [`.lookout.yml`](#.lookout.yml), to override the analyzer default behavior in each repository.
+
+
+# config.yml
+
+**source{d} Lookout** is configured with the `config.yml` file, you can use the template [`config.yml.tpl`](/config.yml.tpl) to create your own. Use the `lookoutd` option `--config` to set the path to it, or use the default location at `./config.yml`. The config file is read on server startup, so it is needed to be restarted to load a new configuration.
```
carlosms:
```suggestion
**source{d} Lookout** is configured with the `config.yml` file, you can use the template [`config.yml.tpl`](/config.yml.tpl) to create your own. Use the `lookoutd` option `--config` to set the path to it, or use the default location at `./config.yml`. The config file is read on server startup, so `lookoutd` needs to be restarted in order to load a new configuration.
```

docs/configuration.md
```diff
@@ -1,6 +1,19 @@
...
+
+The most important things you need to configure for a local installation, are:
+
+1. [Repositories](#repositories): add the URLs of the repositories to be watched.
```
carlosms: This is only the case if you are using GitHub token instead of GitHub App auth. This is getting a bit confusing, and I think we should consider separating both auth methods more clearly in the docs. If that case maybe it makes sense to leave this line as you propose, and then come back to it in a later PR.

config.yml.tpl
```diff
@@ -20,18 +21,18 @@ providers:
...
       # user:
       # token:
       # minInterval: 1m
 
-# only for authorization with GitHub App
+# web interface configuration, only available if authorizing with GitHub App
 web:
   # Secret key to sign JSON Web Tokens
-  signing_key:
+  signing_key: # a non empty string
```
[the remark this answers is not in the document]
dpordomingo: why? it avoid people leaving the line empty...

docs/architecture.md
dpordomingo: I think this section is for developers and people interested in the internals and how the app works. Based on that, I find it useful to explain the purpose of the Database and what's inside.
Are you strongly against adding this info? If so, I'll happily drop it :dancing_men:

docs/CONTRIBUTING.md
dpordomingo: already merged; working on that :dagger: 
thanks!

docs/lookout-sdk.md
dpordomingo: I meant: _if the analyzer could not be called_;
what about this other way?
```suggestion
`lookout-sdk` will exit with a non `0` exit code if the gRPC call to the analyzer
ends with an error or it could not be performed.
```

docs/configuration.md
dpordomingo: What if reworded this other way?
```suggestion
1. [Repositories](#repositories): define the repositories to be watched.
```
This way we only link to `Repositories` section, where it's explained the two different ways to specify the repositories to be watched.

config.yml.tpl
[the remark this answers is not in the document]
carlosms: But that's already taken care of by the comment above it, and https://github.com/src-d/lookout/blob/master/cmd/lookoutd/web.go#L59. Otherwise, why don't we add `non empty string` to each and every one of the fields?

docs/architecture.md
carlosms: The way I see this section, it's meant to be a high level overview of the architecture, useful for example for the admin that has to deploy it, or the developer of a new analyzer.

Going into detail of what the DB stores can be distracting for someone without the proper context (the knowledge of the internals). If we redefine this section to be meant for developers of this repo, we would probably have to explain how we use kallax for persistency...

docs/lookout-sdk.md
carlosms: To me they are the same: ends with an unreachable error code. But ok if you think it's more clear.

docs/configuration.md
carlosms: :+1:

docs/architecture.md
smacker: agree on keeping it high level

config.yml.tpl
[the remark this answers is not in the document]
dpordomingo: fair enough

docker-compose.yml
```diff
@@ -9,6 +9,7 @@ services:
     environment:
       GITHUB_USER: ${GITHUB_USER:-}
       GITHUB_TOKEN: ${GITHUB_TOKEN:-}
+      LOG_LEVEL: ${LOG_LEVEL:-error}
```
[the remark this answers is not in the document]
dpordomingo: ok, removing the commit and considering it for a separate PR

docs/architecture.md
dpordomingo: I believe that saying that it's stored there the comments returned by the analyzers is not low level, and is way high level than explaining how kallax works.
On the other hand, explaining which tables and contents are stored there (even which indexes does it contain) is also useful for admin because it can be useful for them to dimension the service.

Anyhow, I'll drop the parts that you don't like to avoid blocking this PR, and let that part for the future (if I find time or energies).

docs/configuration.md
dpordomingo: I asumed GitBook as the preferred way to render our docs, and it works differently when creating anchors for titles having `.` inside.
- Title: `config.yml`
- GitBook: `#config-yml`
- GitHub: `#configyml`

In case it is required to support both, would you propose an alternative?

docs/architecture.md
dpordomingo: I don't understand this suggestion. Link to source code? bblfsh docs? link to docs explaining the methods that have been proxied and how to use them?

dpordomingo:
Thanks @carlosms for your detailed review (and sorry for making a big PR :wink:)
I had to rebase to work over the new master (in order to take some of your latest changes, and also some from Vadim)
Your suggestions were addressed by 518d15ff4832224007c019a1ebe549860efa2023

Pending things that requires your feedback:
- [ ] Do we have a link for the babelfish proxy methods? [@discussion_r246323093](https://github.com/src-d/lookout/pull/450#discussion_r246323093)
- [ ] Anchors to titles with `.` inside (`#config-yml` vs `#configyml` for `config.yml` title) [@discussion_r246331372](https://github.com/src-d/lookout/pull/450#discussion_r246331372)

docs/architecture.md
carlosms: I was asking about some section we could link. I see we mention it in this file, in the DataService section, but I wanted to know if there is some other place where we define with more detail what methods can you use, an example or something. Because I think we don't have a more detailed explanation in the docs, but I wanted to ask if you know if we have something like that and maybe I didn't see it.

If the only place where we mention it is this file, then I don't think we need a link.

docs/configuration.md
carlosms: For this PR it would be fine if it only works on GitBook, in my opinion. But if the section links do not work on GitHub let's at least open an issue and try to fix it at some point. Maybe we just need to assume the limitations and avoid `.` in titles.

docs/architecture.md
dpordomingo: We don't have such explanation.
If you agree, I can create an issue to require new docs about how to use DataService. We could add examples about how to call it and what data to extract, and also a call to bblfsh using DataService proxy.
Or do you think that current examples in `lookout-sdk/examples/language-analyzer.*` is enough?

docs/configuration.md
dpordomingo: agree, issue created https://github.com/src-d/lookout/issues/469

docs/architecture.md
carlosms: let's create an issue and decide later the specifics :+1:

Issue #445, opened 2019-01-04
GetChanges does not return deleted files


carlosms:
Right now `GetChanges` does not return deleted files.
This issue is to decide if we want to change this or keep it.

It makes sense to filter out deleted files since analyzers are supposed to add comments to new changes, and maybe this edge case is not considered by the current analyzers.
On the other hand we don't remove deleted and context lines from diffs, so maybe a deleted file should also not dissapear from the response. And there might be a use case for an analyzer that is interested in knowing if any file was deleted in a PR.

Currently, the filtering is happening in the regexp scanner, [here](https://github.com/src-d/lookout/blob/d29e69dda99783b387558e1231d0844258f46f71/service/git/scanner.go#L173). If we want to keep this behaviour I think we should make it more explicit and filter in a more obvious place.

PR #451, opened 2019-01-08
fix: Do not skip deleted files in regexp filter

smacker:
Ref: https://github.com/src-d/lookout/issues/445

### DIFF
--- a/service/git/scanner.go
+++ b/service/git/scanner.go
@@ -171,7 +171,7 @@ func (s *regexpFilter) OnStart() error {
 
 func (s *regexpFilter) Fn(f *lookout.File) (bool, error) {
 	if f == nil {
-		return true, nil
+		return false, nil
 	}
 
 	if !s.matchInclude(f.Path) {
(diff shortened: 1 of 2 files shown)


carlosms:
What if I request `IncludePattern = vendor*` but the file `cmd/main.go` gets deleted? Will it be included in the response with the change in this PR?

smacker:
Good point.
I also see that in Vendor & Language filters we apply them on `Base` if `Head == nil`.

I've updated the PR implementing the same logic to be consistent and it actually makes sense at least to me.

service/git/scanner.go
```diff
@@ -217,6 +217,10 @@ func NewChangeFilterScanner(scanner lookout.ChangeScanner, include, exclude stri
 	}
 
 	fn := func(ch *lookout.Change) (bool, error) {
+		if ch.Head == nil {
```
se7entyse7en: Can you please explain when this happen?

smacker: When you delete a file in git.

carlosms:
LGTM. Should we have tests to check the filters apply to base when head is deleted?

PR #456, opened 2019-01-09
List user's organizations in web

### LINKED ISSUE #400: See the list of organizations
Depends on #399.
The goal of the issue is to be able to login with your GitHub account, and see a list of the organizations where the lookout GitHub App is installed, and you are an administrator.

We probably can use the current list of installations, the GitHub username from the JWT, and [this api endpoint](https://developer.github.com/v3/orgs/members/#get-organization-membership), (it is [available for GitHub Apps](https://developer.github.com/v3/apps/available-endpoints/#available-endpoints))

The UI needs to be defined, but for now a `<ul>` will be probably enough.

carlosms:
Fix #400.

This PR is as simple as possible. The UI is of course not much more than a place holder.

For the backend, I reused the existing json responses. [There is a comment](https://github.com/src-d/lookout/blob/master/web/auth.go#L20) to extend them into separate responses once we have more endpoints, but I thought it made sense to keep this PR small and do that later; possible when we have more endpoints and see the need more clearly.

There are possibly other optimizations that could be done in the backend, but for this PR I chose to ignore them. I'm talking about things like adding a cache for the list of installations, or even the list of organizations for a given user.

### DIFF
--- a/frontend/src/api.ts
+++ b/frontend/src/api.ts
@@ -80,3 +80,15 @@ interface MeResponse {
 export function me(): Promise<MeResponse> {
   return apiCall<MeResponse>('/api/me');
 }
+
+export interface Org {
+  name: string;
+}
+
+interface OrgsResponse extends Array<Org> {}
+
+// Returns a list of GitHub organization names where lookout is installed and
+// the logged-in user is an administrator
+export function orgs(): Promise<OrgsResponse> {
+  return apiCall<OrgsResponse>('/api/orgs');
+}
--- a/frontend/src/components/Organizations.tsx
+++ b/frontend/src/components/Organizations.tsx
@@ -1,0 +1,66 @@
+import React from 'react';
+import * as api from '../api';
+import { User } from '../services/auth';
+import Errors from './Errors';
+import Loader from './Loader';
+
+interface OrgsProps {
+  user: User;
+}
+
+interface OrgsState {
+  done: boolean;
+  orgs: api.Org[];
+  errors: string[];
+}
+
+class Organizations extends React.Component<OrgsProps, OrgsState> {
+  public state: OrgsState = {
+    done: false,
+    orgs: [],
+    errors: []
+  };
+
+  public componentDidMount() {
+    return api
+      .orgs()
+      .then(resp => {
+        this.setState({
+          done: true,
+          orgs: resp,
+          errors: []
+        });
+      })
+      .catch(err => {
+        this.setState({
+          done: true,
+          orgs: [],
+          errors: err
+        });
+      });
+  }
+
+  public render() {
+    if (!this.state.done) {
+      return <Loader />;
+    }
+
+    if (this.state.errors.length > 0) {
+      return <Errors errors={this.state.errors} />;
+    }
+
+    const orgs = this.state.orgs.map(org => <li key={org.name}>{org.name}</li>);
+
+    return (
+      <div>
+        <h1>
+          Organizations with source
+          {'{d}'} Lookout installed where {this.props.user.name} is admin:
+        </h1>
+        <ul>{orgs}</ul>
+      </div>
+    );
+  }
+}
+
+export default Organizations;
--- a/web/auth.go
+++ b/web/auth.go
@@ -172,6 +172,7 @@ func (a *Auth) Middleware(next http.Handler) http.Handler {
 		}
 
 		ctx := context.WithValue(r.Context(), userIDKey, claims.ID)
+		ctx = context.WithValue(ctx, userLoginKey, claims.Login)
 		ctx = context.WithValue(ctx, userOAuthToken, &claims.OAuthToken)
 		next.ServeHTTP(w, r.WithContext(ctx))
 	})
@@ -234,13 +235,14 @@ func getGithubUser(client *http.Client) (*User, error) {
 
 type jwtClaim struct {
 	ID         int
+	Login      string
 	OAuthToken oauth2.Token
 	jwt.StandardClaims
 }
 
 // makeToken generates token string for a user
 func (a *Auth) makeToken(ot oauth2.Token, user *User) (string, error) {
-	claims := &jwtClaim{ID: user.ID, OAuthToken: ot}
+	claims := &jwtClaim{ID: user.ID, Login: user.Login, OAuthToken: ot}
 	t := jwt.NewWithClaims(jwt.SigningMethodHS256, claims)
 	ss, err := t.SignedString(a.signingKey)
 	if err != nil {
@@ -268,8 +270,12 @@ var extractor = &request.PostExtractionFilter{
 
 type userContext int
 
-const userIDKey userContext = 1
-const userOAuthToken userContext = 2
+const (
+	_ userContext = iota
+	userIDKey
+	userLoginKey
+	userOAuthToken
+)
 
 // getUserInt gets the value stored in the Context for the key userIDKey, bool
 // is true on success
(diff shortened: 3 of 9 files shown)


smacker:
Before the code review: why do we show only already installed organizations?

Isn't it opposite of the purpose of the web interface: allow enabling repos without going to github app page, similar to travis?

smacker:
Okay. Thanks for the clarification! Though here is a follow up question about how it's going to work.

When I install app from marketplace on organization it asks me if I want to enable it on all repositories or only some of them:

screenshot 2019-01-10 at 18 39 50

From the current implementation, as I understand, if a user chose only "some" repos it will be impossible to enable new repos from web-ui. Please confirm if it is correct or incorrect.

And if it's correct, is it what we expected?

frontend/src/components/Organizations.tsx
```diff
@@ -0,0 +1,55 @@
...
+      .orgs()
+      .then(resp => {
+        this.setState({ 
+          done: true,
+          orgs: resp,
+         });
+      })
+      .catch(err => {
+        console.error(err);
+      });
```
smacker: let's notify user about error. Now it continues showing loader.

frontend/src/api.ts
```diff
@@ -80,3 +80,11 @@ interface MeResponse {
 export function me(): Promise<MeResponse> {
   return apiCall<MeResponse>('/api/me');
 }
+
+type OrgsResponse = string[];
```
smacker: let's use correct type from the beginning. We will need id or type or image or something else anyway. Let's return it as organization object already. (with only field name)

carlosms:
> From the current implementation, as I understand, if a user chose only "some" repos it will be impossible to enable new repos from web-ui. Please confirm if it is correct or incorrect.
> 
> And if it's correct, is it what we expected?

This is a valid point. I think for now it makes sense to say this is the way you install it: giving access to a full organization. And if you "hide" repositories from lookout, then it's normal that you will not see them in the UI. To me this seems reasonable if it's explained clearly in the documentation.

My proposal is to consider the "install on new orgs from web" feature an incremental step that we can consider for future iterations. Let's discuss it in slack or next meeting.

frontend/src/components/Organizations.tsx
carlosms: :+1: I didn't want to deal with UI for errors for now, but it's so simple it makes sense to include. Done in 4cbfb29.

frontend/src/api.ts
carlosms: If I understood you correctly, done in 48c9763

smacker:
Um. It's okay to improve it later. I just wanted to raise my concern because I think the current behavior is rather counter-intuitive from a user point of view. At least I am surprised that it works like that.

frontend/src/api.ts
```diff
@@ -80,3 +80,15 @@ interface MeResponse {
 export function me(): Promise<MeResponse> {
   return apiCall<MeResponse>('/api/me');
 }
+
+export type Org = {
```
smacker: let be consisted with interface/type. In most cases they are interchangeable so I don't really have any preferences here. But if you prefer type instead of an interface please update all responses in this file.

web/github.go
```diff
@@ -0,0 +1,109 @@
...
+		// Use installation transport in a new client
+		client := github.NewClient(&http.Client{Transport: itr})
+
+		if installation.Account == nil || installation.Account.Login == nil {
+			lg.RequestLog(r).Errorf("failed to get GitHub installation organization name")
+			http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+			return
+		}
+
+		org := *installation.Account.Login
```
smacker: You need to check type here. For example logins of installations on our staging are:
- meysken
- carlosms
- campoy
- bzz
- bblfsh
- src-d

only last 2 are organizations

web/github.go
```diff
@@ -0,0 +1,109 @@
...
+		mem, _, err := client.Organizations.GetOrgMembership(context.TODO(), login, org)
```
smacker: do we actually have permissions to do it? I don't think we request this permission for our app:
screenshot 2019-01-11 at 13 51 56

I also not sure if we really should request it.

frontend/src/api.ts
smacker: yes. That's what I meant. Thanks!

frontend/src/api.ts
carlosms: +1 for consistency. Done in 8de2924, please check this is the right syntax for an array interface,

web/github.go
carlosms: Good catch, I only tested with an organization installation. Done in d067be8.

Now the question is... do we want to manage User installations in the web UI? I don't see why not, but at the same time I don't want to make the scope bigger right now. I'm thinking we can limit to organizations for now, and open an issue to check later, when all the functionality is done, if everything works for User installations. WDYT?

web/github.go
carlosms: Ah, I changed the permissions in my test app and forgot to add it to the docs in this PR.

I think we can add a note in https://github.com/src-d/lookout/blob/master/docs/web.md, and open an issue to add it in https://github.com/src-d/lookout/blob/master/docs/configuration.md#authentication-as-a-github-app when we really include it in a release.

What do you mean by this?
> I also not sure if we really should request it.

Do we have another alternative to see if the user in an admin in the organization?

web/github.go
smacker: Yes. Most probably there will be not many changes to support user installations too. It's okay to support it from the start or add later.

web/github.go
smacker: The alternative is to ask `read:org` scope when a user authorizes in web UI. 

I'm not sure which way is better (not from the code point of view but from user one). Most probably if web UI is optional then this extra permission request better be optional too when you actually use web UI.

I'm good if for now we keep asking this permission in the App. But please consider the alternative.

web/github.go
carlosms: I created an issue for this: #467.

web/github.go
carlosms: Off the top of my head, what you propose would be done once, when you log-in, right? And then we would need to store the list of orgs somewhere. Since we said we didn't want to have user management on the DB for now, it would go in the JWT I guess.

But then you have a list that is not up to date, and we would have to compromise:
- assume you need ask the user to manually log out and log in to see new organizations
- have a low expiration date in the JWT, which could be annoying for users asking to log in too often

It's not so clear to me how well it would work, and it does not look a trivial change. My proposal is to continue with the mechanism in this PR, as it was defined in the design doc, and discuss your proposal later. If we see it has advantages it can be switched later.
For now I added the new permission requirement in fe6cfa4.

web/github.go
```diff
@@ -0,0 +1,114 @@
...
+			return
+		}
+
+		// a GitHub App can also be installed in a User account
+		if *installation.Account.Type != "Organization" {
+			continue
+		}
+
+		org := *installation.Account.Login
+		mem, _, err := client.Organizations.GetOrgMembership(context.TODO(), login, org)
```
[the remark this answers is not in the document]
carlosms: Done in df7c892

web/github.go
smacker: we have a valid token oauth in JWT already. We use it in `/me` endpoint that is called everytime you open UI. You can call orgs endpoint with the same endpoint. No need to cache and it will expired at the same time as `/me`.

carlosms: :thinking: somehow I missed that and I was convinced we were only storing the user id; now it's an easy change. Still, I would prefer to have this PR merged to unblock next tasks, the api result should not change if we switch one method for the other later.

smacker: yep. Let's merge it. And change it later if we want. At least you are aware of the alternative now :)
It's also possible that using oauth token would make simpler to enable not-enabled repos for example. Just keep it in mind.

web/auth.go
```diff
@@ -268,8 +270,11 @@ var extractor = &request.PostExtractionFilter{
 
 type userContext int
 
-const userIDKey userContext = 1
-const userOAuthToken userContext = 2
+const (
+	userIDKey      userContext = 1
```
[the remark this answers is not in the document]
carlosms: I just extended the previous constant definitions. But it's a good place for iota, done in 40f26d2.

PR #457, opened 2019-01-09
WIP Tangle Implicit Rule FIXED

dpordomingo:
solves the bug found by https://github.com/src-d/lookout/pull/446

If the `Makefile.web` is newer than `Makefilep`, and `make` runs its _implicit rules_, it will fail;
thanks to 236ca4e the _implicit rule_ about `Makefile.web` is not run.


dpordomingo:
The result of the passing build can be seen here https://travis-ci.org/src-d/lookout/builds/477876579

PR #458, opened 2019-01-09
Dedup comments returned by analyzers

### LINKED ISSUE #263: Duplicated comments when analyzer returns dupes
If an analyzer returns twice the same comment, then they're posted twice into the PR.
(same file::line:analyzer::message)

I discovered it working with [style-analyzer/typos_checker](https://github.com/irinakhismatullina/style-analyzer/tree/feature/typos-analyzer/lookout/style/typos_checker) and posting comments for go, because the `bblfsh/go-driver` returns the same suggestion twice https://github.com/bblfsh/go-driver/issues/28

se7entyse7en:
Closes #263. This address scenario a) in the mentioned issue.

### DIFF
--- a/analysis.go
+++ b/analysis.go
@@ -2,6 +2,7 @@ package lookout
 
 import (
 	"google.golang.org/grpc"
+	log "gopkg.in/src-d/go-log.v1"
 	"gopkg.in/src-d/lookout-sdk.v0/pb"
 )
 
@@ -88,4 +89,48 @@ func (g AnalyzerCommentsGroups) Count() int {
 	}
 
 	return count
+}
+
+// Dedup filters duplicated comments
+func (g AnalyzerCommentsGroups) Dedup() AnalyzerCommentsGroups {
+	var result []AnalyzerComments
+	type Key struct {
+		File string
+		Line int32
+		Text string
+	}
+
+	for _, group := range g {
+		dupComments := 0
+		analyzerName := group.Config.Name
+		var newComments []*Comment
+		uniqueCommentsMap := map[Key]*Comment{}
+		for _, comment := range group.Comments {
+			key := Key{
+				File: comment.File,
+				Line: comment.Line,
+				Text: comment.Text,
+			}
+			if _, ok := uniqueCommentsMap[key]; ok {
+				dupComments++
+				continue
+			}
+
+			uniqueCommentsMap[key] = comment
+			newComments = append(newComments, comment)
+		}
+
+		if len(newComments) > 0 {
+			result = append(result, AnalyzerComments{
+				Config:   group.Config,
+				Comments: newComments,
+			})
+		}
+
+		if dupComments > 0 {
+			log.Warningf("analyzer %s generated %d duplicated comments", analyzerName, dupComments)
+		}
+	}
+
+	return result
 }
--- a/cmd/server-test/multi_analyzer_test.go
+++ b/cmd/server-test/multi_analyzer_test.go
@@ -38,15 +38,24 @@ func (suite *MultiDummyIntegrationSuite) TearDownTest() {
 func (suite *MultiDummyIntegrationSuite) TestSuccessReview() {
 	suite.sendEvent(successJSON)
 
-	suite.GrepAll(suite.r, []string{
+	str := suite.GrepAll(suite.r, []string{
 		"processing pull request",
 		`msg="posting analysis" app=lookoutd comments=4`,
-		`{"analyzer-name":"Dummy1","file":"another.go","line":3,"text":"This line exceeded`,
-		`{"analyzer-name":"Dummy1","file":"another.go","line":3,"text":"This line exceeded 120 chars."}`,
-		`{"analyzer-name":"Dummy2","file":"another.go","line":3,"text":"This line exceeded`,
-		`{"analyzer-name":"Dummy2","file":"another.go","line":3,"text":"This line exceeded 120 chars."}`,
 		`status=success`,
 	})
+
+	dummy1First := `{"analyzer-name":"Dummy1","file":"another\.go","text":"The file has increased in 4 lines\."}.*` +
+		`{"analyzer-name":"Dummy1","file":"another\.go","line":3,"text":"This line exceeded 120 chars\."}.*` +
+		`{"analyzer-name":"Dummy2","file":"another\.go","text":"The file has increased in 4 lines\."}.*` +
+		`{"analyzer-name":"Dummy2","file":"another\.go","line":3,"text":"This line exceeded 120 chars\."}`
+	dummy2First := `{"analyzer-name":"Dummy2","file":"another\.go","text":"The file has increased in 4 lines\."}.*` +
+		`{"analyzer-name":"Dummy2","file":"another\.go","line":3,"text":"This line exceeded 120 chars\."}.*` +
+		`{"analyzer-name":"Dummy1","file":"another\.go","text":"The file has increased in 4 lines\."}.*` +
+		`{"analyzer-name":"Dummy1","file":"another\.go","line":3,"text":"This line exceeded 120 chars\."}`
+	expr := `(?s)(` + dummy1First + `|` + dummy2First + `)`
+
+	found := suite.EgrepWholeFromString(str, expr)
+	suite.Require().True(found)
 }
 
 func TestMultiDummyIntegrationSuite(t *testing.T) {
--- a/server/server.go
+++ b/server/server.go
@@ -384,7 +384,7 @@ func mergeMaps(global, local map[string]interface{}) map[string]interface{} {
 }
 
 func (s *Server) post(ctx context.Context, e lookout.Event, comments lookout.AnalyzerCommentsGroups, safe bool) error {
-	comments, err := comments.Filter(func(c *lookout.Comment) (bool, error) {
+	comments, err := comments.Dedup().Filter(func(c *lookout.Comment) (bool, error) {
 		yes, err := s.commentOp.Posted(ctx, e, c)
 		if err != nil {
 			ctxlog.Get(ctx).Errorf(err, "comment posted check failed")
--- a/util/cmdtest/grep.go
+++ b/util/cmdtest/grep.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"regexp"
 	"strings"
 	"time"
 )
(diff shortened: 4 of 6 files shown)


server/server.go
```diff
@@ -164,7 +164,7 @@ func (s *Server) HandleReview(ctx context.Context, e *lookout.ReviewEvent, safeP
 		if err != nil {
 			return nil, err
 		}
-		return resp.Comments, nil
+		return s.dedupComments(resp.Comments), nil
```
smacker: maybe it's better to deduplicate comments from all analyzers?

server/server.go
```diff
@@ -433,6 +433,32 @@ func (s *Server) status(ctx context.Context, e lookout.Event, st lookout.Analysi
 	}
 }
 
+func (s *Server) dedupComments(comments []*lookout.Comment) []*lookout.Comment {
```
smacker: I would better make it a method of `lookout.AnalyzerCommentsGroups` same as `Filter`.

smacker:
one more thing. I would add a warning if we deduplicate analyzer comments. So analyzer developer at least knows that something is wrong during development. Currently, an analyzer can send 100000000 dups and we post only one silently.

server/server.go
se7entyse7en: I don't know. Anyway I miss some context, is the final user aware of which analyzer generated the comment? Or do they appear under the same lookout bot?

Because in general I'd say that there's no difference for the same comment being generated by differente analyzer, hence dedupping on the whole set of comments sounds good. But could be the final user interested in for example being notified about all comments of all analyzer, but addressing only those of some analyzers? In this scenario the user would want to have all comments and the deduplication should be applied at single analyzer level. Obviously this doesn't apply if everything comes out as posted by a single bot.

server/server.go
se7entyse7en: Good point! I agree 👍

server/server.go
smacker: Currently, everything is posted by the same single bot. And there is no way to distinguish which analyzer produced the comment from a user point of view except link at the footer:
screenshot 2019-01-10 at 18 07 58

I'm also not sure if we should deduplicate all the comments but it is what looks more logically correct for me.

se7entyse7en: Given that there's no way to distinguish which analyzer produced the comment I agree that deduplicating all the comments is more consistent.

se7entyse7en:
@smacker

cmd/server-test/multi_analyzer_test.go
```diff
@@ -41,8 +41,7 @@ func (suite *MultiDummyIntegrationSuite) TestSuccessReview() {
 	suite.GrepAll(suite.r, []string{
 		"processing pull request",
 		"posting analysis",
-		`{"analyzer-name":"Dummy1","file":"another.go","line":3,"text":"This line exceeded`,
-		`{"analyzer-name":"Dummy2","file":"another.go","line":3,"text":"This line exceeded`,
+		`"file":"another.go","line":3,"text":"This line exceeded`,
```
smacker: This test checks if lookout works correctly with more than 1 analyzer. Now it's not clear if the second analyzer works.

smacker:
Great job! But multi-analyzer test doesn't check anything anymore.

cmd/server-test/multi_analyzer_test.go
se7entyse7en: You're right! I'm gonna fix it, thanks! 👍

se7entyse7en:
@smacker improved testing [here](https://github.com/src-d/lookout/pull/458/commits/9b3babbd88719e694e95190889a23679f5f9074c)

util/cmdtest/grep.go
```diff
@@ -91,17 +93,127 @@ func (s *IntegrationSuite) Grep(r io.Reader, substr string) (bool, *bytes.Buffer
...
+	}
+
+	return s.scanAndFind(r, func(buf *bytes.Buffer, line string) bool {
+		return reg.MatchString(line)
+	}, "Egrep")
+}
+
+// EgrepFromString reads the lines in str until finds matching regex expression with timeout
+// return result
+func (s *IntegrationSuite) EgrepFromString(str string, expr string) bool {
```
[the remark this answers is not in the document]
se7entyse7en: They're similar but not duplicated, in my opinion it's clearer as is.

util/cmdtest/grep.go
```diff
@@ -91,17 +93,127 @@ func (s *IntegrationSuite) Grep(r io.Reader, substr string) (bool, *bytes.Buffer
...
 	}
 
 	return found, buf
 }
+
+func (s *IntegrationSuite) iterAndFind(str []string, fn func(string) bool, caller string) bool {
+	var found bool
+
+	foundch := make(chan bool, 1)
+	go func() {
```
smacker: I don't see why you need goroutine here. You already have an array of strings. You don't need to wait for anything.

smacker:
that is a complex solution :-D

Maybe just add grep for `generated 2 duplicated comments` will be enough?

se7entyse7en:
😂 The problem with that is that those `2 duplicatead comments` could be generated by the same analyzer whose comments were posted. So still no guarantee regarding the fact that both worked. And those utils could be usefull also for other use cases.

util/cmdtest/grep.go
se7entyse7en: You're right I forgot to remove it

se7entyse7en: Addressed [here](https://github.com/src-d/lookout/pull/458/commits/4e8405868224d4b0d5fbf307dd514b67462eee60)

se7entyse7en:
Thanks! Waiting for @carlosms then 😉

cmd/server-test/multi_analyzer_test.go
```diff
@@ -38,12 +38,22 @@ func (suite *MultiDummyIntegrationSuite) TearDownTest() {
...
+	expr1 := `level=warning msg="analyzer Dummy2 generated 2 duplicated comments".*` +
+		`{"analyzer-name":"Dummy1","file":"another\.go","text":"The file has increased in 4 lines\."}.*` +
+		`{"analyzer-name":"Dummy1","file":"another\.go","line":3,"text":"This line exceeded 120 chars\."}`
+	expr2 := `level=warning msg="analyzer Dummy1 generated 2 duplicated comments".*` +
+		`{"analyzer-name":"Dummy2","file":"another\.go","text":"The file has increased in 4 lines\."}.*` +
+		`{"analyzer-name":"Dummy2","file":"another\.go","line":3,"text":"This line exceeded 120 chars\."}`
+	expr := `(?s)(` + expr1 + `|` + expr2 + `)`
+
+	found := suite.EgrepWholeFromString(str, expr)
+	suite.Require().True(found)
```
carlosms: Now the test check that the server warns about the duplicated comments, but we are not testing that any comment is actually posted, right?

analysis.go
```diff
@@ -79,3 +80,46 @@ func (g AnalyzerCommentsGroups) Filter(fn CommentsFilterFn) ([]AnalyzerComments,
...
+
+		if len(newComments) > 0 {
+			result = append(result, AnalyzerComments{
+				Config:   group.Config,
+				Comments: newComments,
+			})
+		}
+
+		if dupComments > 0 {
+			log.Warningf("analyzer %s generated %d duplicated comments", group.Config.Name, dupComments)
```
dpordomingo: I don't understand de logic behind it was decided to deduplicate comments between analyzers:
- It most likely never happen,
- even if it happens, it should not be considered a dupe because,
  - dupes from different analyzers should have different feedbacks,
  - from the [testacase](https://github.com/src-d/lookout/pull/458/files#diff-ae0924ee5d8a959a1bbe8162f321185eR71), when the dupe from the third analyzer is logged as a dupe, it can not be reproduced without the global context of source{d} Lookout, so the maintainer of that analyzer wont be able to fix it, nor even reproduce the bug.

se7entyse7en: Regarding the decision of deduplicating the comments *as a whole* instead of doing it *by analyzer* is discussed [here](https://github.com/src-d/lookout/pull/458#discussion_r246689092).

> from the testacase, when the dupe from the third analyzer is logged as a dupe, it can not be reproduced without the global context of source{d} Lookout, so the maintainer of that analyzer wont be able to fix it, nor even reproduce the bug.

I don't know if understood what you mean. Actually an analyzer producing a comment which is the same of another analyzer is not a bug given that they are supposed to work independently. So it's not a bug that should be addressed by the analyzer's maintainer. With this PR they're filtered out to improve the user experience. Anyway your comment made me think about these two scenarios:
1. a duplicated comment is a duplicate of *another* analyzer,
2. a duplicated comment is a duplicate of the *same* analyzer.

As said before scenario 1. is not a bug of the analyzer, on the other hand scenario 2. is (in my opinion as I don't see the purpose of doing it) a bug of the analyzer.

This warning log comment is actually printed independently from being a duplicate of *another* or the *same* analyzer. As you correctly pointed out, in case of a bug in the analyzer, its maintainer should be notified. Hence I think that we could improved this warning by splitting it in order to identify whether the duplicate is originated from scenario 1 or scenario 2.

I would use the followings:
* for scenario 1:
```
log.Infof("analyzer %s generated %d comments already generated by others, skipping.", group.Config.Name, dupComments)
```
* for scenario 2:
```
log.Warningf("analyzer %s generated %d duplicated comments", group.Config.Name, dupComments)
```

cmd/server-test/multi_analyzer_test.go
carlosms: My bad, I thought this was all a warning message
```go
	expr1 := `level=warning msg="analyzer Dummy2 generated 2 duplicated comments".*` +
		`{"analyzer-name":"Dummy1","file":"another\.go","text":"The file has increased in 4 lines\."}.*` +
		`{"analyzer-name":"Dummy1","file":"another\.go","line":3,"text":"This line exceeded 120 chars\."}`

```
but it's actually the warning of duplicated and the 2 comments posted :+1:

analysis.go
se7entyse7en: Splitted the duplicated comments counting and corresponding log messages [here](https://github.com/src-d/lookout/pull/458/commits/ae9066a13abea6891c0d4f0a8a974f24091ff1c4).

dpordomingo: I agree on using different messages if all of them are deduplicated.
But I don't agree on deduplicating comments that other analyzers already posted.
When @smacker [said](https://github.com/src-d/lookout/pull/458#discussion_r246689092) that the only way to know which analyzer returned a comment is the link in the footer, for me that's a valid way to distinguish them, so the user can use that link to notify each analyzer maintainer about a bug on their analyzer.

example:
consider those two comments both in the same line `113` at same `example.go` file:
-  `This line is not properly formatted; any feedback?, open issue [here](https://github.com/src-d/style-analyzer/issues)`
- `This line is not properly formatted; any feedback?, open issue [here](https://github.com/src-d/lookout-gometalint-analyzer/issues)`

The user can report an issue in both analyzers if both comments are wrong, only if different comments from different analyzers are not deduplicated.

A different problem is that we don't differentiate both comments in the messages that source{d} Lookout post on behalf the analyzers, so the user may be confused; but it's a UX problem that should be considered in a different issue, instead of dropping the second comment because the current UI is not clear enough.

se7entyse7en: I agree with you that there's also a UX problem and that it should be addressed. In my opinion the best thing is to _remove_ the duplicates of the same analyzer and to _merge_ the duplicates of different analyzers. By doing this we can then show something like
```This line is not properly formatted; any feedback?, open issue [here](https://analyzers.sourced.tech/feedback/?analyzers=style-analyzer,lookout-gometalint-analyzer&prurl=<url of the PR>&...other useful informations...)```

I would show a simple form similar to github's one to open an issue, that on submit automatically open the issue on each analyzer's repo. I would personally be annoyed by copy-pasting the same issue for multiple analyzers.

Regarding what we want to do **now** before thinking about the UX, I don't have a strong opinion on which approach is better. I slightly prefer avoid duplicating comments globally at the cost of being able to give a feedback only to one analyzer. It depends on what we want to give more importance to (being annoyed by duplicated comments or let the use give a feedback), and also on when (and if) we want to address the UX issue.

I'd also ask @carlosms and @smacker opinion.

Comment (dpordomingo) [reply analysis.go]: final decition comes from last TFM [applications/2019-01-15#Line80](https://github.com/src-d/minutes/pull/430/files#diff-76b4489f9540c5a62d9393ef188c40bbR80)
>_To keep the comments (from different analyzers) separated. Add the name of the analyzer in a different PR._

**Comment by se7entyse7en** [comment]
The discussion has been addressed [here](https://github.com/src-d/lookout/pull/458/commits/9882a541c8285871f277ff7f6423dfe9c645918f)

File: analysis.go [hunk 8]
```diff
@@ -79,3 +80,49 @@ func (g AnalyzerCommentsGroups) Filter(fn CommentsFilterFn) ([]AnalyzerComments,
...
+func (g AnalyzerCommentsGroups) Dedup() AnalyzerCommentsGroups {
+	var result []AnalyzerComments
+	type Key struct {
+		File         string
+		Line         int32
+		Text         string
+		AnalyzerName string
+	}
+
+	uniqueCommentsMap := map[Key]*Comment{}
```
dpordomingo: Can we assume @carlosms that each group contains the comments for each different analyzer?
If so, the index map `uniqueCommentsMap` could be moved into the first loop, and the `Key.AnalyzerName` property would be unnecessary.

se7entyse7en: you're right @dpordomingo, there's no more need to initialize it outside the outer loop 👍

se7entyse7en: Addressed [here](https://github.com/src-d/lookout/pull/458/commits/78381b43a61fe204a144b424244a625fa7c25ac3)

se7entyse7en:
Gonna rebase and merge then 👍

PR #460, opened 2019-01-10
WIP Tangle Implicit Rule REPRODUCED

dpordomingo:
reproduce the bug found by https://github.com/src-d/lookout/pull/446

If the `Makefile.web` is newer than `Makefilep`, and `make` runs its _implicit rules_, it will fail.


dpordomingo:
The result of the failing build can be seen here https://travis-ci.org/src-d/lookout/builds/477825224

PR #473, opened 2019-01-16
Fix callback test

dpordomingo:
It does not matter the URL wrote in the Request because the callback handler is called directly and it does not check the URL called, but it is still needed to point to the right URL when testing the callback.


dpordomingo:
Let me know if I should open an Issue to ensure that the tests should validate that the called url is the one that should be called.

PR #474, opened 2019-01-16
Let web-start work without assets

dpordomingo:
The docs suggested to call `make -f Makefile.web web-start` to run separately the frontend and the backend, but it can not be used this way if the assets have not been generated yet because `with_static` go tag is set by default every time.
```shell
$ make -f Makefile.web web-start
LOG_LEVEL=DEBUG go run -tags "with_static" cmd/lookoutd/*.go web
web/static_files.go:13:2: cannot find package "github.com/src-d/lookout/web/assets" in any of:
	/projects/src/github.com/src-d/lookout/vendor/github.com/src-d/lookout/web/assets (vendor tree)
	/usr/local/go/src/github.com/src-d/lookout/web/assets (from $GOROOT)
	/projects/src/github.com/src-d/lookout/web/assets (from $GOPATH)
Makefile.web:67: recipe for target 'web-start' failed
make: *** [web-start] Error 1
```

Now it can be called:
- `WITHOUT_STATIC=true make -f Makefile.web web-start`
- `make -f Makefile.web web-serve`

I also fixed:
- docs,
- `web-serve` make target

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -104,7 +104,7 @@ In case you want to locally run the web interface from sources, you can choose o
     ```shell
     $ make -f Makefile.web web-serve
     ```
-    And navigate to [http://127.0.0.0:8080](http://127.0.0.0:8080)
+    And navigate to [http://127.0.0.1:8080](http://127.0.0.1:8080)
 - using `create-react-app` dev server, with live reload for web assets changes, running in separated terminals the backend: 
     ```shell
     $ make -f Makefile.web web-start
@@ -114,7 +114,7 @@ In case you want to locally run the web interface from sources, you can choose o
     $ make -f Makefile.web web-dependencies # if you didn't do it yet
     $ yarn --cwd frontend start
     ```
-    Configure the GitHub App authorization callback URL to `http://127.0.0.0:3000/callback`, and navigate to [http://127.0.0.0:3000](http://127.0.0.0:3000)
+    Configure the GitHub App authorization callback URL to `http://127.0.0.1:3000/callback`, and navigate to [http://127.0.0.1:3000](http://127.0.0.1:3000)
 
 ### Testing
 
--- a/docs/web.md
+++ b/docs/web.md
@@ -23,7 +23,7 @@ Open [http://127.0.0.1:8080](http://127.0.0.1:8080) in your browser.
 
 The web interface requires the usage of a Github App as authorization method, and requires GitHub App OAuth credentials.
 
-The GitHub App authorization callback must be processed by the `/callback` frontend route, so configure it when creating the GitHub App (the default value would be) `http://127.0.0.0:8080/callback`.
+The GitHub App authorization callback must be processed by the `/callback` frontend route, so configure it to your hostname when creating the GitHub App. The default value would be `http://127.0.0.1:8080/callback`.
 
 Please follow the instructions on how to get the GitHub App credentials in the [main configuration guide](configuration.md#authentication-as-a-github-app), and set them in `config.yaml` as follows:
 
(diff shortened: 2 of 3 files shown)


docs/web.md
```diff
@@ -23,7 +23,7 @@ Open [http://127.0.0.1:8080](http://127.0.0.1:8080) in your browser.
 
 The web interface requires the usage of a Github App as authorization method, and requires GitHub App OAuth credentials.
 
-The GitHub App authorization callback must be processed by the `/callback` frontend route, so configure it when creating the GitHub App (the default value would be) `http://127.0.0.0:8080/callback`.
+The GitHub App authorization callback must be processed by the `/callback` frontend route, so configure it when creating the GitHub App (the default value would be) `http://127.0.0.1:8080/callback`.
```
carlosms: Let's change also the phrase, it sounds a bit off.
```suggestion
The GitHub App authorization callback must be processed by the `/callback` frontend route, so configure it to your hostname when creating the GitHub App. The default value would be `http://127.0.0.1:8080/callback`.
```

carlosms:
I would prefer to find a way to avoid the env var. If we were reusing the same target for serve and backend-only; e.g. `make -f Makefile.web web-serve` & `WITHOUT_STATIC=true make -f Makefile.web web-serve`; then it would make sense.
But if we already have 2 targets, the makefile should take care of the requirement, in my opinion. I think it would be more clear to either:
- Remove `web-start` and just put in the docs `go run cmd/lookoutd/*.go web`
- Change `web-start` to not require the env var, something like
```
web-start:
	$(GO_RUN_ENV) $(GOCMD) run cmd/lookoutd/*.go web
```

what do you think?

dpordomingo:
I initially considered both of them, but then I realized that doing so we'll miss already set `GO_RUN_ARGS`; is it ok? I'm not sure.

An alternative could be something like:
```shell
GO_RUN_ARGS_NO_STATIC := $(GO_RUN_ARGS)
GO_RUN_ARGS += -tags "$(WITH_STATIC_TAG)"

GOCMD = go
GORUN = $(GOCMD) run $(GO_RUN_ARGS)

...

web-start:
	$(GO_RUN_ENV) $(GOCMD) run $(GO_RUN_ARGS_NO_STATIC) cmd/lookoutd/*.go web

web-serve: | web-dependencies web-build web-bindata
	$(GO_RUN_ENV) $(GOCMD) run $(GO_RUN_ARGS) cmd/lookoutd/*.go web
```

so calling:
- `web-start` will run the server **WITHOUT** assets,
- `web-serve` will run the server **WITH** assets.

Do you have any other idea?

docs/CONTRIBUTING.md
```diff
@@ -104,17 +104,17 @@ In case you want to locally run the web interface from sources, you can choose o
     ```shell
     $ make -f Makefile.web web-serve
     ```
-    And navigate to [http://127.0.0.0:8080](http://127.0.0.0:8080)
+    And navigate to [http://127.0.0.1:8080](http://127.0.0.1:8080)
 - using `create-react-app` dev server, with live reload for web assets changes, running in separated terminals the backend: 
     ```shell
-    $ make -f Makefile.web web-start
+    $ WITHOUT_STATIC=true make -f Makefile.web web-start
```
dpordomingo:
```suggestion
    $ WITHOUT_STATIC=true make -f Makefile.web web-starts
```

carlosms:
Yes, that looks fine. Although I would just skip `GO_RUN_ARGS_NO_STATIC` to simplify; `GO_RUN_ARGS` is not part of the common makefile ([ref](https://github.com/src-d/ci/tree/v1#custom-build-parameters)), it's not meant to be configured by the users. Up to you.

dpordomingo:
wow!
that's true...
I misunderstood the opposite because the misleading and name and how it was defined:
- using UPPERCASE, usually used for user variables,
- appending to its prev value, as it could be already defined...
New commit considering your suggestions at: f7a3abc

PR #478, opened 2019-01-17
Increases test coverage of service/git package

se7entyse7en:
`service/git/service.go`: 63.3% -> 98.2%

### DIFF
--- a/service/git/service_test.go
+++ b/service/git/service_test.go
@@ -8,9 +8,12 @@ import (
 	"github.com/src-d/lookout"
 
 	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 	fixtures "gopkg.in/src-d/go-git-fixtures.v3"
+	"gopkg.in/src-d/go-git.v4/plumbing"
 	"gopkg.in/src-d/go-git.v4/plumbing/cache"
+	"gopkg.in/src-d/go-git.v4/plumbing/object"
 	"gopkg.in/src-d/go-git.v4/plumbing/storer"
 	"gopkg.in/src-d/go-git.v4/storage/filesystem"
 )
@@ -19,10 +22,6 @@ type ServiceSuite struct {
 	suite.Suite
 	Basic  *fixtures.Fixture
 	Storer storer.Storer
-}
-
-func TestServiceSuite(t *testing.T) {
-	suite.Run(t, new(ServiceSuite))
 }
 
 func (s *ServiceSuite) SetupSuite() {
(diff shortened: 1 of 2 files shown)


service/git/service_test.go
```diff
@@ -46,147 +47,305 @@ func (s *ServiceSuite) TearDownSuite() {
...
+	expectedTypesWrapped := [2]lookout.ChangeScanner{&TreeScanner{}, &DiffTreeScanner{}}
+	testNames := [2]string{"without base - tree scanner", "with base - diff tree scanner"}
+	for i, changeReq := range changeRequests {
+		s.T().Run(testNames[i], func(t *testing.T) {
+			require := require.New(t)
+			resp, err := dr.GetChanges(context.TODO(), changeReq)
+
+			require.NoError(err)
+			require.NotNil(resp)
+			require.IsType(&lookout.FnChangeScanner{}, resp)
```
smacker: why do we need to test type of the struct that implements scanner?

service/git/service_test.go
```diff
@@ -46,147 +47,305 @@ func (s *ServiceSuite) TearDownSuite() {
...
+			require.IsType(expectedTypesWrapped[i], resp.(*lookout.FnChangeScanner).Scanner)
```
smacker: hm... Could you please explain to me what exactly this test(s) does? I see it checks types. But why do we need to test implementation details (types) instead of results in response?

se7entyse7en: The idea is that okay that `GetChanges` returns a `lookout.ChangeScanner` which is actually an interface, but depending on the scenario we return different types of `lookout.ChangeScanner`. The idea of this test was to ensure that depending on the scenario we're returning the correct `lookout.ChangeScanner`. I don't know how to test that we're correctly returning the expected scanner except for running it, but there other test cases testing the correct behaviour of each scanner.

> instead of results in response?

What do you mean?

smacker: > What do you mean?
I mean test `TestGetChangesWithPattern` would check that it returns results only that match pattern. 
And `TestTreeChangesDeleteFile` returns or does not return changes when a file is deleted.

I'm not sure it's a good idea to check which struct implements an interface. For example, if I implement pattern scanner as different structure (without using `FnChangeScanner`) but it would still return only files matching a pattern this test will fail. But all logic is correct and works as expected. It makes me question the test.

@carlosms wdyt?

dpordomingo: What we should ideally test is the list of `lookout.Change` returned by the `lookout.ChangeScanner` returned by `git.Service::GetChanges` because that's the contract signed by `git.Service::GetChanges`. The process of how the list of `lookout.Change` is built, is an implementation detail of `git.Service::GetChanges` that imo should not be tested.

smacker: I would prefer to test here too. For example, scanner works fine but we don't pass options to it from service correctly so it returns wrong results.

+1 for refactoring of tests. Repeating checks is error-prone. Service&scanner should return the same result for simple cases when service uses only one scanner.

se7entyse7en: Ok, I got the point. And yes it's better to refactor the tests and also test here the output of the returned scanner. Initially I hadn't given that we already have `scanner_test.go` as @carlosms pointed. But it's true what @smacker says:

> scanner works fine but we don't pass options to it from service correctly so it returns wrong results.

smacker: So final tests will be:
- `GetChangesInvalidReference` - 1 case
- `GetChangesInvalidTree` - 1 case
- `GetChangesNoBase` - 1 case
- `GetChangesWithPattern` - 3 cases inside for IncludePattern/ExcludePattern/both
- `GetChangesWithContent` - 1 case
- `GetChangesWithPatternAndContent`  - 1 case, tests that scanners don't conflict

that should cover all logic of `GetChanges` methods of the service.

smacker: another approach for the last 3 cases would be: you check not the type as right now but the instance on scanners and check that the fields passed in the service a correctly set in scanners. Then we have tests for scanners already.
But I really think just re-running scanners would be easier to implement and less fragile.

service/git/service_test.go
```diff
@@ -46,147 +45,415 @@ func (s *ServiceSuite) TearDownSuite() {
...
+		&lookout.FilesRequest{Revision: s.buildRefPointer("", "referenceName", headHash)},
+	}
+
+	testNames := [3]string{
+		"empty revision",
+		"empty string in hash",
+		"empty string in internal repo url",
+	}
+
+	for i, filesReq := range filesRequests {
```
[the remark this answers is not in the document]
se7entyse7en: Much cleaner to keep them separated.

se7entyse7en:
This is updated.

Issue #401, opened 2018-12-07
Set a default config file for each organization


carlosms:
Depends on #400.
This issue is to be able to define a default `lookout.yml` for all the repositories (current and future) of an organization.

To do this we'll need:
- [x] New table to store the config file contents for each organization #483
- [x] Endpoint to perform the update. A user should only be able to change the org config file it they are an admin of the organization. #484
- [x] UI should be simple. A textarea where you can paste the yaml contents will be enough for now. #484
- [x] The default config will be merged with the `lookout.yml` commited in the repository. In this case the one in the repository has priority over the organization default. #493

Issue #316, opened 2018-10-09
Add constraints and indexes to DB 


smacker:
The code assumes that some rows are uniq. We need to enforce it on DB level.
We also need to add indexes to the columns we use in queries. For example, we query `ReviewTarget` by `Provider` and `InternalID`.

PR #483, opened 2019-01-18
New DB models for organization default config

carlosms:
Part of #401.

Instead of making a numeric primary key from the organization ID, I thought it made sense to make it more generic to accommodate future providers. With (Provider, InternalID) as primary key it will be easier to add GitLab Groups, Bitbucket Teams, etc.
InternalID is a string just in case some provider uses strings as uuid (seems to be the case for [bitbucket](https://developer.atlassian.com/bitbucket/api/2/reference/resource/teams)).

I made this PR from a local branch to make it easier to do sequential PRs.


store/migrations/1547744983_organizations.up.sql
```diff
@@ -0,0 +1,11 @@
+BEGIN;
+
+CREATE TABLE organization (
```
se7entyse7en: AFAIU the couple `(provider, internal_id)` should be unique. In that case we could enforce it at db level by adding the corresponding constraint.

store/migrations/1547744983_organizations.up.sql
```diff
@@ -0,0 +1,11 @@
...
+	id uuid NOT NULL PRIMARY KEY,
+	provider text NOT NULL,
```
se7entyse7en: Given that we're [querying by  `provider` and `internal_id`](https://github.com/src-d/lookout/pull/483/files#diff-94c7f4f10a53dfdc07f94f92324cde86R263) we could add an index on both columns in order to speed up the queries.

se7entyse7en:
I don't know [kallax](https://github.com/src-d/go-kallax), so maybe I missed something 😅.

store/migrations/1547744983_organizations.up.sql
dpordomingo: I'm not sure, but from `go-kallax/README.md`:
>_Specifies the column has a unique constraint._
the `unique` constraint could be at column level, so composite uniqueness could be not supported by the ORM.

store/models/models.go
```diff
@@ -97,3 +97,24 @@ type Comment struct {
 func newComment(r *ReviewEvent, c *lookout.Comment) *Comment {
 	return &Comment{ID: kallax.NewULID(), ReviewEvent: r, Comment: *c}
 }
+
+// Organization is a persisted model for an Organization (e.g. a GitHub App
+// installation). It contains settings for a group of repositories.
+// The primary key should be (Provider,InternalID), but kallax does not support
+// composite or string keys
+type Organization struct {
```
dpordomingo: If `go-kallax` does not support unique keys for more than 1 column, I'd also say so in the docs

store/db.go
```diff
@@ -246,3 +246,53 @@ func (o *DBCommentOperator) posted(ctx context.Context, e *lookout.ReviewEvent,
...
+// Config returns the stored config for the given (provider, orgID). If there
+// are no records in the DB, it returns "" without error.
+func (o *DBOrganizationOperator) Config(ctx context.Context, provider string, orgID string) (string, error) {
+	m, err := o.getOrganization(ctx, provider, orgID)
+	if err != nil && err != kallax.ErrNotFound {
+		return "", err
+	}
+
+	if err == kallax.ErrNotFound {
+		return "", nil
```
dpordomingo: Why do we want to return an empty config if there is no such Organization? Why is that better than returning a not found, and let the caller to decide?
How can the caller know if the stored config is empty, or there is no Org?

store/models/models.go
```diff
@@ -97,3 +97,24 @@ type Comment struct {
...
+	kallax.Model `pk:"id"`
+	ID           kallax.ULID
+	Provider     string
+	InternalID   string
+	Config       string
```
dpordomingo: What are we storing in `Config`? It can be whatever thing depending on the Provider?

store/migrations/1547744983_organizations.up.sql
se7entyse7en: Ok, @smacker's issue addresses both my concerns.

store/models/models.go
dpordomingo: Oh... I see... it stores the RAW version of the `.lookout.yml` for all repos under that org.
>_The default config will be merged with the `.lookout.yml` commited in the repository. In this case the one in the repository has priority over the organization default._
(from https://github.com/src-d/lookout/issues/401 )

store/migrations/1547744983_organizations.up.sql
se7entyse7en: I suppose that this sql script is executed in the DB as-is, so even if it's not supported by the ORM at least we can prevent "wrong data" being stored. But I don't know how the ORM will fail in case of a write that doesn't satisfy the constraint.

store/migrations/1547744983_organizations.up.sql
carlosms: Answering both conversations here, since they are related.

You are right, we should enforce the uniqueness it at the DB level, and add indexes. I simply left the automatically generated migration file, but now that you mention it I'm not sure if we should add the missing details here in this PR, or in #316 with the other missing changes.

I will take a look and if it's straightforward (I guess it should) I'll push it here.

store/migrations/1547744983_organizations.up.sql
dpordomingo: I guess as any other write error?

se7entyse7en: I don't know kallax so it may be just a generic write error as you said or it may crash in unpredictable ways (I'm thinking for example about nil pointer exception that won't let the user undestand what actually happened).

store/db.go
```diff
@@ -246,3 +246,53 @@ func (o *DBCommentOperator) posted(ctx context.Context, e *lookout.ReviewEvent,
...
+
+	_, err = o.organizationStore.Save(m)
+	return err
+}
+
+// Config returns the stored config for the given (provider, orgID). If there
+// are no records in the DB, it returns "" without error.
+func (o *DBOrganizationOperator) Config(ctx context.Context, provider string, orgID string) (string, error) {
+	m, err := o.getOrganization(ctx, provider, orgID)
+	if err != nil && err != kallax.ErrNotFound {
```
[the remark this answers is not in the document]
carlosms: Much better, done in cfa84a1

store/models/models.go
carlosms: ok, done in cf97909

store/db.go
carlosms: The callers will be:
- the web backend to report the current default value
- the `Server.HandleReview`, to merge the default with the local config

My reasoning here is that I wanted a "helper" to get the the default config for a given organization, rather than a method to retrieve the raw content from the DB. They don't care if the organization has an entry in the DB, if the entry has an empty string stored, or if there is no entry; they just want the string content.

With that in mind, this method works like a `map`,. If the organization does not have a config set, you get the zero value, `""`. I wasn't sure if this "helper" behaviour should go in `db.go`, or if it's too "high level" and should go as a wrapper somewhere else. WDYT?

store/migrations/1547744983_organizations.up.sql
dpordomingo: I really hope it handle write errors

store/db.go
dpordomingo: I see that logic out of the DB package, but creating that other helper somewhere only to wrapp the functionality `if err == NotFound return ""`... Let's keep it here if you prefer so.

store/migrations/1547744983_organizations.up.sql
smacker: +1 for constraint

store/db.go
smacker: @dpordomingo keep in mind this package / struct isn't direct mapping to DB. We have ORM (kallax for that). It implements [repository pattern](https://docs.microsoft.com/en-us/dotnet/standard/microservices-architecture/microservice-ddd-cqrs-patterns/infrastructure-persistence-layer-design#the-repository-pattern). I used name "operator" instead of "repository" to avoid confusion with git repositories (as it's more important concept for lookout)

carlosms:
Unique index added in 8d9f230, please take another look.

store/db.go
dpordomingo: /aside;
I think that since it's a repository, it should focus only on accessing the data, instead of returning custom data to fit some client logic. 
>_Repositories are classes or components that encapsulate the logic required to access data sources._

But as I said, I'm ok if we keep it as it is, and avoid creating more wrappers somewhere.

carlosms:
This branch now contains the commits from #484. Squashed and rebased on master.

carlosms:
This branch was included in the merge of #493.

PR #484, opened 2019-01-18
Add default organization config endpoints and frontend

carlosms:
Part of #401, depends on #483.

My first idea was to use the organization numeric ID as the URL param for the API, but [the GitHub api](https://developer.github.com/v3/apps/#find-organization-installation) uses the org name. It certainly makes for nicer routes in the frontend.

I updated go-github to be able to use `FindOrganizationInstallation`.

I've made the PR from local branch to local branch to make it easier to review incremental PRs, once accepted everything will be merged to master.


se7entyse7en:
I just took a look, but I need to refresh my js and react knowledge 😅.

web/github.go
```diff
@@ -46,8 +63,52 @@ func (g *GitHub) installations(ctx context.Context) ([]*github.Installation, err
...
+			return nil, errNotFound
+		}
+
+		return nil, err
+	}
+
+	return installation, nil
+}
+
+func (g *GitHub) admin(w http.ResponseWriter, r *http.Request, installation *github.Installation, login string) (bool, error) {
```
smacker: why do you need `http.ResponseWriter` and  `http.Request` here?
The method just checks if this login is admin.

web/github.go
```diff
@@ -70,45 +131,153 @@ func (g *GitHub) Orgs(w http.ResponseWriter, r *http.Request) {
...
+				ID:   installation.GetAccount().GetID(),
+				Name: installation.GetAccount().GetLogin(),
+			})
 		}
 	}
 
 	successJSON(w, r, orgs)
 }
+
+// TODO (@carlosms) check if this makes more sense as a middleware
```
smacker: imo it makes sense to rewrite as a middleware. It should also simplify the code.

web/github.go
```diff
@@ -70,45 +131,153 @@ func (g *GitHub) Orgs(w http.ResponseWriter, r *http.Request) {
...
+
+// Org writes in the response the individual organization requested by the URL
+// parameter "orgName", only if the user is an admin
+func (g *GitHub) Org(w http.ResponseWriter, r *http.Request) {
+	installation, err := g.orgInstallation(w, r)
+	if err != nil {
+		return
+	}
+
+	idStr := strconv.FormatInt(installation.GetAccount().GetID(), 10)
```
smacker: it's a string, isn't it? https://github.com/src-d/lookout/pull/483/files#diff-d87db3c51b2eaa55b7443de6a8e0f17cR33

web/github.go
```diff
@@ -70,45 +131,153 @@ func (g *GitHub) Orgs(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, fmt.Sprintf("Bad Request. Body is not a valid JSON: %s", err), http.StatusBadRequest)
+		return
+	}
+
+	installation, err := g.orgInstallation(w, r)
+	if err != nil {
+		return
+	}
+
+	// TODO: check config is valid yaml
```
smacker: let's fix this TODO before merge. UI is just a text field. It's very easy to send incorrect input.

frontend/src/components/Organization.tsx
```diff
@@ -0,0 +1,115 @@
...
+        });
+      });
+  }
+
+  public render() {
+    if (!this.state.done) {
+      return <Loader />;
+    }
+
+    const org = this.state.org;
```
smacker: don't we have linter error here?

frontend/src/components/Organization.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+    if (this.state.errors.length > 0 || org === undefined) {
```
smacker: `Errors` component expects non-empty list of error. But this check allows empty one.

frontend/src/components/Organization.tsx
```diff
@@ -0,0 +1,115 @@
...
+    );
+  }
+
+  private handleConfigChange(event: React.ChangeEvent<HTMLTextAreaElement>) {
+    this.setState({ config: event.target.value });
+  }
+
+  private handleConfigSave() {
+    if (this.state.org === undefined) {
+      // tslint:disable-next-line
```
smacker: can be removed after rebase on master

frontend/src/components/Organization.tsx
```diff
@@ -0,0 +1,115 @@
...
+      console.error('handleConfigSave called with undefined state.org');
```
smacker: actually, it looks like we should throw an exception here. And later we can catch it using [error boundaries](https://reactjs.org/docs/error-boundaries.html) if we want instead of a silent broken button.

frontend/src/components/Organization.tsx
carlosms: OK I'll remove it before merging. For now I'll leave the commit history as it is to not break the existing review comments.

web/github.go
carlosms: I refactored some things and didn't change the arguments. Done in 60b6bc1.

web/github.go
carlosms: Done in 928e9b1.

frontend/src/components/Organization.tsx
carlosms: No. I changed in 2f85ec7, I don't mind one way or the other.

frontend/src/components/Organization.tsx
carlosms: _In theory_ we should not have 0 errors and undefined org. But we can break this easily, done 2f85ec7.

frontend/src/components/Organization.tsx
carlosms: It seems [error boundaries do not apply in handlers](https://reactjs.org/docs/error-boundaries.html#how-about-event-handlers), but I followed the example to render the error in 341048e.

web/github.go
carlosms: I wanted to wait a bit until we have other api endpoints (for example to edit if the organization repos are enabled/disabled by default).

web/github.go
dpordomingo: since `admin` returns a boolean, I wonder if it would be better `isAdmin`
```suggestion
func (g *GitHub) isAdmin(w http.ResponseWriter, r *http.Request, installation *github.Installation, login string) (bool, error) {
```

web/github.go
```diff
@@ -46,8 +64,52 @@ func (g *GitHub) installations(ctx context.Context) ([]*github.Installation, err
...
+			return nil, errNotFound
+		}
+
+		return nil, err
+	}
+
+	return installation, nil
+}
+
+func (g *GitHub) admin(ctx context.Context, installation *github.Installation, login string) (bool, error) {
```
dpordomingo: since `admin` returns a boolean, I wonder if it would be better `isAdmin`
```suggestion
func (g *GitHub) isAdmin(ctx context.Context, installation *github.Installation, login string) (bool, error) {
```

web/github.go
```diff
@@ -70,45 +132,159 @@ func (g *GitHub) Orgs(w http.ResponseWriter, r *http.Request) {
...
-
 		// a GitHub App can also be installed in a User account
-		if *installation.Account.Type != "Organization" {
+		if installation.GetAccount().GetType() != "Organization" {
 			continue
 		}
 
-		org := *installation.Account.Login
-		mem, _, err := client.Organizations.GetOrgMembership(r.Context(), login, org)
+		admin, err := g.admin(r.Context(), installation, login)
```
dpordomingo: since `admin` returns a boolean, I wonder if it would be better `isAdmin`
```suggestion
		admin, err := g.isAdmin(r.Context(), installation, login)
```

web/github.go
```diff
@@ -70,45 +132,159 @@ func (g *GitHub) Orgs(w http.ResponseWriter, r *http.Request) {
...
+		return nil, fmt.Errorf(http.StatusText(http.StatusInternalServerError))
+	}
+
+	// a GitHub App can also be installed in a User account
+	if installation.GetAccount().GetType() != "Organization" {
+		http.Error(w, http.StatusText(http.StatusNotFound), http.StatusNotFound)
+		return nil, fmt.Errorf(http.StatusText(http.StatusNotFound))
+	}
+
+	admin, err := g.admin(r.Context(), installation, login)
```
dpordomingo: since `admin` returns a boolean, I wonder if it would be better `isAdmin`
```suggestion
	admin, err := g.isAdmin(r.Context(), installation, login)
```

frontend/src/components/Organization.tsx
```diff
@@ -58,15 +58,19 @@ class Organization extends React.Component<OrgProps, OrgState> {
...
-    const org = this.state.org;
-
-    if (this.state.errors.length > 0 || org === undefined) {
+    if (this.state.errors.length > 0) {
       return <Errors errors={this.state.errors} />;
     }
 
+    if (this.state.org === undefined) {
+      return (
+        <Errors errors={["Internal error, 'org' should not be undefined"]} />
```
dpordomingo: are we exposing this error to the user? can he fix it?

frontend/src/components/Organization.tsx
smacker: oh, no need to change. I meant eslint would complain here and ask to write it as `const {org} = this.state`. Good to know that tslint doesn't.

frontend/src/components/Organization.tsx
carlosms: If we reach this point it's because we introduced a bug, this shouldn't happen. In this case, I'd say it's ok to render the error so the user can report it to us and describe the steps they made to reproduce it.
Otherwise the button will fail silently, without any warning as [@smacker said here](https://github.com/src-d/lookout/pull/484#discussion_r249486588), and that's also bad.

web/github.go
carlosms: ok, done in 8dbf661.

frontend/src/components/Organization.tsx
dpordomingo: I'd mention that it would be great if the error would be reported to us, but not a strong opinion; or maybe not needed to be done now/here.

dpordomingo:
t All the feedback seems to be addressed but the one needed to be done just before merging:
- [#discussion_r249485515](https://github.com/src-d/lookout/pull/484#discussion_r249485515) remove `tslint:disable-next-line` before merging

So I think we're ready to move on

carlosms:
The tslint line was already removed by 341048e.

frontend/src/components/Organization.tsx
carlosms: Let's keep that in mind for the future, at this stage I think it's enough that if something breaks at least you see why. How to report feedback can be done later when we have input from Product on the UI.

Issue #486, opened 2019-01-23
GetChanges is returning files that didn't change in the PR


dpordomingo:
Given this history:
```
* 7450de7 (origin/master) Change 'README.md' file
|
| * dd8c39e (fork/candidate) Change 'golang.go' file
| |
| /   
* 83f7e4d
| 
* 7ccf9be
```

For a pull request to merge `fork/candidate` against `origin/master`, `GetChanges` rpc is returning `golang.go` and `README.md`.

I guess it's because `DataService` is fetching changes with kind of:
```git diff 7450de7..dd8c39e```
that returns **all** differences between `7450de7` and `dd8c39e` (what is wrong)

instead of doing:
```git diff 7450de7...dd8c39e```
that returns the differences between common ancestor (`83f7e4d`) and `dd8c39e`, representing the changes made by the branch being pullrequested.

Keeping this, forces the analyzer to avoid posting comments about files that were not modified in the PR being reviewed (ie. `README.md`), because otherwise GitHub API will return an unhandled `Path is invalid` error:
```
{
  "message": "Unprocessable Entity",
  "errors": []github.Error{}[
    github.Error{
      Resource:"Review",
      Field:"",
      Code:"custom",
      Message:"Path is invalid"
    }
  ],
}
```
(described by https://github.com/src-d/lookout/issues/487)

dpordomingo:
I think GitHub API is returning an unhandled `Path is invalid` error due to https://github.com/google/go-github/issues/540

carlosms:
Let's make sure we include a a test with this scenario in https://github.com/src-d/lookout-test-fixtures

PR #488, opened 2019-01-24
Logs enhancement

### LINKED ISSUE #487: Cryptic errors in some situations
related to https://github.com/src-d/lookout/issues/486

There are some situations that cause unhandled errors like the following one. 
```shell
[2019-01-23T11:17:34.461935361+01:00] ERROR event processing failed 
app=lookoutd 
error=posting analysis failed: %!s(PANIC=runtime error: invalid memory address or nil pointer dereference) 
event-id=b44c158d09bebb40b85ce23763dd9ab179926019 
event-type=*pb.ReviewEvent 
github.pr=63 
head=refs/pull/63/head 
repo=https://github.com/dpordomingo/testing-repo.git
```
When it happens, it's difficult to know what's causing the error.
In this case, I found that it was related to https://github.com/google/go-github/issues/540 

It would be a good idea to provide more details in those logs.

dpordomingo:
fix #487

### before:
```shell
ERROR event processing failed
app=lookout
event-id=b44c158d09bebb40b85ce23763dd9ab179926019
event-type=*pb.ReviewEvent
github.pr=63
head=refs/pull/63/head
repo=https://github.com/dpordomingo/testing-repo.git
error=posting analysis failed: %!s(PANIC=runtime error: invalid memory address or nil pointer 
```
**problems:**
- no info about what caused the error
- error message is a `PANIC` without trace

### after
```shell
ERROR event processing failed

.... (same data as before changes)

error=posting analysis failed: github api error: review could not be pushed:
    POST https://api.github.com/repos/dpordomingo/testing-repo/pulls/63/reviews
    422 Unprocessable Entity
    [{Resource:Review
      Message:Path is invalid
    }]
```
**benefits:**
- error message contains where was caused the problem
  - GitHub API endpoint
  - GitHub API real error message 

these :point_up: new details would have been really useful when debugging https://github.com/src-d/lookout/issues/486

**cons:**
- the workaround done inside [`handleAPIError`](https://github.com/src-d/lookout/pull/488/files#diff-15bfbf8e99100aba04e081d16e13036bR419) could be done directly in `google/go-github` project, but it should be considered separately, so I'd handle it in [a google/go-github PR](https://github.com/dpordomingo/go-github/pull/1/files#diff-9bd6c2575adb220521a096df92e2d502R659)

### DIFF
--- a/provider/github/review.go
+++ b/provider/github/review.go
@@ -40,7 +40,8 @@ func createReview(
 	requests := splitReviewRequest(req, batchReviewComments)
 	for i, req := range requests {
 		_, resp, err := client.PullRequests.CreateReview(ctx, owner, repo, number, req)
-		if err = handleAPIError(resp, err); err != nil {
+
+		if err = handleAPIError(resp, err, "review could not be pushed"); err != nil {
 			return err
 		}
 
@@ -99,7 +100,7 @@ func getPostedComment(ctx context.Context, client *Client, owner, repo string, n
 	var reviews []*github.PullRequestReview
 	for {
 		rs, resp, err := client.PullRequests.ListReviews(ctx, owner, repo, number, listReviewsOpts)
-		if handleAPIError(resp, err) != nil {
+		if handleAPIError(resp, err, "pull request reviews could not be listed") != nil {
 			return nil, err
 		}
 
@@ -118,7 +119,7 @@ func getPostedComment(ctx context.Context, client *Client, owner, repo string, n
 
 		for {
 			comments, resp, err := client.PullRequests.ListReviewComments(ctx, owner, repo, int64(number), review.GetID(), listCommentsOpts)
-			if handleAPIError(resp, err) != nil {
+			if handleAPIError(resp, err, "review comments could not be listed") != nil {
 				return nil, err
 			}
 
--- a/provider/github/watcher.go
+++ b/provider/github/watcher.go
@@ -312,7 +312,7 @@ func (w *Watcher) doPRListRequest(ctx context.Context, client *Client, username,
 
 	prs, resp, err := client.PullRequests.List(ctx, username, repository, &github.PullRequestListOptions{})
 	if err != nil {
-		return resp, nil, ErrGitHubAPI.Wrap(err)
+		return resp, nil, ErrGitHubAPI.Wrap(err, "pull requests could not be listed")
 	}
 
 	if isStatusNotModified(resp.Response) {
@@ -333,7 +333,7 @@ func (w *Watcher) doEventRequest(ctx context.Context, client *Client, username, 
 	)
 
 	if err != nil {
-		return resp, nil, ErrGitHubAPI.Wrap(err)
+		return resp, nil, ErrGitHubAPI.Wrap(err, "repository events could not be listed")
 	}
 
 	if isStatusNotModified(resp.Response) {
--- a/provider/github/client.go
+++ b/provider/github/client.go
@@ -414,14 +414,25 @@ func (t *limitRoundTripper) RoundTrip(req *http.Request) (*http.Response, error)
 
 var _ http.RoundTripper = &limitRoundTripper{}
 
-func handleAPIError(resp *github.Response, err error) error {
+func handleAPIError(resp *github.Response, err error, msg string) error {
 	if err != nil {
-		return ErrGitHubAPI.Wrap(err)
+		if e, ok := err.(*github.ErrorResponse); ok {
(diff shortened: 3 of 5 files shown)


provider/github/review.go
```diff
@@ -99,7 +134,7 @@ func getPostedComment(ctx context.Context, client *Client, owner, repo string, n
 	var reviews []*github.PullRequestReview
 	for {
 		rs, resp, err := client.PullRequests.ListReviews(ctx, owner, repo, number, listReviewsOpts)
-		if handleAPIError(resp, err) != nil {
+		if handleAPIError(resp, err, "pull request reviews could not listed") != nil {
```
carlosms:
```suggestion
		if handleAPIError(resp, err, "pull request reviews could not be listed") != nil {
```

provider/github/review.go
```diff
@@ -118,7 +153,7 @@ func getPostedComment(ctx context.Context, client *Client, owner, repo string, n
 
 		for {
 			comments, resp, err := client.PullRequests.ListReviewComments(ctx, owner, repo, int64(number), review.GetID(), listCommentsOpts)
-			if handleAPIError(resp, err) != nil {
+			if handleAPIError(resp, err, "review comments could not list") != nil {
```
carlosms:
```suggestion
			if handleAPIError(resp, err, "review comments could not be listed") != nil {
```

provider/github/review.go
```diff
@@ -40,7 +40,11 @@ func createReview(
 	requests := splitReviewRequest(req, batchReviewComments)
 	for i, req := range requests {
 		_, resp, err := client.PullRequests.CreateReview(ctx, owner, repo, number, req)
-		if err = handleAPIError(resp, err); err != nil {
+
+		if err = handleAPIError(resp, err, "review could not be pushed"); err != nil {
+			ctxlog.WithLogFields(ctx, log.Fields{
```
carlosms: is this line doing anything? you don't use the returned ctx or logger.

dpordomingo: This line is adding the comments to the logger `Fields`, so they will appear later when performing the real `Errorf` at [server/server.go::L114](https://github.com/src-d/lookout/pull/488/files#diff-91bbeda7eb98a7adc57b9e47e2cf5c2bR114).

An alternative could be to get the logger `Fields` and append this new one (but when doing so, tests didn't pass because of a `nil` dereference, and I didn't want o try harder)

carlosms: I've tested your branch and yes, your code is working but it shouldn't :sweat_smile: 
The original `ctx` log fields should not be modified... https://github.com/src-d/lookout/issues/491

dpordomingo: fair enough. Since I didn't want to block this enhancement till we have an agreement on how to add content to the error, we can follow that in a separate PR.

dpordomingo:
As I said https://github.com/src-d/lookout/pull/488#discussion_r251038793 I moved the disputed commit to a new PR https://github.com/src-d/lookout/pull/492 so the enhancements proposed by this one wont be blocked :dagger:

server/server.go
```diff
@@ -110,7 +110,8 @@ func (s *Server) HandleEvent(ctx context.Context, e lookout.Event) error {
 	if err == nil {
 		status = models.EventStatusProcessed
 	} else {
-		logger.Errorf(err, "event processing failed")
+
+		ctxlog.Get(ctx).Errorf(err, "event processing failed")
```
carlosms: After extracting the commits to #492, I think this change is not needed anymore

dpordomingo: that's true

provider/github/watcher.go
```diff
@@ -333,7 +333,7 @@ func (w *Watcher) doEventRequest(ctx context.Context, client *Client, username,
 	)
 
 	if err != nil {
-		return resp, nil, ErrGitHubAPI.Wrap(err)
+		return resp, nil, ErrGitHubAPI.Wrap(err, "events repository could not be listed")
```
carlosms:
```suggestion
		return resp, nil, ErrGitHubAPI.Wrap(err, "repository events could not be listed")
```

dpordomingo:
I have no clue about why integration tests are now failing...

dpordomingo:
I saw it's already failing in `master` since last Friday
https://travis-ci.org/src-d/lookout/builds/484454336

dpordomingo:
And it is now passing.
https://travis-ci.org/src-d/lookout/builds/485567163
Heisenbug?
image

carlosms:
It's probably because of #490.

PR #493, opened 2019-01-28
Merge organization default config

carlosms:
Part of #401, depends on #483.

The default organization config is be merged with the `.lookout.yml` committed in the repository. The one in the repository has priority over the organization default.

I've made the PR from local branch to local branch to make it easier to review incremental PRs, once accepted everything will be merged to master.


server/server.go
```diff
@@ -350,6 +392,33 @@ func (s *Server) concurrentRequest(ctx context.Context, conf map[string]lookout.
...
+		return local
+	}
+
+	merged := make(map[string]lookout.AnalyzerConfig)
+	for k, v := range global {
+		merged[k] = v
+	}
+
+	for k, v := range local {
+		if globalV, ok := merged[k]; ok {
```
se7entyse7en: I think we're not correctly giving priority to local configuration here. If `k` is already present in `merged` we're just overriding the `settings` of the global configuration with the local ones. But what about if my local configuration has different `Addr` or `Disabled` or `Feedback`?

This should be:
```
		if globalV, ok := merged[k]; ok {
			v.Settings = mergeMaps(globalV.Settings, v.Settings)
			merged[k] = v
			continue
		}
```

server/server.go
```diff
@@ -350,6 +392,33 @@ func (s *Server) concurrentRequest(ctx context.Context, conf map[string]lookout.
...
+			globalV.Settings = mergeMaps(globalV.Settings, v.Settings)
```
se7entyse7en: I'd use `mergeSettings` given that we're actually merging settings.

event.go
```diff
@@ -16,14 +16,81 @@ type Event interface {
...
 
-type PushEvent = pb.PushEvent
-type ReviewEvent = pb.ReviewEvent
+// PushEvent represents a Push to a git repository. It wraps the pb.PushEvent
+// adding information only relevant to lookout, and not for the analyzers.
+type PushEvent struct {
+	pb.PushEvent
+	// OrganizationID is the organization to which this event's repository
+	// belongs to
+	OrganizationID string
```
dpordomingo: I think it should be recorded the `Provider` causing the event, same way we do for `Organization`

server/server.go
dpordomingo: The `Addr` or `Feedback` fields are readed from the server conf instead of the org or repo configurations ([see docs](https://docs.sourced.tech/lookout/configuration#lookout-yml) about repo `.lookout.yml`).

`Disabled` field from org+repo configurations can only disable the analyzer because if it's already disabled by the server conf, this field from org+repo configuration is ignored by `lookoutd` ([sourcecode](https://github.com/src-d/lookout/blob/master/server/server.go#L292)).

server/server.go
```diff
@@ -62,10 +63,13 @@ func NewServer(
 	analyzers map[string]lookout.Analyzer,
 	eventOp store.EventOperator,
 	commentOp store.CommentOperator,
+	organizationOp store.OrganizationOperator,
```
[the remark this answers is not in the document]
se7entyse7en: IMO ~6 parameters are already many arguments to be passed to a function, simply because is annoying when reading to code to mentally keep track of everything that is being passed around. In this case maybe we could simply pass a struct of operators. You made me recalled Scala's standard library where you can define functions up to 22 parameters 😂 (https://www.scala-lang.org/api/2.12.1/scala/Function22.html).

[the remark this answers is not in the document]
dpordomingo: Passing a `conf` struct seems to be the better way.
On the other hand,
```go
func NewWatever(a, b, c, d, e, f, g, h, y type) Watever {
    return Watever{a, b, c, d, e, f, g, h, y}
}
```
seems a bit weird, since `NewWatever` func is doing nothing but pass arguments to the inside.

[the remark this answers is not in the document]
se7entyse7en: If `NewWhatever` doesn't need to do anything particular, then it means that we could simply `Whatever{a, b, c, d, e, f, g, h, y}` from outside the package.

event.go
carlosms: it's already included in `pb.PushEvent`, https://github.com/src-d/lookout/blob/73ede45df4f6f548090c96bed5b88bf2a8cf0fa9/vendor/gopkg.in/src-d/lookout-sdk.v0/pb/event.pb.go#L118

server/server.go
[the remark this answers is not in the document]
smacker: Just remember guys that `Whatever{...}` wouldn't create compiling error as `NewWhatever(...)` when you add one more argument.

server/server.go
```diff
@@ -62,10 +63,13 @@ func NewServer(
...
 	reviewTimeout time.Duration,
 	pushTimeout time.Duration,
 ) *Server {
-	return &Server{false, p, fileGetter, analyzers, eventOp, commentOp, reviewTimeout, pushTimeout}
+	return &Server{false, p, fileGetter, analyzers,
```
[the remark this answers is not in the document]
carlosms: Also fixed in in 1589995.

server/server.go
se7entyse7en: Ok, thanks for the clarification @dpordomingo.

dpordomingo:
@carlosms It seems all the comments were already addressed, so this is ready to be merged

carlosms:
I want to add a couple of test first

carlosms:
Tests added in 74ce6f0, including a fix for missing merge code on push events 5ed4836.
Rebased on top of master, CI passes, ready to merge.

PR #494, opened 2019-01-28
Fix pointer references in ctxlog

### LINKED ISSUE #491: ctxtlog WithLogFields modifies the original context values
Context here: https://github.com/src-d/lookout/pull/488/files#r250727419

I think the problem is probably here, we should be making a copy of the map https://github.com/src-d/lookout/blob/master/util/ctxlog/context.go#L47

carlosms:
Fix #491.


dpordomingo:
:pray:

PR #495, opened 2019-01-28
Enhance comment footer on PR messages

### LINKED ISSUE #472: Enhance comment footer on PR messages
slightly related to https://github.com/src-d/lookout/issues/306 (same context)

Currently there is no an easy way to know where a lookout message come from, so different messages from different analyzers looks the same.

We could itterate a lot about it, like adding a picture of the analyzer, or different templates for them, but I think we're not there yet.

But we could, at least, identify the analyzer sending the message, using the footer on every comment posted by lookout.

dpordomingo:
fix #472 
depends on #510

If the template uses old `sprintf` format (with `%s`), `lookoutd` will log a Warning:
```
WARN invalid footer template: old footer template with '%s' placeholder is no longer supported:
'_If you have %s feedback about this comment made by the analyzer {{.Name}}, please, [tell us]({{.Feedback}})._'
app=lookoutd
```
and no footer will be added; same behavior in other cases like: wrong template, no enough data passed to the template...

ussage, from docs in this PR:

### Add a Custom Message to the Posted Comments

You can configure **source{d} Lookout** to add a custom message to every comment that each analyzer returns. This custom message will be created passing to the template defined by `providers.github.comment_footer`, the `name` and `feedback` defined for each analyzer configuration.

If the template (`providers.github.comment_footer`) is empty, or the analyzer configuration does not define any of the values that the template requires, the custom message won't be added.

Example:
```yaml
providers:
  github:
    comment_footer: "Comment made by analyzer {{.Name}}, [report]({{.Feedback}})."
```

Will require that each analyzer defines its `name` and `feedback` in order to render its custom message appended to its comments, so:

```yaml
analyzers:
  - name: Fancy Analyzer
    addr: ipv4://localhost:9930
    feedback: http://example.com/report-issue
  - name: Awesome Analyzer
    addr: ipv4://localhost:9931
```

Comments from `Fancy Analyzer` will be appended with the message:
>_Comment made by analyzer Fancy Analyzer, [report](http://example.com/report-issue)._

but comments from `Awesome Analyzer` wont be appended with a custom message because its configuration lacks of a key defining its `feedback` value.

### DIFF
--- a/config.yml.tpl
+++ b/config.yml.tpl
@@ -8,7 +8,7 @@ analyzers:
 
 providers:
   github:
-    comment_footer: '_If you have feedback about this comment, please, [tell us](%s)._'
+    comment_footer: "_{{if .Feedback}}If you have feedback about this comment made by the analyzer {{.Name}}, please, [tell us]({{.Feedback}}){{else}}Comment made by the analyzer {{.Name}}{{end}}._"
     # The minimum watch interval to discover new pull requests and push events
     watch_min_interval: 2s
     # Authorization with GitHub App
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -39,14 +39,14 @@ The `providers.github` key configures how **source{d} Lookout** will connect wit
 ```yaml
 providers:
   github:
-    comment_footer: "_If you have feedback about this comment, please, [tell us](%s)._"
+    comment_footer: "_Comment made by '{{.Name}}'{{with .Feedback}}, [tell us]({{.}}){{end}}._"
     # app_id: 1234
     # private_key: ./key.pem
     # installation_sync_interval: 1h
     # watch_min_interval: 2s
 ```
 
-`comment_footer` key defines a format-string that will be used for custom messages for every message posted on GitHub; see how to [add a custom message to the posted comments](#add-a-custom-message-to-the-posted-comments)
+`comment_footer` key defines the [go template](https://golang.org/pkg/text/template) that will be used for custom messages for every message posted on GitHub; see how to [add a custom message to the posted comments](#add-a-custom-message-to-the-posted-comments)
 
 ### Authentication with GitHub
 
@@ -141,16 +141,30 @@ analyzers:
 
 ### Add a Custom Message to the Posted Comments
 
-You can configure **source{d} Lookout** to add a custom message to every comment that each analyzer returns. This custom message will be created following the rule:
-```
-sprintf(providers.github.comment_footer, feedback)
-```
-If any of those two keys are not defined, the custom message won't be added.
+You can configure **source{d} Lookout** to add a custom message to every comment that each analyzer returns. This custom message will be created from the template defined by `providers.github.comment_footer`, using the configuration set for each analyzer.
 
-Example:
-```text
-"_If you have feedback about this comment, please, [tell us](%s)._"
+If the template (`providers.github.comment_footer`) is empty, or the analyzer configuration does not define any of the values that the template requires, the custom message won't be added.
+
+For example, for this configuration, each analyzer needs to define `name` and `settings.email`:
+
+```yaml
+providers:
+  github:
+    comment_footer: "Comment made by analyzer {{.Name}}, [email me]({{.Settings.email}})."
+
+analyzers:
+  - name: Fancy Analyzer
+    addr: ipv4://localhost:9930
+    settings:
+      email: admin@example.org
+  - name: Awesome Analyzer
+    addr: ipv4://localhost:9931
 ```
+
+Comments from `Fancy Analyzer` will have this footer appended:
+>_Comment made by analyzer Fancy Analyzer, [email me](admin@example.org)._
+
+but comments from `Awesome Analyzer` wont have a footer message because in its configuration it's missing the `settings.email` value.
 
 
 ## Timeouts
--- a/provider/github/poster.go
+++ b/provider/github/poster.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
+	"text/template"
 
 	"github.com/src-d/lookout"
 	"github.com/src-d/lookout/util/ctxlog"
@@ -32,18 +33,27 @@ const (
 
 // Poster posts comments as Pull Request Reviews.
 type Poster struct {
-	pool *ClientPool
-	conf ProviderConfig
+	pool           *ClientPool
+	conf           ProviderConfig
+	footerTemplate *template.Template
 }
 
 var _ lookout.Poster = &Poster{}
 
 // NewPoster creates a new poster for the GitHub API.
-func NewPoster(pool *ClientPool, conf ProviderConfig) *Poster {
-	return &Poster{
-		pool: pool,
-		conf: conf,
+func NewPoster(pool *ClientPool, conf ProviderConfig) (*Poster, error) {
+	tpl, err := newFooterTemplate(conf.CommentFooter)
+	if ErrEmptyTemplate.Is(err) {
+		log.DefaultLogger.Warningf("no footer template being used: %s", err)
+	} else if err != nil {
+		return nil, err
 	}
+
+	return &Poster{
+		pool:           pool,
+		conf:           conf,
+		footerTemplate: tpl,
+	}, nil
 }
 
 // Post posts comments as a Pull Request Review.
@@ -157,14 +167,10 @@ func (p *Poster) createReviewRequest(
 	}
 
 	var bodyComments []string
-	tmpl := p.conf.CommentFooter
-
 	for _, aComments := range aCommentsList {
 		ctx, _ := ctxlog.WithLogFields(ctx, log.Fields{
 			"analyzer": aComments.Config.Name,
 		})
-
-		url := aComments.Config.Feedback
 
 		forBody, ghComments := convertComments(ctx, aComments.Comments, dl)
 
@@ -175,13 +181,13 @@ func (p *Poster) createReviewRequest(
 		ghComments = mergeComments(ghComments)
 
 		for i, c := range ghComments {
-			body := addFootnote(c.GetBody(), tmpl, url)
+			body := addFootnote(ctx, c.GetBody(), p.footerTemplate, &aComments.Config)
 			ghComments[i].Body = &body
 		}
 
 		bodyComments = append(
 			bodyComments,
-			addFootnote(strings.Join(forBody, "\n\n"), tmpl, url),
+			addFootnote(ctx, strings.Join(forBody, "\n\n"), p.footerTemplate, &aComments.Config),
 		)
 		req.Comments = append(req.Comments, ghComments...)
 	}
--- a/provider/github/review.go
+++ b/provider/github/review.go
@@ -2,14 +2,15 @@ package github
 
 import (
 	"context"
-	"fmt"
 	"sort"
 	"strings"
+	"text/template"
 	"time"
 
 	"github.com/google/go-github/github"
 	"github.com/src-d/lookout"
 	"github.com/src-d/lookout/util/ctxlog"
+	errors "gopkg.in/src-d/go-errors.v1"
 	log "gopkg.in/src-d/go-log.v1"
 )
 
@@ -28,6 +29,13 @@ const commentsSeparator = "\n<!-- lookout comment separator -->\n---\n"
 
 // comment can contain footer with link to the analyzer
 const footnoteSeparator = "\n<!-- lookout footnote separator -->\n"
+
+var (
+	ErrEmptyTemplate = errors.NewKind("empty footer template")
+	ErrOldTemplate   = errors.NewKind("old footer template: '%%s' placeholder is no longer supported: '%s'")
+	ErrParseTemplate = errors.NewKind("error parsing footer template: %s")
+	ErrTemplateError = errors.NewKind("error generating the footer: %s")
+)
 
 // createReview creates pull request review on github using multiple http calls
 // in case of too many comments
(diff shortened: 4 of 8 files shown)


docs/configuration.md
```diff
@@ -39,14 +39,14 @@ The `providers.github` key configures how **source{d} Lookout** will connect wit
 ```yaml
 providers:
   github:
-    comment_footer: "_If you have feedback about this comment, please, [tell us](%s)._"
+    comment_footer: "_Comment made by analyzer '{{.Name}}', [report]({{.Feedback}})._"
```
carlosms: Why not use the default value? `_If you have feedback about this comment made by the analyzer {{.Name}}, please, [tell us]({{.Feedback}})._`

provider/github/review.go
```diff
@@ -207,13 +208,64 @@ func splitReviewRequest(review *github.PullRequestReviewRequest, n int) []*githu
 	return result
 }
 
+var (
+	ErrEmptyTemplate   = errors.NewKind("empty footer template")
+	ErrOldTemplate     = errors.NewKind("old footer template: '%%s' placeholder is no longer supported: '%s'")
+	ErrParseTemplate   = errors.NewKind("error parsing footer template: %s")
+	ErrNoDataForFooter = errors.NewKind("there is no enough data to generate the footer")
```
carlosms:
```suggestion
	ErrNoDataForFooter = errors.NewKind("there is not enough data to generate the footer")
```

provider/github/poster.go
```diff
@@ -32,17 +33,24 @@ const (
...
+	footerTemplate *template.Template
 }
 
 var _ lookout.Poster = &Poster{}
 
 // NewPoster creates a new poster for the GitHub API.
 func NewPoster(pool *ClientPool, conf ProviderConfig) *Poster {
+	tpl, err := newFooterTemplate(conf.CommentFooter)
+	if err != nil {
+		log.DefaultLogger.Warningf("no footer template being used: %s", err)
```
carlosms: I think we discussed via chat to not exit with error for an older template format. But thinking again about it, I've changed my mind and I think it goes more in line with other misconfigured settings to exit `lookoutd` if the template is invalid.
Otherwise someone that doesn't update the template may ignore this warning, and then wonder whey the comments don't have a footer anymore.

docs/configuration.md
```diff
@@ -141,17 +141,33 @@ analyzers:
 
 ### Add a Custom Message to the Posted Comments
 
-You can configure **source{d} Lookout** to add a custom message to every comment that each analyzer returns. This custom message will be created following the rule:
-```
-sprintf(providers.github.comment_footer, feedback)
-```
-If any of those two keys are not defined, the custom message won't be added.
+You can configure **source{d} Lookout** to add a custom message to every comment that each analyzer returns. This custom message will be created passing to the template defined by `providers.github.comment_footer`, the `name` and `feedback` defined for each analyzer configuration.
```
carlosms: The sentence is a bit awkward, what about:
```suggestion
You can configure **source{d} Lookout** to add a custom message to every comment that each analyzer returns. This custom message will be created from the template defined by `providers.github.comment_footer`, using the `name` and `feedback` set for each analyzer.
```

docs/configuration.md
```diff
@@ -141,17 +141,33 @@ analyzers:
...
+
+If the template (`providers.github.comment_footer`) is empty, or the analyzer configuration does not define any of the values that the template requires, the custom message won't be added.
 
 Example:
-```text
-"_If you have feedback about this comment, please, [tell us](%s)._"
+```yaml
```
carlosms: Instead of splitting the confing into two, I think it would be more clear to show it as one file. Something like:

--------

For example, for this configuration, each analyzer needs to define `name` and `feedback`:
```yaml
providers:
  github:
    comment_footer: "Comment made by analyzer {{.Name}}, [report]({{.Feedback}})."

analyzers:
  - name: Fancy Analyzer
    addr: ipv4://localhost:9930
    feedback: http://example.com/report-issue
  - name: Awesome Analyzer
    addr: ipv4://localhost:9931
```

Comments from `Fancy Analyzer` will have this footer appended:
>_Comment made by analyzer Fancy Analyzer, [report](http://example.com/report-issue)._

but comments from `Awesome Analyzer` will not have a footer message because its configuration is missing a `feedback` value.

docs/configuration.md
dpordomingo: This is part of the docs; I preferred to be concise to be readable in one line; but if you really want, I can change it.

provider/github/poster.go
dpordomingo: ¬¬ I hate you :P

ok, I'll do so ;)

dpordomingo: done by 5d7b11ed209f0551fa2839892fba91aa231de279

dpordomingo:
ready for a new review plus https://github.com/src-d/lookout/pull/495#discussion_r251955505

docs/configuration.md
```diff
@@ -141,17 +141,30 @@ analyzers:
...
-```text
-"_If you have feedback about this comment, please, [tell us](%s)._"
+If the template (`providers.github.comment_footer`) is empty, or the analyzer configuration does not define any of the values that the template requires, the custom message won't be added.
+
+For example, for this configuration, each analyzer needs to define `name` and `feedback`:
+
+```yaml
+providers:
+  github:
+    comment_footer: "Comment made by analyzer {{.Name}}, [report]({{.Feedback}})."
```
se7entyse7en: Could the maintainer of multiple analyzers be willing to have a different footer message for each analyzer? Additionally the maintainer could also want to have different message for different providers for the same analyzer. Anyway I don't consider this as blocking, but just something that we could consider if this scenario is possible. Who's the responsible of choosing the footer message?

provider/github/review.go
```diff
@@ -207,13 +216,56 @@ func splitReviewRequest(review *github.PullRequestReviewRequest, n int) []*githu
...
+	footer, err := getFootnote(tmpl, analyzerConf)
+	if err != nil {
+		ctxlog.Get(ctx).Warningf("footer could not be generated: %s", err)
+	}
+
+	return comment + footer
+}
+
+func getFootnote(tmpl *template.Template, analyzerConf *lookout.AnalyzerConfig) (string, error) {
+	if analyzerConf == nil || analyzerConf.Feedback == "" || analyzerConf.Name == "" {
```
se7entyse7en: Isn't it more flexible to avoid checking for `Feedback` and `Name` in the configuration? I mean if would just execute the template and throw an error if some keys are missing, but I would not enforce having `Feedback` and `Name` in the configuration, for two reasons:
1. I might still want to use a static message for some reason (so no need to for extra keys in analyzer configuration),
2. I might want to use a custom footer whose keys are not `Feedback`  and `Name`. As an example:

````
providers:
  github:
    comment_footer: "_Comment made by analyzer '{{.Name}}' (version '{{.Version}}'), [report]({{.Feedback}}) or [email]({{.Email}})._"
analyzers:
  - name: Fancy Analyzer
    addr: ipv4://localhost:9930
    feedback: http://example.com/report-issue
    version: 1.0.0
    email: superdev@sourced.tech
  - name: Awesome Analyzer
    addr: ipv4://localhost:9931
```

docs/configuration.md
dpordomingo: Currently, the footer template is defined by `lookoutd` administrator.
I also agree that it should be (somehow) up to the analyzer maintainer, but I think it will come at some point as a feature request from @src-d/product

provider/github/review.go
se7entyse7en: > it is not expected to pass custom fields to the footer, because `AnalyzerConfig` has already a fixed schema.

You're right, I missed that. Maybe by putting something in the `Settings`. Anyway

> For sure I think we can (should) improve the way the footer is configured from the analyzers, but imo this PR shouldn't be the place.

I agree that it is certainly not to be done in this PR and maybe wait for inputs from @src-d/product. 👍

carlosms: I agree with @se7entyse7en, if we are giving the flexibility of a template, it makes sense to allow templates that don't use any fields. Skipping `feedback` in the `yml` should be fine if it's not used by the template.

If we want to enforce a failure for an empty `feedback`, IMO it should be done with an exit on initialization.

dpordomingo: @se7entyse7en Template can already take values from `settings` key in `lookoutd` config.
```yml
providers:
  github:
    comment_footer: "_[open an issue]({{.Feedback}}) or send us an [email]({{.Settings.email}})._"
analyzers:
  - name: Fancy Analyzer
    feedback: http://example.com/report-issue
    settings:
        email: admin@example.org
```

dpordomingo: @carlosms if we don't validate `conf.Feedback`, it might be posted "wrong footers" when `conf.Feedback` is not provided;
let's consider this footer template:
```yml
    comment_footer: "If you have feedback, [open an issue]({{.Feedback}})"
```
if `conf.Feedback` is not provided for a certain analyzer, its footer will look as it follows:
```
If you have feedback, [open an issue]()
```
^ with a broken link.

Why? because the default value of the `conf.Feedback` is an empty string, so the template will be executed with no errors.

Should I parse the template to check if `{{.Feedback}}` placeholder is present? If we do so, then we can return the error if the `conf.Feedback` is empty only when it is required by the template.

imho :point_up: this is too much for this use case, and should be considered in a new issue about developing a fully customizable footer, but whatever you prefer.

se7entyse7en: Ok! I missed that, that's cool! So we can actually avoid enforcing `feedback` in the conf right?

dpordomingo: As I said, we can ignore the absence of `conf.Feedback` when executing the template if:
- we don't care about broken links:
  ```
  If you have feedback, [open an issue]()
  ```
- we parse the template to ensure that `{{.Feedback}}` is not being used

se7entyse7en: Yup, sorry I've just realized that I didn't refresh the page 😅, and I missed your comment. I never worked with templates so far, so I don't know how much effort would it take. But I'm ok in doing it in a separated issues. As you prefer.

dpordomingo:
About your `.conf.Feedback` complain, I'd purpose two different solutions to skip validating wether it is provided or it's not.
- **(a)** since footer is a go template, we can move the responsability of avoiding broken links to the source{d} Lookout administrator, in a way that they should provide a reliable template, not looking bad if something is missing
- **(b)** the problem with `.conf.Feedback` appears because it has a default value, so template execution returns no error if it is empty; if we move `conf.feedback` to `conf.settings.feedback`, and then we execute the template only with the content of `conf.Settings`, we would benefit of:
  1. no default values for non defined keys,
  1. clear docs: "_the footer is built with the value of analyzer `.Name` and `.Settings`_"
  1. no internal values will be exposed to the template, like `conf.Addr` or `conf.disabled` or any other value that could be added to the analyzer config.

option **(a)** is already implemented by 63fb1fd
if you prefer **(b)** just let me know.

provider/github/review.go
dpordomingo: I would not choose to parse the template; I'd choose one of https://github.com/src-d/lookout/pull/495#issuecomment-461103028

Since 63fb1fd already addresses the suggestion made in this thread, I'll consider it as "resolved"; if you want to consider another approach, just say so :dancer:

dpordomingo:
I'd rather **(b)**, but since it requires a small change in the design, I'd prefer to obtain your validation before start working on it.

se7entyse7en:
I really like option **(b)**. But it's ok for me to close this as-is and then open another issue for that. 👍

carlosms:
I prefer **a**. `settings` can be overwritten by the end user with `.lookout.yml` file in their repo. We can reconsider improvements when we introduce the feedback based on comment reactions.

config.yml.tpl
```diff
@@ -8,7 +8,7 @@ analyzers:
 
 providers:
   github:
-    comment_footer: '_If you have feedback about this comment, please, [tell us](%s)._'
+    comment_footer: "_If you have feedback about this comment made by the analyzer {{.Name}}{{with .Feedback}}, please, [tell us]({{.}}){{end}}._"
```
carlosms: What do you get if feedback is not defined? 

> If you have feedback about this comment made by the analyzer style-analyzer.

The sentence feels unfinished.

se7entyse7en:
> `settings` can be overwritten by the end user with `.lookout.yml` file in their repo.

Oh, right. That's dangerous.

config.yml.tpl
dpordomingo: If the repo maintainers want to override the feedback URL, why not letting them do it?
Maybe they misconfigured the analyzer with their local `.lookout.yml`, and they want to be informed about that. Or maybe that way the repo contributors can ask the repo maintainer to disable the analyzer in that repo because it produces no valuable comments...

dpordomingo:
@carlosms @se7entyse7en I see reasonable to let the repo maintainers override the feedback URL that is being used in the comments posted in their repos?

Maybe they misconfigured the analyzer with their local `.lookout.yml`, and they want to be informed about that. Or maybe that way the repo contributors can ask the repo maintainer to disable the analyzer in that repo because it produces no valuable comments...

config.yml.tpl
dpordomingo: goot catch; addressed by https://github.com/src-d/lookout/pull/495/commits/d34cab394be1372f18a3706b99005e300ae7380e

carlosms:
Still, I see that as a different use case
`settings` was introduced to setup different options for the analyzer, so it works better for the specific repository.

I think it makes sense that the feedback link can be changed by the people deploying lookout and the analyzers (let's say the SaaS admin); but not any users of the SaaS. The comment is posted by the gh app bot, and the contents of those comments should be only managed by the owner of said bot.

What if I installed lookout on my repo, and changed the feedback link to some malicious link? The comment would still be posted as lookout bot, and the admin would be responsible.

dpordomingo:
So then all PR comments were already addressed.

se7entyse7en:
👏

PR #518, opened 2019-02-08
Limit travis builds to master and release tags

### LINKED ISSUE #517: Run travis on master only

carlosms:
Fix #517

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,9 @@
+branches:
+  only:
+  - master
+  - /^v\d+\.\d+(\.\d+)?(-\S*)?$/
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,3 +1,9 @@
+branches:
+  only:
+  - master
+  - /^v\d+\.\d+(\.\d+)?(-\S*)?$/
```
dpordomingo: if we don't need to be strict with version names, this could be simplified as 
```suggestion
  - /^v\d+.*$/
```

carlosms: Yes, but if we leave the first regexp, it can serve as check to avoid tag builds if we make a mistake with the semver. e.g. a tag of `v1.2..3` will not be run on travis.

PR #528, opened 2019-02-11
Ensures clean cache on start-up

### LINKED ISSUE #485: No PRs are reviewed if GitHub API responses from Cache
When GitHub API answer from cache [provider/github/watcher.go#L363](https://github.com/src-d/lookout/blob/master/provider/github/watcher.go#L363)
```go
resp.Header.Get("X-From-Cache") == "1"
```
 
the project PRs are not listed for review because it is returned a `NoErrNotModified` error [provider/github/watcher.go#L318](https://github.com/src-d/lookout/blob/master/provider/github/watcher.go#L318)

so no PR is handled [provider/github/watcher.go#L198](https://github.com/src-d/lookout/blob/master/provider/github/watcher.go#L198).

This is a problem if lookout didn't review that `refs/pull/%d/head` yet, but it is returned from cache by GitHub API. I think it can happen if the DB is dropped between reviews, or if the last review failed.

se7entyse7en:
Closes #485. 

Actually I think that the cleanest solution should be to add a `DeleteAll` method [here](https://github.com/gregjones/httpcache/blob/master/diskcache/diskcache.go#L36) that calls the underlying `EraseAll` [here](https://github.com/peterbourgon/diskv/blob/master/diskv.go#L519). But the repository doesn't seem to be very active.

### DIFF
--- a/cmd/lookoutd/common.go
+++ b/cmd/lookoutd/common.go
@@ -10,6 +10,7 @@ import (
 	"runtime"
 	"time"
 
+	"github.com/gregjones/httpcache/diskcache"
 	"github.com/src-d/lookout"
 	"github.com/src-d/lookout/provider/github"
 	"github.com/src-d/lookout/provider/json"
@@ -25,7 +26,6 @@ import (
 	"github.com/src-d/lookout/util/cli"
 	"github.com/src-d/lookout/util/grpchelper"
 
-	"github.com/gregjones/httpcache/diskcache"
 	"github.com/jinzhu/copier"
 	"github.com/sanity-io/litter"
 	"google.golang.org/grpc"
@@ -83,6 +83,9 @@ type queueConsumerCommand struct {
 }
 
 var defaultInstallationsSyncInterval = 5 * time.Minute
+
+// diskcachePath is the path used on disk for caching responses
+var diskcachePath = "/tmp/github"
 
 // Config holds the main configuration
 type Config struct {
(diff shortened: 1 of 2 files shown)


provider/github/watcher.go
```diff
@@ -195,7 +195,11 @@ func (w *Watcher) processRepoPRs(
 		return c.watchMinInterval, nil
 	}
 
-	if err != nil && !NoErrNotModified.Is(err) {
+	if NoErrNotModified.Is(err) {
+		return c.watchMinInterval, nil
```
carlosms: The previous code was calling `handlePrs` if the error is `NoErrNotModified`, now we skip it. Is it safe in any case? Was it a bug, then?

se7entyse7en: It was not a bug, with this change everything works as before, but I think it's just clearer as we avoid calling `handlePrs`. 
The `doPRListRequest` method returns `nil` for `prs` when a `NoErrNotModified` error occurs, and `handlePrs` just returns a `nil` error if you pass it a zero-length `prs`.

se7entyse7en:
@dpordomingo the reason I'm doing it in the `serve` command is because currently it is the only one using a memory queue. Using a persistent queue doesn't require resetting the cache as unacknowledged jobs are not lost, but persisted in the queue.

dpordomingo:
I wonder what would happen if it is started `serve` first, and then stopped, and then `watch`/`work`.
Won't it reproduce the bug?

provider/github/watcher.go
carlosms: I see, thanks

carlosms:
> I wonder what would happen if it is started `serve` first, and then stopped, and then `watch`/`work`.
> Won't it reproduce the bug?

If you really put effort into it, you can reproduce the bug of course. Or you could be using `watch`/`work`, delete the DB and RabbitMQ, and then the persistent http cache will prevent the processing of older PRs as you would expect.

I think it makes sense to also clear the cache for `watch` as @dpordomingo suggests, for consistency.

dpordomingo:
I'd say: that unacknowledged cache should be deleted before being started. If I'm not wrong, I remember that @rporres said that we should assume that it is not guaranteed that jobs in RabbitMQ are not lost. If it happens, that cache might prevent the processing of old or pending PRs.

se7entyse7en:
I've centralized the cache creation as @dpordomingo suggested.

cmd/lookoutd/common.go
```diff
@@ -84,6 +84,9 @@ type queueConsumerCommand struct {
 
 var defaultInstallationsSyncInterval = 5 * time.Minute
 
+// DiskcachePath is the path used on disk for caching responses
+var DiskcachePath = "/tmp/github"
+
```
smacker: any reason to make this variable public?

se7entyse7en: no reason 👍

Issue #536, opened 2019-02-12
Use a fixed version of esc


carlosms:
For `make dependencies` and `pack-migrations` we are using the latest master of https://github.com/mjibson/esc, we should be using a fixed version.

dpordomingo:
Using dep it can be locked. I'll pull request it tomorrow morning after the Empathy Session :dagger:

PR #582, opened 2019-03-06
Use a fixed version of esc

dpordomingo:
fix https://github.com/src-d/lookout/issues/536

In order to avoid updates in `esc` that would create different assets (what would cause our CI to reject the build), we could lock to a specific version.

Since `esc` does not [release](https://github.com/mjibson/esc/releases) frequently (its last release [v0.1.0](https://github.com/mjibson/esc/releases/tag/v0.1.0) was more than 1year ago) we need to use our own way to target a version.

### Alternatives:
- commit its `vendor` content instead of ignoring it (doing so we'll avoid loosing `esc` in case something disappears as it happened with `bindata`)
- treat it like any other dependency, and add it into the main `Gopkg.toml` and `vendor`.


smacker:
Actually we are using/going to use `esc` in different projects. And not only apps projects.
Do you think it makes sense to include it into `ci` repo, similar to [dep](https://github.com/src-d/ci/blob/v1/Makefile.main#L245)?

Makefile
```diff
@@ -22,6 +21,9 @@ $(MAKEFILE):
 	git clone --quiet --depth 1 -b $(CI_BRANCH) $(CI_REPOSITORY) $(CI_PATH);
 -include $(MAKEFILE)
 
+dependencies:
+	$(MAKE) -C tools dependencies
```
se7entyse7en: Can you explain to me what this line does? I'm not very familiar with makefiles yet.

se7entyse7en: Oh ok, I guess that it is equivalent to `cd tools && make dependencies` right?

smacker: you understood it correctly.

dpordomingo:
About adding it to `ci` I'd wait till we migrate to `go mod` and avoid blocking this because of that.
But yes, eventually `esc` should be provided by `ci`

smacker:
if we are talking about ci I'm not sure fixed version should depend on package manager as go mod or dep. As an example, `godep` target depends on neither of them.

smacker:
also CI is failing

dpordomingo:
`godep` target relies on `dep` installer, but `esc` has not that thing.
Should we try to replicate it and create one for `esc`? I don't think so.

Is it better to `wget` the `zip` containing its sources?https://github.com/mjibson/esc/archive/0ea7db170df78dcddf3e223365f444163147fe89.zip
I don't know... I just tried to take an approach closer to how we handle `DEPENDENCIES`: with `go get`, but with a manager being able to use locked versions.

For sure, I'm open to your suggestions :dancing_men:
They're very welcomed.

alternative;
halt this till we migrate to `go mod` and even propose an alternative in `src-d/ci`; but I'm not sure if it would be better just to move on, and progress in consecutive itterations.

smacker:
I really like `wget` solution. It is similar to `godep` one and simpler. And I think we can pull request it directly into `ci` repo.

On another hand, I'm absolutetly okay with merging this PR if we have any problems with esc right now (I don't remember any)

dpordomingo:
The problem we already have with `esc` is called `Sword of Damocles` :wink: 
Whenever they update it, it might break our builds as it did in the past a couple of times.

Anyhow; if @carlosms also preffers to download the `zip` and use it, I can go with it.

se7entyse7en:
IMO if the migration to `go mod` is happening in the short term, whatever solution is ok provided that it's not too hacky or "ugly". 😄

carlosms:
I don't have a strong preference here.
What I like about `wget` is that it fits better into the `ci` repo as @smacker said. To apply the code in bblfsh-web we would need to add all the files in `tools`, instead of just a new `make` command.

On the other hand having a `Gopkg.toml` allows us to monitor it with dependabot.

How would it look with go mod? What would it take to be able to have this in our makefile?
```
DEPENDENCIES = \
	gopkg.in/src-d/go-kallax.v1 \
	github.com/mjibson/esc@0ea7db1
```
Something like this?
```
export GO111MODULE=on
mkdir tools
cd tools/
go mod init
go get -u github.com/mjibson/esc@0ea7db1
```

Issue #597, opened 2019-03-21
check token permissions on start and exit early


smacker:
Without repo:status permission, for example, lookout produces error like
```
21T17:31:47.422411061Z" level=error msg="event processing failed" app=lookoutd error="posting analysis failed: github api error: POST https://api.github.com/repos/EgorBu/best-practices-experiments/pulls/6/reviews: 404 Not Found []" event-id=872ad4684e184e8d059e3357565648604c9ec373 event-type=*pb.ReviewEvent head=refs/pull/6/head repo="https://github.com/EgorBu/best-practices-experiments.git"
```

which is super difficult to understand how to fix. (give more permissions to the token)

We can check token permission when lookout starts to report missing permissions and exit immediately.

dpordomingo:
It should also log a better error if it happens once the app is running (in case of a scope change, for example)

dpordomingo:
Also I was able [to post a comment on Egor's PRs](https://github.com/EgorBu/best-practices-experiments/pull/6#pullrequestreview-217398595) with no problem, but I got the same problem trying to update PR status.
```
lookout_1   | time="2019-03-21T17:49:04.425896303Z" level=info msg="posting analysis"
  app=lookoutd comments=1 event-id=872ad4684e184e8d059e3357565648604c9ec373
  event-type="*lookout.ReviewEvent" github.pr=6 head=refs/pull/6/head provider=github
  repo="https://github.com/EgorBu/best-practices-experiments.git"
lookout_1   | time="2019-03-21T17:49:07.949939976Z" level=error msg="posting status failed"
  app=lookoutd error="github api error: commit status could not be pushed:
    POST https://api.github.com/repos/EgorBu/best-practices-experiments/statuses/66850d582783b57654816e4d09b2aaace6017d18:
    404 Not Found []"
  event-id=872ad4684e184e8d059e3357565648604c9ec373 event-type="*lookout.ReviewEvent"
  github.pr=6 head=refs/pull/6/head provider=github
  repo="https://github.com/EgorBu/best-practices-experiments.git" status=success
```
If the user is not able to update the PR status, shouldn't be able to push comments with Lookout.

PR #602, opened 2019-03-27
Check permissions on init for token mode

smacker:
task: #597

### DIFF
--- a/provider/github/client.go
+++ b/provider/github/client.go
@@ -529,7 +529,7 @@ func CanPostStatus(client *Client, repo *repositoryInfo) error {
 	}
 
 	if r.GetPermission() != "admin" && r.GetPermission() != "write" {
-		return fmt.Errorf("client does not have write access to repository %s", repo.FullName)
+		return fmt.Errorf("token doesn't have write access to repository %s", repo.FullName)
 	}
 
 	return err
--- a/provider/github/client.go
+++ b/provider/github/client.go
@@ -481,8 +481,8 @@ func handleAPIError(resp *github.Response, err error, msg string) error {
 // ValidateTokenPermissions checks that client has necessary permissions required by lookout
 // returns error if any is missed
 func ValidateTokenPermissions(client *Client) error {
-	// actually we don't need ALL repo scope but we ask for it in documentation
-	// so let's be consistent
+	// we need `repo` access to be able to read from private repositories
+	// and `repo:status` which is part of `repo` for posting statuses
 	required := []string{"repo"}
 
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
--- a/provider/github/client.go
+++ b/provider/github/client.go
@@ -476,3 +476,14 @@ func ValidateTokenPermissions(client *Client) error {
 
 	return nil
 }
+
+// ClientCanPush check if the client has push access to a repository
+// returns error if it permission is missed
+func ClientCanPush(client *Client, repo *repositoryInfo) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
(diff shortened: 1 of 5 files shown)


smacker:
Moving back to in progress. I don't see easy wait to mock existing tests. Need to try different things.

smacker:
fixed

provider/github/token_transport.go
```diff
@@ -70,13 +70,21 @@ func NewClientPoolFromTokens(
...
+			return nil, err
+		}
 
 		if _, ok := byClients[client]; !ok {
 			byClients[client] = []*repositoryInfo{}
 		}
 
 		byClients[client] = append(byClients[client], repos...)
 		for _, r := range repos {
+			err := ClientCanPush(client, r)
```
carlosms: This is actually checking if the token owner can post comments on the repo, right? It took me a moment, it looks like we will be _pushing_ code to the repo. Maybe it should be called `IsCollaborator`.

provider/github/client.go
```diff
@@ -443,3 +453,47 @@ func handleAPIError(resp *github.Response, err error, msg string) error {
 		msg,
 	)
 }
+
+// ValidateTokenPermissions checks that client has necessary permissions required by lookout
+// returns error if any is missed
+func ValidateTokenPermissions(client *Client) error {
+	// actually we don't need ALL repo scope but we ask for it in documentation
```
carlosms: If we know we don't it would also make sense to update the docs.

provider/github/token_transport.go
smacker: It checks if the token owner can update status. It requires `push` access:

> Users with push access in a repository can create commit statuses for a given SHA.

https://developer.github.com/v3/repos/statuses/#create-a-status

`IsCollaborator` can't be used because collaborator can have only `pull` access. (oops, I don't check for it, I'll fix it).

How about `CanPostStatus`?

provider/github/client.go
smacker: I'm not sure why we say we require "repo" access.

Looks like it was introduced here https://github.com/src-d/lookout/pull/398
@dpordomingo do you remember why we require `repo` instead of `repo:status`?

smacker: figured. we need it for private repos. I'll remove this comment from the code.

provider/github/token_transport.go
carlosms: much better, yes :+1: Otherwise it was a bit confusing

smacker: fixed

provider/github/client.go
dpordomingo: ups, sorry, but blame plane landed in @carlosms :trollface: 
https://github.com/src-d/lookout/pull/294/commits/62c64c107e4854b7a7dfd4f990eaa726a8af2b5b#diff-76e731333fb756df3bff5ddb3b731c46R38

provider/github/client.go
```diff
@@ -443,3 +477,60 @@ func handleAPIError(resp *github.Response, err error, msg string) error {
...
+		return err
+	}
+
+	r, _, err := client.Repositories.GetPermissionLevel(ctx, repo.Owner, repo.Name, user)
+	if err != nil {
+		return err
+	}
+
+	if r.GetPermission() != "admin" && r.GetPermission() != "write" {
+		return fmt.Errorf("client does not have write access to repository %s", repo.FullName)
```
carlosms: To be consistent with the error on `ValidateTokenPermissions` I suggest
```suggestion
		return fmt.Errorf("token doesn't have write access to repository %s", repo.FullName)
```

smacker:
I added one more test. CI passed. Merging.

Issue #596, opened 2019-03-20
test how lookout behaves with draft PRs on github


smacker:
github doesn't run anything on draft prs (most probably it doesn't send notification to github apps).
but lookout doesn't use hooks. We need to test it doesn't post comments to draft prs (to be consistent with all other apps) and does post when status changed to "ready for review"

smacker:
Lookout treats Draft PRs as normal PRs and posts comments. Example: https://github.com/smacker/lookout-test/pull/1

PR #615, opened 2019-04-02
skip draft pull requests

smacker:
github doesn't run any checks on draft PRs
emulate this behaviour by skipping prs as long as they are draft

fix: #596

### DIFF
--- a/provider/github/watcher_test.go
+++ b/provider/github/watcher_test.go
@@ -155,6 +155,30 @@ func (s *WatcherTestSuite) TestWatch() {
 	}
 }
 
+func (s *WatcherTestSuite) TestWatchDraftPR() {
+	called := false
+	ctx, cancel := context.WithCancel(context.Background())
+	s.mux.HandleFunc("/repos/mock/test/pulls", func(w http.ResponseWriter, r *http.Request) {
+		if called {
+			cancel()
+		}
+		fmt.Fprint(w, `[{"id":1},{"id":2,"draft":true},{"id":3}]`)
+		called = true
+	})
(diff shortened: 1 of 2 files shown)


smacker:
wait-wait guys. I'll add the test as we agreed to skip it now.

smacker:
re-pushed with the test

se7entyse7en:
LGTM 👍

provider/github/watcher_test.go
```diff
@@ -155,6 +155,30 @@ func (s *WatcherTestSuite) TestWatch() {
...
 }
 
+func (s *WatcherTestSuite) TestWatchDraftPR() {
+	called := false
+	ctx, cancel := context.WithCancel(context.Background())
+	s.mux.HandleFunc("/repos/mock/test/pulls", func(w http.ResponseWriter, r *http.Request) {
+		if called {
+			cancel()
+		}
+		fmt.Fprint(w, `[{"id":1},{"id":2,"draft":true},{"id":3}]`)
```
carlosms: Wound't it be better to use lookout-test-fixtures here?

smacker: 1. It's a unit test, usually, it's a bad idea to use huge fixtures in unit tests. Unit tests should test the smallest part of the logic possible and it should be easy to understand what they do.
2. We don't have fixtures for PRs list in lookout-test-fixtures repository.

carlosms: I don't entirely agree. In my opinion using fixture data does not automatically transform a unit test into an integration test. In this case the fixture is even vendored, so it's not so different from a hardcoded string.

But I don't want to block this PR for this.

PR #619, opened 2019-04-05
Update javascript driver in staging

meyskens:
The style analyzer now supports JS driver version v2.7.1.
This updates the deployment to undo the pinning of the old version.

Signed-off-by: Maartje Eyskens <maartje@eyskens.me>

### DIFF
--- a/.helm-staging.yml
+++ b/.helm-staging.yml
@@ -49,9 +49,11 @@ bblfshd-sidecar:
     tag: v2.12.0
   drivers:
     languages:
+      # for js we have to lock old version due to style-analyzer
       javascript:
         repository: bblfsh/javascript-driver
         tag: v2.7.1
+      # for the rest we can use latest
       python:
         repository: bblfsh/python-driver
         tag: v2.8.2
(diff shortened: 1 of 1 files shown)


.helm-staging.yml
```diff
@@ -49,11 +49,9 @@ bblfshd-sidecar:
     tag: v2.12.0
   drivers:
     languages:
-      # for js we have to lock old version due to style-analyzer
       javascript:
         repository: bblfsh/javascript-driver
-        tag: v1.2.0
-      # for the rest we can use latest
+        tag: v2.7.1
```
dpordomingo: `2.7.2` is already available.
https://hub.docker.com/r/bblfsh/javascript-driver/tags
We could maybe update to the current latest one?

carlosms: Yes, 2.7.2 is the one included in the bblfsh/bblfshd:v2.12.0-drivers image, we should be using than one please. Unless 2.7.1 is a hard requirement from @src-d/machine-learning.

smacker:
Just a note: currently style-analyzer is in CrashLoopBackOff on staging so we can deploy this without worrying that it would break current analyzer.

.helm-staging.yml
zurk: I think it should be good to use 2.7.2, nothing serious was changed there.

carlosms:
Let's follow style-analyzer and pin it to 2.7.1, then.

PR #623, opened 2019-04-08
[WIP] simplify enry service code

smacker:
this commit introduces constructors for Fn scanners and use them in enry
service instead of custom constructor for each enry feature

It doesn't simplify too much in terms of lines of code (removed only 50 lines) but it removes one level of abstraction which makes it simpler in my opinion.

Enry service is an example, I can update other services in the same way if we agree it's good idea.
@carlosms @se7entyse7en WDYT?


se7entyse7en:
> it removes one level of abstraction which makes it simpler in my opinion.

I agree with this as the layer of abstraction is also very thin. Maybe I'd have added an `AppendFn` method to `FnChangeScanner`.

As a side note, while looking at `GetChanges` and `GetFiles` I was wondering whether the order of the filter functions that we add is the optimal one. I mean that the filters that we know potentially prune the most should be the first ones for better performance.

se7entyse7en:
Well I think that the most important is that the filter for excluding vendor is the first one and it's already as is.

smacker:
ping @carlosms

Issue #154, opened 2018-08-17
Retry for poster


smacker:
Most probably should be umbrella.

We need to decide if we want some retry specific for github or it should be common retry in server (thought for example http errors don't make much sense for json poster)

We need to cover different errors differently:
1. Temporary error from remote server (for example 500) - should retry N (or forever?) times with an increasing delay
2. Network errors - similar to point 1
3. Permanent errors from remote server (for example 400) - shouldn't retry but mark them somehow (for example put in giveup queue), so we can retry them manually if error was cased by our code
4. Internal errors - not sure what to do here. Most probably similar to point 3.

bzz:
3 solutions discussed for storing re-try state:
 - in memory, global Map of URLs
 - RDBMS, add re-try count
 - external system: Queue or Redis

first 2 solution would be different from the 3rd, preferred one. 
So let's keep it open for now until we have a queue.

dpordomingo:
We already have a queue (RabbitMQ).
Considering the different errors described by @smacker, what would be the retry policy?

se7entyse7en:
Just for curiosity, is there some link to issue/PR/minutes/design doc containing the rationale behind the choice of RabbitMQ as queue instead for example of Kafka?

se7entyse7en:
BTW the easiest thing would be to just re-enqueue failed jobs indefinitely or add a `retried` field. Then we can improve this solution by having another queue for "dead" jobs (as @smacker said a 400 could be interpreted as a permanent error). If we're going to do manual operations on this dead jobs we could also consider using a DB as an alternative instead of using another queue. I'd use a queue if we plan to have another consumer that consumes from this queue and performs something.

I still have to think about how we could handle the `increasing delay`.

se7entyse7en:
One thing that we could do for jobs that failed with a recoverable error, is to push them into a "recoverable" queue with a `scheduled_for` key, and have consumers reading from it. If the read job has a `scheduled_for` datetime that is elapsed, then it is retried otherwise it's re-enqueued or the consumer waits. In the case of republishing in the queue, the delay won't be exact, but I don't think this is important.

smacker:
Kafka isn't a queue.

For other alternatives: we don't have any specific requirements for a queue. More or less any other queue would work for us. But RabbitMQ is most probably the most popular and flexible solution plus it is supported by src-d/go-queue.

smacker:
It's more or less clear how to deal with errors mentioned in the description.
[...]

PR #637, opened 2019-04-22
Poster queue

smacker:
Part of #154 

This PR introduces the ability to call poster through Queue.
Serve & work command use it by default now.

QueuePoster doesn't wrap calls to status endpoint because it can cause an incorrect result. ("success/failure" can be posted before "processing")
Currently, the server doesn't do anything if status posting failed besides logging it.
For me, it looks like making this feature asynchronous would require quite complex logic which might be isn't important for us. So I just left it as is for now.

p.s. The code in `cmd/lookoutd` is very diffictult to understand due to multiple composition of stucts. And very often the names are very confusing. For example `ServeCommand` inherites  `queueConsumerCommand` and `gocli.PlainCommand` structures. But we don't have `queueConsumer` command. Also `serve` does much more, like running watcher or data server, and it's very difficult to figure out where those methods come from.


smacker:
I have tried different approaches on how to fix tests but all of them had some flaws.
At the end, I implemented this super stupid but not very reliable way. Making it more robust (waiting until something instead of `time.Sleep`) would create another monster similar to the one we have now, so I'm not sure we want it.

PR #652, opened 2019-06-20
Deploy the Helm chart from the new stable charts repo

meyskens:
Signed-off-by: Maartje Eyskens <maartje@eyskens.me>


carlosms:
Is it ok that https://src-d.github.io/ is a 404?

meyskens:
@carlosms yup the charts are at https://src-d.github.io/charts/stable/ where `/charts` is the repo, Helm also uses `index.yaml` so you will get a 404 on the path also.

